Issue #111, opened 2025-06-16
broken output when data is empty


yegor256:
```
$ cat foo.phi
{⟦ m ↦ "" ⟧}
$ phino rewrite --nothing < foo.phi
Φ ↦ ⟦
  m ↦ Φ.org.eolang.string(
    α0 ↦ Φ.org.eolang.bytes(
      α0 ↦ ⟦
        Δ ⤍ ,
        ρ ↦ ∅
      ⟧
    )
  ),
  ρ ↦ ∅
⟧
```

I believe, it should be `Δ ⤍ --` instead.

0crat:
@yegor256 Hey there, bug-finder extraordinaire! 🐛 You've just scored yourself 12 points for that sharp eye of yours. Nice catch! Your total is now [+171](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep 'em coming - if you spot anything else that [seems off](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html), [don't hold back](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html). Every [bug report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) helps us level up our project. You're making a real difference!

PR #113, opened 2025-06-16
bug(#111): empty string to bytes

maxonfjvipon:
Closes: #111


maxonfjvipon:
@rultor merge

maxonfjvipon:
@rultor stop

maxonfjvipon:
@rultor merge

src/Misc.hs
```diff
@@ -88,7 +88,13 @@ numToHex num = btsToHex (unpack (toLazyByteString (word64BE (doubleToWord num)))
 -- "68-65-6C-6C-6F"
 -- >>> strToHex "world"
 -- "77-6F-72-6C-64"
+-- >>> strToHex ""
+-- "--"
+-- >>> strToHex "h"
+-- "68-"
```
maxonfjvipon, line 94: @yegor256 these function call results are auto calculated in Haskell. So we can claim them as micro tests. Should be enough for such small bug

Issue #613, opened 2026-03-07
rewriting takes 659 seconds on  2528 classes


volodya-lombrozo:
When we are trying to optimise a middle-sized project (with only 2528 classes) with using `phino`,  it takes around 659 seconds to apply all the rewriting rules to these classes:

```
Finished rewriting 2528 file(s) in 659 seconds
```

It's too much time for such a number of classes. Is it possible to speed `phino` up?

Here is the project on which we run `phino`:
https://github.com/apache/hadoop/tree/trunk/hadoop-common-project/hadoop-common

`phino` version is `0.0.0.43`.

This issue was initially created in the [hone-maven-plugin](https://github.com/objectionary/hone-maven-plugin) repo:
https://github.com/objectionary/hone-maven-plugin/issues/434

volodya-lombrozo:
@yegor256 I don't posses this information

maxonfjvipon:
@volodya-lombrozo I'm closing it for now because it's more related to hone repo

PR #675, opened 2026-05-05
bug(#218): anonymous meta variables

### LINKED ISSUE #218: `𝜏?` syntax would be helpful (for matching what's not used later)
This is my rule:

```
name: foo
pattern: '⟦ 𝜏1 ↦ 𝑒1, 𝜏2 ↦ 𝑒2 ⟧'
result: '⟦ x ↦ "yes!" ⟧'
```

Here, none of the substitutions are used in the result. However, I need to invent names for them, like `𝜏1` and `𝜏2`. 

Would be more convenient to use `𝜏?` (or `𝜏*`), which means `𝜏 whatever`. It's uniquely matched, but not used in the result.

Maybe, for the sake of brevity, we may agree that `𝜏` (without index) means `𝜏 whatever`. If index is specified, it's a specific match, possibly used in the result.

sherenkov:
## Summary

A meta variable written without any index — bare `𝜏`, `𝐵`, `𝑒`, `δ` (and the ASCII `!a`, `!B`, `!e`, `!d`, `!F`, `!t`) — now means **anonymous**: it matches anything in the pattern but cannot be referenced in `result`, `when`, `where`, or `having`. Each occurrence is matched independently, so two bare `𝜏`s in the same pattern do **not** need to bind to the same value. Use indexed names (`𝜏1`, `𝐵1`, …) whenever you need to refer to the matched value later.

This was settled in [#218](https://github.com/objectionary/phino/issues/218) (uniform across every meta sigil; runtime error rather than parse-time rejection when an anonymous name is referenced).

### Implementation

- `src/Parser.hs` — bare metas now parse to a sentinel name `_anon_<ch>_<offset>` derived from the parser offset, instead of binding to the canonical short name (`a`/`B`/`e`/...).
- `src/Matcher.hs` — adds `isAnonMeta` / `filterAnon`. Anonymous keys are kept in the substitution map (the rewriter needs them to reconstruct the matched fragment) but filtered out before reaching result-build, conditions and `where` extras.
- `src/Builder.hs` — lookup failure for an `_anon_*` key produces a clearer message: *"anonymous meta variable cannot be referenced (it has no index)"*.
- `src/Rewriter.hs`, `src/Rule.hs` — apply `filterAnon` at the public-substitution boundary (result-build in the rewriter; conditions and `where` extras in the rule engine).
- `src/CST.hs`, `src/Printer.hs` — render anonymous metas back as the bare sigil (no suffix) and hide them from `match` output.
- `resources/*.yaml` — built-in normalization rules updated: bare metas referenced elsewhere in the rule renamed to `𝜏1`/`𝐵1`/`𝑒1`; rules like `dd` (`⊥.𝜏`) and `dc` (`⊥(𝜏 ↦ 𝑒)`) get the new wildcard semantics for free.
- `test-resources/**/*.yaml` — fixtures using bare metas referenced in conditions/result/where renamed to indexed (`𝜏0`/`𝐵0`/`𝑒0`).
- `README.md` — short paragraph documenting the new convention.
- Tests added to `test/MatcherSpec.hs`, `test/BuilderSpec.hs`, `test/ParserSpec.hs` covering: bare metas don't conflict with each other, named metas still require equality, an anonymous meta on the right-hand side errors clearly, and the parser produces sentinel names for bare metas.

Fixes #218.

## Test plan

- [x] `make all` passes locally — coverage 69% (threshold 65%), hlint clean, fourmolu clean.
- [x] All 1019 existing tests pass.
- [x] New unit tests cover the anonymous semantics directly.


src/Parser.hs
```diff
@@ -117,13 +117,25 @@ global = choice [symbol "Q", symbol "Φ"]
...
+-- Real meta names always start with a prefix character (a/B/e/d/F/t),
+-- so the "_anon_" prefix can't collide with a user-supplied name.
+-- Within a single parse, each occurrence gets a distinct offset and thus a
+-- distinct name. Cross-parse collisions are mitigated downstream by filtering
+-- "_anon_*" keys out of substitutions before they reach result-build,
+-- conditions or `where` extras.
+anonMetaName :: Char -> Parser T.Text
+anonMetaName ch = do
+  off <- getOffset
+  return (T.pack ("_anon_" ++ ch : '_' : show off))
```
maxonfjvipon: @sherenkov this "technique" with special name for such anon metas across the code is dirty hack. Today it's `"_anon_"`, tomorrow we will change it and everything will be broken. It would be better to add optionality to `AtMeta` in AST and properly handle it in Matcher.

maxonfjvipon (changes requested):

src/Parser.hs
sherenkov: @maxonfjvipon refactored

src/Builder.hs
```diff
@@ -62,15 +69,17 @@ contextualize (ExApplication ex (BiTau at bexpr)) context =
 contextualize ex _ = ex
 
 buildAttribute :: Attribute -> Subst -> Built Attribute
-buildAttribute (AtMeta meta) (Subst mp) = case Map.lookup meta mp of
-  Just (MvAttribute attr) -> Right attr
-  _ -> Left (metaMsg meta)
+buildAttribute (AtMeta meta) (Subst mp) = case lookupNamedMeta meta mp of
+  Right (MvAttribute attr) -> Right attr
+  Right _ -> Left (metaMsg meta)
+  Left err -> Left err
```
maxonfjvipon: @sherenkov I believe it can be simplified and reduced to:
```haskell
buildAttribute (AtMeta meta) (Subst mp) = case lookupNamedMeta meta mp of
  Right (MvAttribute attr) -> Right attr
  _ -> Left (metaMsg meta)
```

src/Builder.hs
```diff
@@ -62,15 +69,17 @@ contextualize (ExApplication ex (BiTau at bexpr)) context =
...
 buildAttribute attr _ = Right attr
 
 buildBytes :: Bytes -> Subst -> Built Bytes
-buildBytes (BtMeta meta) (Subst mp) = case Map.lookup meta mp of
-  Just (MvBytes bytes) -> Right bytes
-  _ -> Left (metaMsg meta)
+buildBytes (BtMeta meta) (Subst mp) = case lookupNamedMeta meta mp of
+  Right (MvBytes bytes) -> Right bytes
+  Right _ -> Left (metaMsg meta)
+  Left err -> Left err
```
maxonfjvipon: @sherenkov the same as above

src/Builder.hs
```diff
@@ -84,15 +93,17 @@ buildBinding (BiTau attr expr) subst = do
 buildBinding (BiVoid attr) subst = do
   attribute <- buildAttribute attr subst
   Right [BiVoid attribute]
-buildBinding (BiMeta meta) (Subst mp) = case Map.lookup meta mp of
-  Just (MvBindings bds) -> uniqueBindings bds
-  _ -> Left (metaMsg meta)
+buildBinding (BiMeta meta) (Subst mp) = case lookupNamedMeta meta mp of
+  Right (MvBindings bds) -> uniqueBindings bds
+  Right _ -> Left (metaMsg meta)
+  Left err -> Left err
```
maxonfjvipon: @sherenkov the same as above

src/Builder.hs
```diff
@@ -84,15 +93,17 @@ buildBinding (BiTau attr expr) subst = do
...
 buildBinding (BiDelta bytes) subst = do
   bts <- buildBytes bytes subst
   Right [BiDelta bts]
-buildBinding (BiMetaLambda meta) (Subst mp) = case Map.lookup meta mp of
-  Just (MvFunction func) -> Right [BiLambda func]
-  _ -> Left (metaMsg meta)
+buildBinding (BiMetaLambda meta) (Subst mp) = case lookupNamedMeta meta mp of
+  Right (MvFunction func) -> Right [BiLambda func]
+  Right _ -> Left (metaMsg meta)
+  Left err -> Left err
```
maxonfjvipon: @sherenkov the same as above

src/Builder.hs
```diff
@@ -130,19 +141,21 @@ buildExpression (ExApplication expr (BiTau battr bexpr)) subst = do
...
-  Just (MvExpression expr scope) ->
+buildExpression (ExMeta meta) (Subst mp) = case lookupNamedMeta meta mp of
+  Right (MvExpression expr scope) ->
     let res = Right (expr, scope)
      in case expr of
           ExFormation bds -> uniqueBindings bds >> res
           _ -> res
-  _ -> Left (metaMsg meta)
+  Right _ -> Left (metaMsg meta)
+  Left err -> Left err
```
maxonfjvipon: @sherenkov the same as above

src/Builder.hs
```diff
@@ -130,19 +141,21 @@ buildExpression (ExApplication expr (BiTau battr bexpr)) subst = do
...
 buildExpression (ExMetaTail expr meta) subst = do
   let (Subst mp) = subst
   (expression, scope) <- buildExpression expr subst
-  case Map.lookup meta mp of
-    Just (MvTail tails) -> Right (buildExpressionWithTails expression tails subst, scope)
-    _ -> Left (metaMsg meta)
+  case lookupNamedMeta meta mp of
+    Right (MvTail tails) -> Right (buildExpressionWithTails expression tails subst, scope)
+    Right _ -> Left (metaMsg meta)
+    Left err -> Left err
```
maxonfjvipon: @sherenkov the same as above

src/Builder.hs
```diff
@@ -40,8 +40,15 @@ data BuildException
   | CouldNotBuildBytes {_bts :: Bytes, _msg :: String}
   deriving (Exception)
 
-metaMsg :: Text -> String
-metaMsg = printf "meta '%s' is either does not exist or refers to an inappropriate term" . T.unpack
+metaMsg :: Maybe Text -> String
+metaMsg Nothing = "anonymous meta variable cannot be referenced (it has no index)"
+metaMsg (Just name) = printf "meta '%s' is either does not exist or refers to an inappropriate term" (T.unpack name)
+
+lookupNamedMeta :: Maybe Text -> Map.Map Text MetaValue -> Either String MetaValue
```
maxonfjvipon: @sherenkov let's rename just to `lookupMeta`

src/Matcher.hs
```diff
@@ -149,24 +155,26 @@ matchExpression' (ExPhiMeet prefix idx expr) (ExPhiMeet prefix' idx' expr') scop
...
 matchBindingExpression _ _ _ = []
 
--- Match expression with deep nested expression(s) matching
-matchExpressionDeep :: Expression -> Expression -> Expression -> [Subst]
+-- Match expression with deep nested expression(s) matching, returning the
+-- matched target subexpression alongside each substitution.
+matchExpressionDeep :: Expression -> Expression -> Expression -> [(Expression, Subst)]
 matchExpressionDeep ptn tgt scope =
-  let matched = matchExpression' ptn tgt scope
+  let here = [(tgt, s) | s <- matchExpression' ptn tgt scope]
```
maxonfjvipon: @sherenkov why do we need a tuple `(tgt, s)` here if result substitution already contain target expression?

src/Rule.hs
```diff
@@ -224,59 +228,10 @@ meetCondition cond (subst : rest) ctx = do
...
-            )
-            (Just subst)
-            extras'
-        | subst <- substs
-        ]
-    logDebug "Extra substitutions have been built"
-    pure (catMaybes res)
-
-matchExpressionWithRule :: Expression -> Y.Rule -> RuleContext -> IO [Subst]
+matchExpressionWithRule :: Expression -> Y.Rule -> RuleContext -> IO [(Expression, Subst)]
```
maxonfjvipon: @sherenkov I don't see a reason behind the return type changing

src/Rule.hs
```diff
@@ -286,22 +241,88 @@ matchExpressionWithRule expr rule ctx =
...
 
-matchProgramWithRule :: Program -> Y.Rule -> RuleContext -> IO [Subst]
+matchProgramWithRule :: Program -> Y.Rule -> RuleContext -> IO [(Expression, Subst)]
 matchProgramWithRule (Program expr) = matchExpressionWithRule expr
+
+-- Variants of meetMaybeCondition/extraSubstitutions that thread the matched
+-- fragment through alongside each substitution. Conditions filter pairs;
+-- `where` extras attach new bindings to the substitution while keeping the
+-- fragment fixed.
+meetMaybeConditionPaired :: Maybe Y.Condition -> [(Expression, Subst)] -> RuleContext -> IO [(Expression, Subst)]
```
maxonfjvipon: @sherenkov also I don't see reason in this 'Paired' suffix in function name

maxonfjvipon (changes requested):
@sherenkov the main concern is about changing return type in `matchProgram` which leads to changing a lot of code in `Rule` and `Builder` module. It's either not well explained/documented or does not make any sense at all. At least for now it looks a bit ugly

src/Builder.hs
sherenkov: Renamed to `lookupMeta` in e59953a.

src/Builder.hs
sherenkov: Collapsed in e59953a — `metaMsg` already produces the right error for both anonymous and named cases.

src/Builder.hs
sherenkov: Done in e59953a.

src/Builder.hs
sherenkov: Done in e59953a.

src/Builder.hs
sherenkov: Done in e59953a.

src/Builder.hs
sherenkov: Done in e59953a.

src/Builder.hs
sherenkov: Done in e59953a.

src/Matcher.hs
sherenkov: The substitution does **not** always contain the matched target — it only contains entries for *named* metas that fired during the match. Counter-examples:

- Pattern `⊥(𝜏 ↦ 𝑒)` (both anonymous): the substitution is empty after a successful match, so we have no way to know which subexpression to hand to the replacer.
- Pattern `⊥.𝜏0`: the substitution contains `{𝜏0 ↦ α}`, not the full `⊥.α` fragment.

Before #218, the rewriter recovered the matched fragment by rebuilding the pattern from the substitution. With anonymous metas (`Maybe Text` in the AST, no entry in the substitution), that round-trip no longer works — the rebuilt pattern would still contain bare `𝜏` and wouldn't be findable in the target expression.

In e59953a I introduced `type Match = (Expression, Subst)` in `Matcher.hs` with a comment explaining exactly this. The signatures now talk in terms of `Match`, which I hope makes the intent clearer.

src/Rule.hs
sherenkov: See the reply on the `matchExpressionDeep` thread for the full reasoning. tl;dr: with anonymous metas there's no way to recover the matched subexpression from the substitution alone, so it must travel alongside it.

In e59953a the type is now `IO [Match]` where `Match = (Expression, Subst)` is defined and documented in `Matcher.hs`.

src/Rule.hs
sherenkov: Dropped in e59953a — `meetMaybeConditionPaired` is now `meetMaybeCondition`, `extraSubstitutionsPaired` is now `extraSubstitutions`, and `meetConditionPaired` is inlined into `meetMaybeCondition` (no name clash with the existing `[Subst]`-version of `meetCondition` that's still exported for tests).

sherenkov:
@maxonfjvipon all review threads addressed in e59953a — see inline replies. PTAL.

src/Matcher.hs
maxonfjvipon: @sherenkov 
> so we have no way to know which subexpression to hand to the replacer.

But why do we need to know it? If we use anonymous meta, it means we don't need the matched expression/attribute/whatever anymore because we're not gonna use it in `result` or `when` or etc.

So what's the reason to know them? What's the reason to keep them?

sherenkov: @maxonfjvipon refactored in bcdb4d3 — the fragment now lives inside `Subst` itself, so `matchProgram`, `matchExpressionWithRule`, `meetMaybeCondition`, and `extraSubstitutions` are back to plain `[Subst]`. The cross-module noise is gone.

Re *why* keep it: it's not the anonymous-meta value (agreed, that's discardable). It's **the subexpression of the program where the rule fired** — the location the rewriter splices the result into. That's needed for every match, even patterns with zero metas (e.g. `Φ.org.eolang.true.eq(α0 ↦ Φ.org.eolang.false)` → result). Before #218, the rewriter recovered it by re-building the pattern from the substitution and asking the replacer to find a node `==` to it. That round-trip relied on every meta in the pattern having an entry. Anonymous metas leave no entry, so the rebuild produces a pattern still containing bare `𝜏`/`𝐵`/`𝑒` and `buildExpressionsThrows` errors on it.

Now the matcher attaches the fragment in `matchExpressionDeep` (the only place that knows it), the Rewriter reads it via `substTarget` once in `buildAndReplace'`, and the Builder/Printer/etc. ignore it — no user-visible surface. PTAL.

sherenkov: And maybe it can speed up rewriting ( #613 )

sherenkov: Well, Claude is not so optimistic on that:
"
  Theoretical scope of the speedup                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                              
  What the PR actually removes from the hot path (in src/Rewriter.hs:88):                                                                                                                                                                                                     
                            
  - Master: buildAndReplace' calls buildExpressionsThrows ptn substs to rebuild the pattern from each Subst, then hands the rebuilt expression to the replacer, which walks the program comparing nodes by ==.                                                                
  - PR: the matcher already attached the matched fragment in matchExpressionDeep (src/Matcher.hs:196), so buildAndReplace' reads it via substTarget and skips the rebuild entirely.                                                                                           
                                                                                                                                                                                                                                                                              
  What it does not change:  
  - matchExpressionDeep still recursively matches every subexpression — same complexity as before.                                                                                                                                                                            
  - replaceExpression / replaceExpressionFast in Replacer.hs still traverse the whole program tree and compare expr == ptn at each node. That structural-equality scan over multi-thousand-line .phi files is almost certainly the dominant cost in #613, and the PR doesn't  
  touch it.                                                                                                                                                                                                                                                                 
  - Slight new cost: tryBuildAndReplaceFast now does a hasAnonExpr walk over pattern and result before taking the fast path. Patterns/results are tiny so it's negligible, but it isn't free.                                                                                 
                                                                                                                                                                                             
  So: expect a small constant-factor improvement proportional to (#matches × pattern size × #rules). Not the order-of-magnitude #613 needs.  
"

maxonfjvipon: @sherenkov The #613 is not in the scope of this task

> That's needed for every match, even patterns with zero metas (e.g. Φ.org.eolang.true.eq(α0 ↦ Φ.org.eolang.false) → result). Before #218, the rewriter recovered it by re-building the pattern from the substitution and asking the replacer to find a node == to it.

Can you show me the place where this subexpression is needed for every match?

sherenkov: You're right, dropping the #613 angle — that wasn't the point.

The place is `Replacer.replaceExpression'` (src/Replacer.hs:46-47):

```haskell
replaceExpression' state@(expr, ptns@(ptn : _ptns), repls@(repl : _repls)) ctx =
  if expr == ptn
    then replaceExpression' (repl expr, _ptns, _repls) ctx
```

The replacer walks the program tree and uses structural equality (`expr == ptn`) to locate the node to splice. Same shape on the fast path — `findAndReplace` at src/Replacer.hs:72 does `` ptn `isPrefixOf` xs `` over binding lists. Either way, *some* concrete subexpression has to be handed in, on every match, regardless of meta count.

On `master`, `buildAndReplace'` synthesised that `ptn` per-match by calling `buildExpressionsThrows ptn substs` (src/Rewriter.hs:86 on master) — i.e. it rebuilt the pattern from the substitution and let the replacer find an `==` node. That round-trip is what breaks once the AST allows `ExMeta Nothing` / `BiMeta Nothing` / `BtMeta Nothing`: anonymous metas leave no entry in the substitution, so the rebuild yields a pattern still containing bare 𝜏 / 𝐵 / 𝑒, which never structurally equals any node in the target program. The rule then silently fails to apply.

So this isn't about *using* the anonymous-meta value (you're right, that's discardable). It's that the pre-#218 design used the substitution as a round-trip key to recover the firing site, and anonymous metas invalidate that round-trip. The fix is to have the matcher — the only place that already knows the firing site — carry it forward. Hence the `Maybe Expression` field inside `Subst`.

maxonfjvipon: @sherenkov now I see. The problem is that:
- when you build substitution you build `substEmpty` for anonymous meta variables
- then when you try to build expression from pattern using substitution so you can use it in the future in rewriter - you can't do it because subst is empty, there's no value by anonymous meta key
- to hack it - you suggest to store original expression in substitution

I don't like the idea, at least for now, because it's not substitution anymore and the design is not such clear as it was before.

I suggest to try this approach. Right now subst is hash map where key is String (meta) and value is an expression/attribute/whatever. 1 to 1 mapping. Anonymous attributes mean that singe anonymous meta attribute may be mapped to several expressions/attributes. For example:
```
pattern: !e.x(~0 -> !e)
expression: Q.w.e.x(~0 -> $.a)
```

Here we use `!e` twice in pattern. In expression this meta matches with two expressions: `Q.w.e` and `$.a`.
I suggest to changes `Subst` signature so it's not a one to one map, but one to many: `String -> [Expression]`. For named metas it should be prohibited to have more than one element in the substitution, but for anonymous ones it's allowed.

With such approach after matching this expression we'll get: `[!e -> [Q.w.e, $.a]]`

Then, when we will need to build expression from pattern to use in rewriter, we simply take the expression from a list in order in which they occur.

But it should be prohibited to use these anonymous metas in building result expression.

sherenkov: Pushed dfc5ae4 — kept Subst as a pure substitution and moved the firing site into `type Match = (Expression, Subst)` returned by the matcher. Subst is back to `newtype Subst (Map Text MetaValue)`; only Rewriter unzips Match at the boundary. Cross-module signatures now talk Match where they previously talked Subst (`matchExpressionWithRule`, `matchProgramWithRule`, `meetMaybeCondition`, `extraSubstitutions`). PTAL.

maxonfjvipon: @sherenkov the idea was to get rid of synthetic Match and change Subst structure to `Map Text [MetaValue]` and don't produce `substEmpty` when matching anonymous metas

sherenkov: Pushed d55d9f7 — adopted your `Map Text [MetaValue]` shape. Anonymous metas accumulate under sentinel `anonKey = T.empty` in pattern walk order; named metas stay as singleton lists with the existing consistency check. The firing site is recovered by desugaring anonymous metas in the pattern to synthetic `_anon_0`/`_anon_1`/... names before passing to the standard Builder — no separate Match type. Had to swap `combineMany` arg order in `matchExpression' (ExDispatch ...)` so anonymous values accumulate in walk order. PTAL.

maxonfjvipon: @sherenkov we've already gone though it, special naming for meta attributes is a dirty hack

maxonfjvipon:
@sherenkov sorry, it lasts a long time and to no avail, closing it

sherenkov:
@maxonfjvipon yes, agree with you, the straight forward approach failed, and I can't tailor it to be good enough.

0crat:
@maxonfjvipon Thanks for the review! You've earned **+19 points** for this contribution: **+12** as a basis for conducting the review, **+12** for reviewing 1245 hits-of-code (capped at +16, so +16 applied), and **-5** for having fewer than 12 comments during the review (only 1 comment). Your running score is [+375](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #687, opened 2026-05-12
bug(#685): add print-path benchmarks for the margin hotspot

sherenkov:
Refs #685.

The existing bench suite covers `parse/phi`, `parse/xmir`, and `rewrite/normalize`, but does not exercise the print pipeline — i.e. the path through `Printer.printProgram'` that runs `withMargin`. That's the path #685 is about, and it has been unmeasured by CI.

This PR adds three cases against the same `native.phi` input the existing benches use:

- `print/sweet/multiline` — default user-facing path (SWEET + UNICODE + MULTILINE + default margin).
- `print/sweet/flat` — same with SINGLELINE; on master, this currently pays the full `withMargin` cost too (the layout is computed and then collapsed), so the bench will catch any future "skip margin for SINGLELINE" win.
- `print/salty/multiline` — canonical phi syntax stress case (SALTY desugars to a larger CST, scaling Margin work proportionally).

Each case wraps the pure `printProgram' prog cfg` in `evaluate (length …)` to force the full result inside the timed window.

## Local numbers

On `native.phi` (24,154 lines), against `master` HEAD:

| Bench | avg |
|---|---|
| parse/phi | 0.56s |
| parse/xmir | 2.77s |
| rewrite/normalize | 0.31s |
| **print/sweet/multiline** | **2.76s** |
| **print/sweet/flat** | **2.49s** |
| **print/salty/multiline** | **8.64s** |

Two things worth flagging:

1. `print/sweet/multiline` is at parity with `parse/xmir` (~2.77s) — printing the parsed program takes as long as parsing the XMIR. It's ~9× slower than `rewrite/normalize`.
2. `print/sweet/flat` is only ~10% faster than the multiline case — empirical confirmation that `--flat` doesn't bypass `withMargin` on master.

GHA-measured numbers will land in `README.md` via the existing benchmark-on-push workflow once this merges.

## Test plan

- [x] `cabal bench --enable-benchmarks` builds and runs locally
- [x] All three new cases produce stable numbers (std dev ~3–7%)
- [ ] CI benchmark job runs successfully against `master` after merge (will produce updated `README.md` PR)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #685, opened 2026-05-11
render appends Text inefficiently on large inputs


sherenkov:
Inspired by #613 (cc @volodya-lombrozo).

Profiling `phino` on a real 31K-line file (PureJavaCrc32.phi) shows
nearly all time and allocation spent inside Text append, driven by
Margin.hs measuring widths by rendering to Text:

    $wappend    Data.Text.Internal   89.0% time   99.1% alloc
    withMargin  Margin                2.6% time    0.1% alloc

Run-level totals for that file: ~42K GC collections, ~191 GB allocated
(~6 MB per source line).

Source sites in src/Margin.hs:

- lengthOf (line 122): `T.length (render renderable)` — renders the
  whole subtree to Text just to count characters.
- Lines 46, 58, 70: `T.length (last (T.lines (render main)))` —
  renders the multi-line `main` to Text, splits into lines, takes the
  last line's length.
- Every WithMargin clause calls `lengthOf` on a `toSingleLine`
  projection of the current node (lines 35, 50–52, 62–64, 74–76, 86,
  107, 112). Parent recursion re-renders the same subtrees.

For a deep tree, each nested node renders to Text to measure, then the
parent renders it again — quadratic-ish in tree depth and consistent
with the `$wappend` 89%/99.1% signature in the profile.

P.S. I do not think it is the reason of #613 itself, the real botteneck there seems to be outside of phino's core rendering.

sherenkov:
More investigation shows that render walks  the  tree  and appends Text fragments inefficiently. 

Text is immutable and append is O(n), so building output by repeated append is O(n²) in the number of fragments.

Blind drafting changes - if switched Render class from Data.Text (strict, O(n²) append) to Data.Text.Lazy (O(n) append via chunked chunks) show significant improvements. 

Results (PureJavaCrc32.phi, 31K lines) are:

Time: 9.73s → 2.79s
Allocation: 191GB → 10.5GB
GCcollections: 42,795 → 3,779

Seems good to investigate further and implement.

sherenkov:
@maxonfjvipon @yegor256 please check if this is decent for proceeding. (And assign to me if appropriate)

Added the profiling run results image on master branch.

Here's the summary:
  - Time: 10.41s
  - Allocation: 191 GB
  - $wappend: 89.7% of time
  - withMargin: 2.3% of time
  
  The bottleneck is Text.append in the render pass.

[...]

PR #688, opened 2026-05-12
bug(#685): gate PRs on benchmark regressions

sherenkov:
Refs #685. **Depends on #687** — the print-path bench cases land first; this gate watches them.

## Summary

- Restructures `.github/workflows/benchmark.yml` into two jobs:
  - `benchmark` — existing master-push behavior, unchanged (still updates `README.md` and opens the rolling auto-PR).
  - `regression-check` — new, runs on every `pull_request`.
- Adds `.github/scripts/bench-compare.py` to parse two bench captures, render a markdown comparison table, and exit 1 on any regression above threshold.

## How the gate works

1. **Same-runner A/B.** On the same GHA runner, the job checks out the PR merge-base, builds, runs the bench; then checks out the PR head, builds, runs the bench again. Comparing two runs from the same minute on the same machine removes most environmental variance.
2. **Frozen fixture.** `benchmark/tmp/native.phi` is generated from the *merge-base's* `phino` binary, then copied aside and restored before the head's bench run. PRs that touch the rewrite pipeline cannot accidentally compare different inputs.
3. **Threshold: +20%.** Above expected runner residual (~10%) plus the bench's own std dev (~3–7%). Conservative for the first iteration; easy to tighten once we see real PR data.
4. **Output via job step summary.** The comparison table is appended to `$GITHUB_STEP_SUMMARY` and shown one click into the Checks tab. The job exit status gates the PR (red on regression).
5. **GHC/Cabal versions hoisted to workflow-level env vars** to dedupe across the two jobs (mirrors the existing `haddock.yml` pattern).

## Why no PR comment

The first iteration of this workflow used `gh api` to maintain a single rolling comment on each PR. That failed with `Resource not accessible by integration` because GitHub forces `GITHUB_TOKEN` to be read-only on fork-sourced `pull_request` runs as a security policy, regardless of the `permissions:` block — and most external contributions arrive via fork. The step-summary approach works uniformly for both same-repo and fork PRs. If comment visibility becomes important later, the safe upgrade is a two-workflow split (`pull_request` saves an artifact, `workflow_run` posts the comment in base-repo context).

## Caveats

- This is the first run of the workflow on real PR infrastructure. Iterating in-PR: actionlint flagged an `export PATH="$(…):$PATH"` form (SC2155), and the fork-PR write-token policy required dropping the comment step — both fixed in the commit history of this branch.
- The bench job currently takes ~2 minutes of measurement per run; the regression job runs the bench twice, so expect ~5–7 minutes of CI wall time on top of build.
- If runner variance turns out to make the gate flaky in practice, two cheap mitigations are available without changing the design: run each side N times and compare medians; or raise the threshold.

## Test plan

- [x] `bench-compare.py` smoke-tested locally with synthetic and real inputs (correctly flags regression, marks improved cases, handles new/removed bench names)
- [x] First `regression-check` run on this PR completes successfully
- [x] Comparison table appears in the Checks → job-step-summary view


maxonfjvipon (changes requested):
@sherenkov why python? Let's try to either use Haskell or bash?

sherenkov:
@maxonfjvipon script rewritten to bash

maxonfjvipon (commented):
A few follow-ups — non-blocking, but worth folding in before merge.

Worth confirming both on this PR's own CI run before merging — this is the only end-to-end smoke test the gate gets before it starts blocking real PRs.

.github/workflows/benchmark.yml
```diff
@@ -69,3 +77,68 @@ jobs:
           delete-branch: true
           title: 'New benchmarking results'
           base: master
+  regression-check:
```
maxonfjvipon: @sherenkov maybe let's move this regression check to a separated CI job because now I see this CI skipped in PR when the only first part of it is actually skipped

maxonfjvipon (changes requested):

.github/scripts/bench-compare.sh
```diff
@@ -0,0 +1,84 @@
...
+# blocks with an `avg: X.YYY us` line per case).
+#
+# THRESHOLD: max allowed positive delta as a fraction, e.g. 0.20 for +20%.
+#
+# Prints a markdown comparison table to stdout, exits 1 if any case present in
+# both runs regressed by more than THRESHOLD.
+set -euo pipefail
+
+if [ "$#" -ne 3 ]; then
+  sed -n '6,15p' "$0" >&2
```
maxonfjvipon: Printing usage via `sed -n '6,15p' "$0"` silently drifts the moment the header comments shift by a line. Worth replacing with an explicit `usage()` function so the message is anchored to code, not line numbers.

.github/scripts/bench-compare.sh
```diff
@@ -0,0 +1,84 @@
...
+          d = (h - b) / b
+          pct = d * 100
+          note = ""
+          if (d > threshold) {
+            note = " **REGRESSION**"
+            regressions++
+          } else if (d < -0.05) {
+            note = " (improved)"
+          }
+          printf "%s\t| `%s` | %.0f | %.0f | %+.1f%%%s |\n", k, k, b, h, pct, note
```
maxonfjvipon: Both `base` and `head` columns format with `%.0f`, rounding to integer microseconds, while the delta column is computed on the un-rounded values. For sub-microsecond cases this makes the displayed numbers fail to reconcile with the percentage (e.g. `0 → 0` with `+42.0%`). Bump to `%.2f` (or scale to ns) so a reader can sanity-check the delta by hand. Same applies to the `%.0f` on lines 49 and 51.

sherenkov:
@maxonfjvipon all three follow-ups addressed in 8ba1e7b:

1. **`benchmark.yml` split** — `regression-check` job moved into its own workflow file (`.github/workflows/regression-check.yml`). `benchmark.yml` is now push-only, so PRs no longer see the "skipped" sibling check.
2. **`bench-compare.sh` usage** — `sed -n '6,15p' "$0"` replaced with an explicit `usage()` heredoc, anchored to code rather than line numbers.
3. **`bench-compare.sh` precision** — base/head columns format with `%.2f` so sub-microsecond cases reconcile with the printed delta.

CI is green end-to-end, including the gate against this PR's own merge-base. Ready for re-review.

.github/workflows/regression-check.yml
```diff
@@ -0,0 +1,80 @@
...
+      - id: setup
+        uses: haskell-actions/setup@v2.11.0
+        with:
+          ghc-version: ${{ env.GHC_VERSION }}
+          cabal-version: ${{ env.CABAL_VERSION }}
+      - uses: actions/setup-java@v5
+        with:
+          distribution: 'temurin'
+          java-version: 21
+      - uses: actions/cache@v5
```
maxonfjvipon, line 35: @sherenkov let's use v6

maxonfjvipon (changes requested):

.github/workflows/regression-check.yml
sherenkov: Done in e61788c. Bumped both `actions/cache@v5` instances in this file plus the matching pair in `benchmark.yml` (the two are post-split siblings — kept them in sync).

maxonfjvipon: @sherenkov seems to be my mistake, I thought it's actions/checkout

sherenkov: @maxonfjvipon reverted that

maxonfjvipon (commented):
@sherenkov in general looks good, one more thing and we're good to merge: when job is finished, let's post a comment on PR with printed table and regression result. The author of comment is GHA bot

sherenkov:
@maxonfjvipon implemented in 25bac7c following the eo pattern (`objectionary/eo/.github/workflows/post-comment.yml`):

- `regression-check.yml` now writes the markdown table to `/tmp/benchmark-comment.md`, the PR number to `/tmp/pr-number`, and uploads both as artifacts (`if: always()` so the artifacts ship even when the gate fails).
- New `post-comment.yml` triggers on `workflow_run: regression-check` completion. It runs in base-repo context with a writable `GITHUB_TOKEN`, downloads the artifacts, and posts/updates a **sticky** comment via `mshick/add-pr-comment@v3` (the action tags its comment with `<!-- add-pr-comment:add-pr-comment -->` and updates the same one on subsequent runs).

This works for fork-origin PRs (which is why we can't just post directly from `regression-check` — its token is read-only on forks).

**One bootstrap quirk worth flagging**: GitHub only registers `workflow_run` triggers from workflow files that exist **on the default branch**. So `post-comment` won't fire on this PR itself — it'll start working on the next PR after #688 merges. Verified the producer half end-to-end on the latest CI run (artifacts `benchmark-comment` and `pr-number` are both uploaded and intact).

maxonfjvipon:
@rultor merge

maxonfjvipon (approved):

0crat:
@maxonfjvipon Hey! Nice work on that review 👍 You snagged **+7 points** this time: started with the standard **+12 base**, but lost **-5** since there were only **7 comments** (we need at least **12** to avoid the penalty). Your running total is now [**+374**](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep those detailed reviews coming to maximize your bonus! Don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) for the full breakdown.

0crat:
@sherenkov **Excellent contribution! 🎉** You've earned **+12 points** (+16 base, -4 for 227 hits-of-code exceeding our 100 limit). [Keep up the great work](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) while staying mindful of code complexity! Your total score is now [+80](https://www.eolang.org/zerocracy/objectionary-vitals.html) - check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #709, opened 2026-05-20
Add splice where-function for multi-position sentinel rewriting

### LINKED ISSUE #708: Add `splice` where-function for multi-position sentinel rewriting
The where-function dispatch in `src/Functions.hs:34-47` exposes `concat`, `sed`, `join`, `size`, `tau`, `random-tau`, `random-string`, `number`, `string`, `sum`, `dataize`, `scope`, and `contextualize`, but it has no primitive that replaces every occurrence of a structured sub-formation inside a captured binding group within a single firing. Multi-emit fusion in [objectionary/hone-maven-plugin#570](https://github.com/objectionary/hone-maven-plugin/issues/570) — specifically the `401c-fuse-cps-auto.phr` rule in that repo — needs exactly that primitive in order to splice an auto body in front of every `Φ.hone.emit` marker in a cps body in one shot (for `flatMap` and verbatim `mapMulti`).

The naive workaround does not work. A non-consuming variant of the rule that keeps the source auto formation in `result` so phino's fixed-point loop fires once per emit marker still matches after each rewrite (the same emit marker is still in the body, the auto formation is still adjacent), so phino re-splices forever and never reaches a fixed point. The existing `join` function concatenates two binding groups but cannot find-and-replace inside one; `sed` operates on string scalars only; neither covers a structured-sub-formation match within a binding group. Without a multi-position primitive — or a "this position was already rewritten" sentinel on the marker side — the rule is inexpressible.

The smallest fix is to add a new dispatch entry next to `_join` at `src/Functions.hs:46` (`buildTerm' "splice" = _splice`) and an implementation that takes three arguments — an input binding group, the sentinel formation expression that identifies positions to rewrite, and a replacement binding group — and returns a new binding group with the replacement spliced in at every position where a binding's value structurally matches the sentinel. A proposed YAML signature for rule files is `{meta: 𝐵-out, function: splice, args: [𝐵-in, 𝜏-sentinel-phi, 𝐵-replacement]}`. The function's contract — "splice the replacement in at every match position in one firing" — is part of the definition; the name does not encode multiplicity (the same way `sed` does not need to say "sed-globally"). It forms a natural pair with `join`: `join` concatenates two binding groups, `splice` weaves a replacement into a binding group at every sentinel position. No existing function needs to change.

yegor256:
This adds a `splice` where-function next to `join` in `src/Functions.hs`. It takes three arguments — an input binding group, a sentinel expression, and a replacement binding group — and returns a new binding group with the replacement inserted in front of every binding whose value is a formation with `φ` equal to the sentinel. Every spliced copy has its τ-labelled attributes renamed via `random-tau` so the resulting binding group has no duplicates, and when nothing matches the sentinel the input is returned unchanged.

The motivation is multi-emit fusion in `hone-maven-plugin` (the `401c-fuse-cps-auto.phr` rule), which needs to splice an auto body in front of every `Φ.hone.emit` marker in a cps body in one rule firing. A naïve workaround that keeps the source formation in `result` so phino's fixed-point loop fires once per emit marker re-matches forever; the existing `join` concatenates two binding groups but cannot find-and-replace inside one, and `sed` only operates on string scalars. `splice` is the binding-group analogue of `sed s/.../.../g`.

The change adds three direct unit tests in `FunctionsSpec.hs` (multi-match splice, zero-match passthrough, uniqueness across many positions) and one end-to-end YAML pack in `test-resources/rewriter-packs/custom/splice-counts-positions.yaml` that drives `splice` through the rewriter and counts the spliced bindings via `size`. The README gets a short entry under the where-functions list. No existing function changes.

Closes #708

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,6 +331,13 @@ Here's list of functions that are supported for extensions:
 * `join` - accepts list of bindings and returns list of joined bindings. Duplicated
   `ρ`, `Δ` and `λ` attributes are ignored, all other duplicated attributes are replaced
   with unique attributes using `random-tau` function.
+* `splice` - accepts three arguments: input bindings `𝐵-in`, a sentinel expression,
+  and replacement bindings `𝐵-rep`. Returns a new binding group where `𝐵-rep`
+  is inserted in front of every binding in `𝐵-in` whose value is a formation
+  with `φ` equal to the sentinel. Every spliced copy of `𝐵-rep` has its
+  `τ`-labelled attributes renamed via `random-tau` so the resulting binding
+  group has no duplicates. When no binding matches the sentinel the output
+  equals the input unchanged.
 
 ## Meta variables
 
--- a/README.md
+++ b/README.md
@@ -335,9 +335,11 @@ Here's list of functions that are supported for extensions:
   and replacement bindings `𝐵-rep`. Returns a new binding group where `𝐵-rep`
   is inserted in front of every binding in `𝐵-in` whose value is a formation
   with `φ` equal to the sentinel. Every spliced copy of `𝐵-rep` has its
-  `τ`-labelled attributes renamed via `random-tau` so the resulting binding
-  group has no duplicates. When no binding matches the sentinel the output
-  equals the input unchanged.
+  `τ`-labelled attributes renamed via `random-tau`, so duplicates introduced
+  by repeated insertion are avoided only for those `τ` attributes. If `𝐵-rep`
+  contains non-`τ` special attributes such as `ρ`, `Δ`, `λ`, or `φ`, splicing
+  at multiple positions may still produce duplicates. When no binding matches
+  the sentinel the output equals the input unchanged.
 
 ## Meta variables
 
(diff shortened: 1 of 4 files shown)


yegor256:
@maxonfjvipon please, take a look, we need this new feature

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@yegor256 Thanks for the contribution! You've earned **+12 points** for this: **+16** as a basis, **+11.6** for your 232 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (0.05 × 232, capped at 16), **-4** for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), and **-11.6** for review comments above 4 (as per our policy). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming while considering smaller, more focused contributions to maximize your points! 🎯 Your running score is [+591](https://www.eolang.org/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

0crat:
@maxonfjvipon **Great work on the review!** 🎯 You've earned **+4 points** (+12 base, -8 for missing comments). Your total score is now [+291](https://www.eolang.org/zerocracy/objectionary-vitals.html). Remember to add comments in future reviews for maximum points, and check your [Zerocracy account](https://www.zerocracy.com/) for updates!

PR #715, opened 2026-05-22
fix(#714): normalize formations after each rewrite step

### LINKED ISSUE #714: Combined-mode rewriting drops match when prior rule's result contains an empty `⟦ ρ ↦ ∅ ⟧` binding group
On phino `0.0.0.69`, when two rules are passed in a single `phino rewrite --rule A.phr --rule B.phr` invocation, rule **B fails to match a pattern that rule A has just produced**, even though rule B matches that same pattern when applied separately on rule A's output (i.e. `phino rewrite --rule A.phr ... > tmp && phino rewrite --rule B.phr tmp`). The match failure is specifically triggered when rule A's `result` contains an empty `⟦ ρ ↦ ∅ ⟧` binding group nested inside the produced formation; remove the `captures ↦ ⟦ ρ ↦ ∅ ⟧` line from A's result and the combined run starts working again. I have not isolated the exact branch in `src/Rewriter.hs`, but the symptom appears in `_rewrite` (around `src/Rewriter.hs:140-180`) where `matchExpressionWithRule` is invoked on the program after the previous rule has been applied — somewhere on that path the empty `⟦ ρ ↦ ∅ ⟧` produced by A is presented to the matcher in a form that does not match B's pattern but does match if the program is round-tripped through `printExpression` and parsed again (which is essentially what the sequential invocation does).

Minimal reproducer (copy/paste each file, then run the two commands at the bottom):

`input.phi`:
```
{⟦
  φ ↦ Φ.jeo.class,
  m ↦ ⟦
    φ ↦ Φ.jeo.method,
    body ↦ ⟦
      site ↦ ⟦ φ ↦ Φ.foo.src ⟧
    ⟧
  ⟧
⟧}
```

`A.phr`:
```
---
name: A
pattern: |
  ⟦ φ ↦ Φ.jeo.class, 𝐵-class-head,
    𝜏-m ↦ ⟦ φ ↦ Φ.jeo.method, 𝐵-mh,
      body ↦ ⟦ 𝐵-bh, 𝜏-site ↦ ⟦ φ ↦ Φ.foo.src ⟧, 𝐵-bt ⟧, 𝐵-mt ⟧,
    𝐵-class-tail ⟧
result: |
  ⟦ φ ↦ Φ.jeo.class, 𝐵-class-head,
    𝜏-m ↦ ⟦ φ ↦ Φ.jeo.method, 𝐵-mh,
      body ↦ ⟦ 𝐵-bh,
        𝜏-site ↦ ⟦ φ ↦ Φ.foo.distill, captures ↦ ⟦ ρ ↦ ∅ ⟧,
          body ↦ ⟦ 𝜏-emit ↦ ⟦ φ ↦ Φ.foo.emit ⟧ ⟧ ⟧,
        𝐵-bt ⟧, 𝐵-mt ⟧,
    𝐵-class-tail ⟧
where:
  - meta: 𝜏-emit
    function: random-tau
    args: []
```

`B.phr`:
```
---
name: B
pattern: |
  ⟦ φ ↦ Φ.foo.distill,
    body ↦ ⟦ 𝐵-pre, 𝜏-emit ↦ ⟦ φ ↦ Φ.foo.emit ⟧, 𝐵-post ⟧ ⟧
result: |
  ⟦ φ ↦ Φ.foo.lowered,
    body ↦ ⟦ 𝐵-pre, lo ↦ ⟦ φ ↦ Φ.foo.lo ⟧, 𝐵-post ⟧ ⟧
```

`phino rewrite --max-cycles 5 --max-depth 50 --sweet --rule A.phr input.phi | phino rewrite --sweet --rule B.phr -` produces a `Φ.foo.lowered` formation (expected); `phino rewrite --max-cycles 5 --max-depth 50 --sweet --rule A.phr --rule B.phr input.phi` leaves a `Φ.foo.distill` with an unlowered `Φ.foo.emit` inside (actual). Debug output shows rule B was attempted in cycle 0 right after A fired and was reported as not matched. The bug disappears the moment `captures ↦ ⟦ ρ ↦ ∅ ⟧` is dropped from A's `result` (and the corresponding entry is dropped from B's pattern, which need not contain it).

A targeted fix is likely to live around `tryBuildAndReplaceFast` / `replaceExpressionFast` in `src/Rewriter.hs` and `src/Replacer.hs` — the fast-replacing path presumably preserves an internal representation of the empty `⟦ ρ ↦ ∅ ⟧` group that is not equivalent to the form the matcher expects on the next rule; either teach the fast path to normalize the produced formation the same way the parser does, or fall through to the slower `replaceExpression` whenever the produced result introduces a sub-formation containing only `ρ ↦ ∅`.

kreinba:
## Summary

Fixes #714 — when two rewrite rules are applied sequentially in a single `phino` command, the second rule fails to match an empty formation produced by the first.

## Root cause

The parser always enforces the invariant `withVoidRho` (every formation carries `ρ ↦ ∅`). The Rewriter did not. The fast replacement path (`replaceBindingsFast` → `findAndReplace`) can remove all bindings from a formation — including `BiVoid AtRho` — leaving `ExFormation []`. Both representations print as `⟦⟧`, but `ExFormation [] /= ExFormation [BiVoid AtRho]` structurally, so a follow-up rule whose pattern was parsed (and therefore normalized) never matches.

## Fix

Apply a new helper `normalizeFormations` (a full-tree traversal of `withVoidRho`) to the result of `tryBuildAndReplaceFast` in `Rewriter.hs`, before any equality check or loop detection. This restores the invariant without touching `Builder`/`Replacer` internals — those operate on partial binding lists (fast-path "middles") which must remain un-normalized, otherwise patterns like `[[ !B1, !a -> ?, !B2 ]]` would only match when the void binding is immediately adjacent to `ρ`.

## Test plan

- [x] Added regression test `test-resources/rewriter-packs/custom/sequential-rules-bug-714.yaml` that fails on `master` and passes with the fix:
  - rule A `[[ !B1, !a -> ?, !B2 ]] => [[ !B1, !B2 ]]` (fast path) removes every void binding including `ρ`
  - rule B `[[ ]] => [[ x -> Q ]]` must then match the empty formation
- [ ] `make test` (run by CI)
- [ ] `make hlint`
- [ ] `make fourmolu`

## Files changed

- `src/Misc.hs` — add and export `normalizeFormations`
- `src/Rewriter.hs` — import and apply `normalizeFormations` after each rule application
- `test-resources/rewriter-packs/custom/sequential-rules-bug-714.yaml` — regression test

### DIFF
--- a/src/Misc.hs
+++ b/src/Misc.hs
@@ -16,6 +16,7 @@ module Misc
   , btsToStr
   , btsToNum
   , withVoidRho
+  , normalizeFormations
   , allPathsIn
   , ensuredFile
   , shuffle
@@ -198,6 +199,19 @@ withVoidRho bds = withVoidRho' bds False
         BiVoid AtRho -> bd : withVoidRho' rest True
         BiTau AtRho _ -> bd : withVoidRho' rest True
         _ -> bd : withVoidRho' rest hasRho
+
+-- Recursively ensure all formations have a BiVoid AtRho binding (ρ ↦ ∅).
+-- Fixes in-memory ExFormation [] to ExFormation [BiVoid AtRho] after rewriting,
+-- keeping the invariant that the parser enforces via withVoidRho.
+normalizeFormations :: Expression -> Expression
+normalizeFormations (ExFormation bds) = ExFormation (withVoidRho (map normalizeFormationsB bds))
+normalizeFormations (ExDispatch e a) = ExDispatch (normalizeFormations e) a
+normalizeFormations (ExApplication e b) = ExApplication (normalizeFormations e) (normalizeFormationsB b)
+normalizeFormations e = e
+
+normalizeFormationsB :: Binding -> Binding
+normalizeFormationsB (BiTau a e) = BiTau a (normalizeFormations e)
+normalizeFormationsB b = b
 
 -- Transform dispatch to list of attributes
 -- >>> fqnToAttrs (ExDispatch (ExDispatch (ExDispatch ExGlobal (AtLabel "org")) (AtLabel "eolang")) (AtLabel "number"))
--- a/src/Misc.hs
+++ b/src/Misc.hs
@@ -16,7 +16,7 @@ module Misc
   , btsToStr
   , btsToNum
   , withVoidRho
-  , normalizeFormations
+  , recoverFormations
   , allPathsIn
   , ensuredFile
   , shuffle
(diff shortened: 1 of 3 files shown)


kreinba:
@yegor256 please, check this one

yegor256:
@maxonfjvipon WDYT?

maxonfjvipon (changes requested):
@kreinba let's rename the functions to `recoverFormations` and `recoverFormations'`. Also let's use more meaningful names for function arguments, not single char

kreinba:
@maxonfjvipon I did what you suggested, please review again

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Hey there! 🎉 Nice work on the review - you've earned **+4 points** for this one: **+12** as the baseline reward, but **-8** since no comments were posted (we deduct 8 points when there are absolutely no comments to help improve the code). Your running score is now [**+227**](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep it up! Next time, try dropping a few comments during review to boost your points even more. Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 💪

0crat:
@kreinba Great work on your contribution! 🎉 You've earned **+16 points** and your running score is now [+16](https://www.eolang.org/zerocracy/objectionary-vitals.html). Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the quality contributions coming, and remember to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #722, opened 2026-05-24
Add graft where-function for marker-replacing splice

### LINKED ISSUE #721: graft: new function as a marker-replacing splice variant with automatic binding-key renaming
The `splice` where-function at `src/Functions.hs:273-282` does insert-before semantics: at every binding that matches the sentinel it inserts a renamed copy of the replacement group and keeps the original marker (`pure (fresh ++ bd : tail')`). A sibling mode that *replaces* the marker — `pure (fresh ++ tail')` — would unlock a class of fusion rules that today are inexpressible.

Concrete use case from hone-maven-plugin (objectionary/hone-maven-plugin#570). The `401d-fuse-cps-cps.phr` rule needs to merge two adjacent `Φ.hone.distill` pragmas both with `emit-shape ↦ "cps"`, by walking the first body's binding group and, at every `Φ.hone.emit` marker, replacing it with the second body's opcodes (so the second body consumes the item left on stack by the first body's emit-time machinery and drives its own emits). Insert-before is wrong here because the original marker would fire afterward with no item on stack. Hand-writing a concat-three-binding-groups rule fails too: the 3xx-fold rules emit synthetic keys `i1, i2, i3, …` in every cps body, so concatenating two cps bodies produces `Duplicated attribute 'i3' found …` and phino aborts. The renaming logic already inside `_splice` (`renamed` at `src/Functions.hs:289-300`) is exactly what's needed for the replacement side too.

Suggested shape: either a new function `splice-replace` taking `(input-bindings, sentinel, replacement-bindings)` with the same renaming guarantees as `splice`, or a fourth-argument mode flag on `splice` (e.g. `\"replace-with-rename\"` as a string literal). The smallest change is to factor the recursion in `splice` so the at-match branch can call `pure (fresh ++ tail')` instead of `pure (fresh ++ bd : tail')` when the new mode is active, reusing `renamed` and `validateRep` verbatim.

yegor256:
This adds a new `graft` where-function next to `splice` in `src/Functions.hs`. It accepts the same three arguments — input bindings, sentinel expression, replacement bindings — and applies the same `τ`-label renaming via `random-tau`, but at every sentinel match it puts the renamed copy of the replacement group in place of the marker instead of in front of it.

The existing `splice` is the right tool when the marker must survive (so it fires after the inserted body), but there are fusion rules where the marker must not survive — for example merging two cps bodies that each end in `Φ.hone.emit`, where keeping the original marker would emit a second item against an empty stack. Hand-rolling the marker-drop on top of `splice` does not work because the replacement body may carry its own emits, so a follow-up pattern cannot tell the original marker apart from the new ones. `graft` keeps the operation atomic and the rule self-documenting.

The recursion in `splice` is factored into a shared helper that takes a `keepMarker` flag, so the only behavioural difference is the one branch that drops `bd` from the spliced output. Validation, sentinel matching, fresh `τ`-renaming, and the avoidance set are reused verbatim. README is updated, and tests cover the happy path, the no-match identity case, the all-unique invariant across many positions, and the fail-fast guard on non-label replacement bindings; a YAML rewriter pack exercises `graft` end-to-end through the rule engine.

Closes #721

### DIFF
--- a/README.md
+++ b/README.md
@@ -342,6 +342,13 @@ Here's list of functions that are supported for extensions:
   because such bindings cannot be renamed and would produce duplicates
   when spliced at more than one position. When no binding matches the
   sentinel, the output equals the input unchanged.
+* `graft` - same call shape as `splice` (`𝐵-in`, sentinel, `𝐵-rep`) and the
+  same renaming guarantees, but every matched sentinel binding is replaced by
+  the renamed copy of `𝐵-rep` instead of preserved in place. Use it when the
+  marker must not survive the substitution — for example, when fusing two
+  bodies that both produce the same effect and the original marker would fire
+  a second time. When no binding matches the sentinel, the output equals the
+  input unchanged.
 
 ## Meta variables
 
--- a/src/Functions.hs
+++ b/src/Functions.hs
@@ -45,6 +45,7 @@ buildTerm' "number" = _number
 buildTerm' "sum" = _sum
 buildTerm' "join" = _join
 buildTerm' "splice" = _splice
+buildTerm' "graft" = _graft
 buildTerm' func = _unsupported func
 
 argToBytes :: Y.ExtraArgument -> Subst -> IO Bytes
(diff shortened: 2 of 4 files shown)


yegor256:
@maxonfjvipon WDYT?

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the review! You've earned **+4 points** for this contribution: +12 as a basis, -8 for absolutely no comments posted during the review. Your running score is [+207](https://www.eolang.org/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

0crat:
@yegor256 Thanks for the contribution! You've earned **+12 points** for this: **+16** as a basis, **+9.2** for your 184 hits-of-code (184 × 0.05), but **-4** deducted since you exceeded 100 hits-of-code, and **-8** deducted for no code review, totaling 12 points after applying our [policy caps](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Your running score is [+708](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too! 🚀

PR #729, opened 2026-05-27
Make random-tau scan the document instead of requiring avoid args (#728)

### LINKED ISSUE #728: `random-tau` should scan the document for taken names instead of requiring them as args
`_randomTau` in `src/Functions.hs:77-99` only avoids names that the caller passes in explicitly: it builds the `attrs` list from the supplied args and loops in `randomTau` (line 96-99) until the random `a🌵%d` label is not in that list. Every call site — `Functions.hs:249`, `Functions.hs:299`, `Functions.hs:304`, `Dataize.hs:212` — therefore has to hand the function the surrounding names it should steer clear of.

That contract is easy to get wrong. If a rule author forgets to list a name that is actually live in the expression, `random-tau` can hand back a label that already exists, and nothing complains at rewrite time. In `objectionary/hone-maven-plugin` the collision only surfaces much later as a JVM `VerifyError` once `jeo` assembles the bytecode, which is a long way from the rule that caused it. The stream-fusion rules there (`streams/3xx/307-distinct-to-distill.phr`, `streams/5xx/502-*`, `streams/5xx/521-*`) all pass long lists of opcode names into `random-tau` purely as a collision list, which clutters the `where` blocks for no semantic reason.

The fix is to make uniqueness the engine's job rather than the caller's: thread the set of attribute labels present in the document being rewritten into the build context, scan it once at the start of the run, and seed `randomTau`'s avoid-set from it so `random-tau` needs no arguments at all. To keep that O(1) per call rather than rescanning, hold a monotonic cursor alongside the set and advance it past taken indices, adding each minted name back to the set so later firings (the fixpoint loop re-applies rules across cycles) never reuse one. A welcome side effect: if the minted names are sequential rather than random, the output becomes deterministic, which would let downstream rule test packs assert exact rewritten names instead of matching with metavariables. Two things to weigh: this removes the args form, so it is a breaking change for any current args-based caller and may deserve a deprecation window; and the up-front scan only rules out collisions with names that already exist — it cannot stop a rule whose own `result` introduces a literal tau that later matches a generated one, but that hazard exists today and is independent of this change.

yegor256:
## Summary

Fixes #728. Uniqueness of generated `𝜏`-labels is now the engine's job rather than the caller's. Previously every `random-tau` call site had to hand over a list of surrounding names to avoid; forgetting one could silently mint a colliding label.

- New `Tau` module holds a document-wide avoid-set plus a **monotonic cursor**. `seedTaus` scans the program **once** at the start of a run (in `runRewrite`/`runDataize`/`runMatch`) and records every attribute label already in use. `freshTau` mints the next free `a🌵<n>`, advances the cursor past taken indices (O(1) amortised, no rescanning), and adds each minted name back so the fixpoint loop never reuses one.
- Sequential naming makes the rewritten output **deterministic**, so downstream rule test packs can assert exact names instead of matching with metavariables.
- `random-tau` now takes no meaningful arguments. Arguments are still accepted but ignored, so existing args-based callers (e.g. the stream-fusion rules in `hone-maven-plugin`) keep working — their long collision lists simply become dead weight that can be deleted.
- Internal callers updated: `join` runs in `IO` (dropping an `unsafePerformIO`), `splice`/`graft` no longer build an `avoid` list, and the synthetic binding in `Dataize._dataize` no longer passes the surrounding bindings.

## Design notes / for discussion

- Per @maxonfjvipon's comment, uniqueness is strictly only needed within a single formation. This PR takes the whole-document approach @yegor256 proposed; whole-program uniqueness is a safe superset of per-formation uniqueness, at the cost of a slightly larger avoid-set. Happy to narrow the scope if preferred.
- The up-front scan only rules out collisions with names that already exist; it cannot stop a rule whose own `result` introduces a literal tau that later matches a generated one — but that hazard exists today and is independent of this change.
- Performance: the scan happens once per run during a single AST traversal; minting is O(1) amortised via the cursor. Worth confirming with `make bench`.

## Test plan

- [ ] `make test` — including new `TauSpec` (sequential minting, cursor reset on reseed, skipping taken names)
- [ ] `make hlint`
- [ ] `make fourmolu`
- [ ] Confirm `splice`/`graft` rewriter packs still produce the expected structural counts
- [ ] `make bench` to check the up-front scan does not regress performance

https://claude.ai/code/session_01Si8fmP3oXSotxi4Y14chyX

---
_Generated by [Claude Code](https://claude.ai/code/session_01Si8fmP3oXSotxi4Y14chyX)_


yegor256:
@maxonfjvipon take a look

maxonfjvipon (commented):
@yegor256 let's write a several test yaml packs with new random tau usages across the whole program

yegor256:
@maxonfjvipon better now?

GHX5T-SOL (approved):
Current head `eaa025ecd68b861418c1f03f2359315bbb881a64` looks good to me.

I checked the implementation against #728 and the follow-up discussion: `random-tau` now uses a seeded document avoid-set, legacy non-byte args remain accepted, byte args still fail, and the added rewriter packs cover deterministic minting plus skipping names already present elsewhere in the program. The current hosted matrix is green across cabal/stack, hlint, fourmolu, coverage, resources, reuse, and regression checks. I also ran local diff hygiene and a diff-only secret scan; both passed.

Local note: this machine does not have `cabal`, so I did not claim a local `make test` run; the approval is based on source review plus the green GitHub Actions matrix.

0crat:
@ghx5t-sol Hey! Nice work on that review! 🎉 You snagged **+7 points** this time: started with the standard **+12** base reward, but lost **-5** since we only saw **5** comments (the policy kicks in a penalty when there's fewer than **12**). Your running score is now [+11](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep those detailed reviews coming to maximize your bonus potential, and don't forget to peek at your [Zerocracy account](https://www.zerocracy.com/) for the latest updates!

src/Functions.hs
```diff
@@ -74,29 +73,18 @@ _scope [Y.ArgExpression expr] subst = do
...
+-- Uniqueness is the engine's job: 'freshTau' draws from the document-wide
+-- avoid-set seeded at the start of the run, so no collision list is needed.
+-- Legacy attribute/binding arguments are accepted but ignored; bytes are
+-- still rejected so rule mistakes are not silently accepted.
 _randomTau :: BuildTermMethod
-_randomTau args subst = do
-  attrs <- argsToAttrs args
-  tau <- randomTau attrs
-  pure (TeAttribute (AtLabel (T.pack tau)))
+_randomTau args _ = do
```
maxonfjvipon: @yegor256 now `_randomTau` ignores all the arguments. Let's make it `fail-fast` - 0 arguments required.

maxonfjvipon (changes requested):

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@yegor256 Great work on your contribution! 🎉 You've earned **+12 points** following our policy: **+16** as the base reward, **+12.7** for your 254 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (capped at +16), but **-4** deducted since you exceeded 100 hits-of-code - remember, smaller contributions often mean better quality as explained [here](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Your running score is [+846](https://www.eolang.org/zerocracy/objectionary-vitals.html), and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

PR #745, opened 2026-06-03
Remove redundant scope/contextualize from copy rule, rename xi to xi-free

### LINKED ISSUE #744: Redundant `scope` and `contextualize` functions in `copy` rule and `phino` due to `xi` restriction
Since we have `xi` rule in `copy` rule - it allows us not to end up with endless rewriting
But this restriction makes `scope` and `contextualize` functions usage in `copy` rules meaningless, we can simply remove them and nothing's changed
So what we need to do:
1. remove `scope` and `contextualize` from `copy` rule
2. delete `scope` function from `phino`
3. rename `xi` to `xi-free`
4. fix the tests, add printing `xi-free` in `explain` command

maxonfjvipon:
Closes #744.

Since the `xi` (now `xi-free`) condition in the `copy` rule already prevents endless rewriting, the `scope` and `contextualize` extensions in that rule were redundant — substituting the matched expression directly produces the same result.

### Rule / condition changes
- **copy rule**: drop the `scope`/`contextualize` `where`-clause; substitute the expression directly into the binding. With `𝑒2`/`𝑒3` gone there is a single expression meta, so it is now the unindexed `𝑒` (matching `dc`/`miss`/`alpha`)
- **scope function**: deleted from `phino` (`Functions`, `Render`, README). `contextualize` is kept — it is still used by the `dot` rule
- **xi → xi-free**: renamed the condition across the parser, YAML schema, and rule evaluation for clarity (it checks an expression is "free of ξ outside the formation")
- **explain**: `xi-free` is now printed as `\xifree{ … }` instead of being stripped from the output

The `copy` rule now explains as:

```
\trrule{copy}
  { [[ B_1, \tau -> ?, B_2 ]] ( \tau -> e ) }
  { [[ B_1, \tau -> e, B_2 ]] }
  { if $ \xifree{ e } \;\text{and}\; \isnormal{ e } $ }
  { }
```

### Follow-up: drop dead scope collection from matching
With the `scope` function gone, the "scope" (closest formation) that the matcher collected was dead weight: it never affected matching (`combine` ignored it) and every reader discarded it. So the second commit removes it:
- `MvExpression` drops its scope field (`Expression Expression` → `Expression`)
- the `scope` parameter threaded through the matcher functions is gone
- `buildExpression`/`buildExpressionThrows` return `Expression` instead of a tuple
- `defaultScope` is deleted and `combine'` simplified

Note: the `explain` LaTeX now references a `\xifree` macro that needs defining in the paper preamble (parallel to `\isnormal`).

Verification: `make test` (1038 examples, 0 failures), `make hlint` (no hints), and `make fourmolu` all pass.


maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Hey! Nice work on your contribution! 🎉 You've scored **+4 points** this time: started with **+16** as the base, but had to deduct **-4** for going over 100 hits-of-code (you hit 488) and another **-8** for crossing the 400 threshold. The policy rewards quality over quantity, so maybe try breaking larger contributions into smaller, focused chunks next time - you'll likely earn more points that way! Your running score is [+67](https://www.eolang.org/zerocracy/objectionary-vitals.html), so [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the contributions flowing and don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too!

PR #751, opened 2026-06-03
bug(#749): count alpha index over domain excluding assets

### LINKED ISSUE #749: Positional `~i` (alpha) counts Δ/λ assets in the index, contradicting the paper's ordinal/Domain definitions
## Summary

When resolving a positional argument `~i` (the `alpha` / `Ralpha` rule), `phino` counts **`Δ`/`λ` assets** in the index. The paper defines the index via `ordinal(τ, b̄)`, where the domain `b̄` **excludes assets** (Def. Domain). So `phino`'s positional index is off by the number of assets preceding the target attribute.

## Reproduction (phino 0.0.0.71)

```console
$ echo '{[[ L> Fn, x -> ? ]](~0 -> Q)}' | phino rewrite --normalize --flat
Φ ↦ ⟦ λ ⤍ Fn, x ↦ ∅, ρ ↦ ∅ ⟧( α0 ↦ Φ )      # STUCK: ~0 reduced nothing

$ echo '{[[ L> Fn, x -> ? ]](~1 -> Q)}' | phino rewrite --normalize --flat
Φ ↦ ⟦ λ ⤍ Fn, x ↦ Φ, ρ ↦ ∅ ⟧                 # ~1 renamed to x, then copy filled it
```

Same with a `Δ`-asset in front:

```console
$ echo '{[[ D> 42-, x -> ? ]](~1 -> Q)}' | phino rewrite --normalize --flat
Φ ↦ ⟦ Δ ⤍ 42-, x ↦ Φ, ρ ↦ ∅ ⟧
```

Baseline without an asset (for contrast — here `~0` correctly hits `x`):

```console
$ echo '{[[ x -> ? ]](~0 -> Q)}' | phino rewrite --normalize --flat
Φ ↦ ⟦ x ↦ Φ, ρ ↦ ∅ ⟧
```

So `phino` treats `x` in `⟦ L> Fn, x ↦ ∅ ⟧` as ordinal **1**, counting the `λ`-asset at position 0.

## Expected (per the paper)

The paper's own example table (`foundations.tex`, `tab:ordinals`, row 3) states:

> `⟦ L> Fn, x ↦ ξ.k ⟧`  →  ordinal of `x` = **0**

because:

- **Def. Domain** — the domain `b̄` is "a sequence of all attributes of `b`, *excluding assets*"; "Assets `D` and `L` do not belong to the domain."
- **Def. Ordinal** — an attribute's ordinal is its position *in `b̄`*, starting from zero.
- **The `alpha` application** — in `b(αᵢ ↦ e)`, `e` must attach to the `τ` for which `ordinal(τ, b̄) = i`.

So `~0` should target `x` (ordinal 0), and `~1` should be out of range. `phino` does the opposite.

## Root cause

`phino` computes the positional index over the **raw binding list** (assets included) rather than over the asset-excluding domain `b̄`. In the generated reduction figure the `alpha` condition is even rendered as `index(τ₂) = |B₁|` (raw count of bindings before the focus), which matches the code but not `foundations.tex`'s `ordinal(τ, b̄)`.

## Note: the paper is internally inconsistent on this point

- `foundations.tex` (Def. Domain / Def. Ordinal / `tab:ordinals`) → **skip assets**.
- The generated Fig. 4 (`phino explain --normalize`) → **count assets** (`|B₁|`).

Fixing this means picking one convention and aligning the other:

- If **skip assets** (the foundations definitions) is canonical, the fix is in the matcher: index over `b̄`, not the raw list.
- If **count assets** is canonical, then `foundations.tex`'s Domain / Ordinal / `tab:ordinals` need updating, and the prose's `ordinal(τ, b̄)` wording should change.

## Impact

Affects which void slot a positional `~i` renames whenever a formation has `Δ`/`λ` assets before the target attribute. It does not affect confluence (it is a *which-slot* question, not soundness), but it is a faithfulness divergence between the implementation and the paper's stated definitions.

maxonfjvipon:
Closes #749.

## Problem

When resolving a positional argument `~i` (the `alpha` rule), `phino` compared `index(𝜏2)` against the **raw** count of bindings before the void slot, which counts `Δ`/`λ` **assets**. The paper defines the index via `ordinal(𝜏, b̄)` over the domain `b̄`, which **excludes** assets (Def. Domain). So `~0` got stuck whenever a formation had an asset before the target attribute:

```console
$ echo '{[[ L> Fn, x -> ? ]](~0 -> Q)}' | phino rewrite --normalize --flat
Φ ↦ ⟦ λ ⤍ Fn, x ↦ ∅, ρ ↦ ∅ ⟧( α0 ↦ Φ )      # STUCK
```

## Fix

`length` stays an honest "count all bindings" operation. A dedicated `domain` number is introduced that counts only non-asset bindings (the domain `b̄`). The `alpha` rule now uses `domain: 𝐵1` instead of `length: 𝐵1`.

```console
$ echo '{[[ L> Fn, x -> ? ]](~0 -> Q)}' | phino rewrite --normalize --flat
Φ ↦ ⟦ λ ⤍ Fn, x ↦ Φ, ρ ↦ ∅ ⟧                 # ~0 hits x
```

`phino explain --normalize` now renders the condition as `index(𝜏2) = |b̄1|` (`\vert \overline{ B_1 } \vert`), aligning the generated figure with the paper's Domain / Ordinal definitions.

## Tests

- New rewriter packs: `~0` skips a `λ`/`Δ` asset to hit `x` (ordinal 0); `~1` targets `ρ` (ordinal 1); `~2` is genuinely out of range.
- New `domain(...)` condition-DSL parse test.
- Updated the `explain` LaTeX golden for the `alpha` rule.

`make test` (1046 examples), `make hlint`, and `make fourmolu` all pass.


maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the contribution! You've earned **+16 points** for providing a solid code review, which demonstrates your commitment to quality 🎯 Your contribution perfectly aligns with our policy of rewarding meaningful engagement in the review process. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+95](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #764, opened 2026-06-06
Execution-driven dataization and morphing (Phase 1)

maxonfjvipon:
**Draft — Phase 1 of #760 (execution-driven dataization & morphing).** Follow-up to #761 (Phase 0, merged).

Goal: make `phino` actually dataize/morph by *interpreting* `resources/morphing.yaml` and `resources/dataization.yaml`, replacing the hand-written `morph`/`dataize'` in `Dataize.hs` — so the rules phino runs are literally the ones it explains (and the paper prints).

## Plan

1. **Guard first (this commit)** — characterization test pinning the exact ordered reduction labels for representative programs, so the rewrite can be proven behavior-preserving.
2. **Context-dependent `where`-functions** — `normalize` (run the rewriter to a fixpoint), `lambda` (resolve the λ binding + call the atom), `global` (universe/Q lookup). These need `DataizeContext`, which the current `BuildTermFunc` lacks — the main plumbing piece.
3. **Generic small-step driver** interpreting `[MorphRule]`/`[DataizeRule]` (`match → where → when → then`), replacing `morph`/`dataize'`.
4. **Per-relation allow-list validation** — now meaningful, since the M/D-only functions become registered (a `rewrite --rule` must not reach `morph`/`global`/`lambda`).
5. **Equivalence test** — YAML-driven `dataize`/`morph` ≡ the current executor over the corpus.

Behavior must stay byte- and sequence-identical throughout; the guard from step 1 enforces it.

## Status

- [x] Characterization guard (reduction label sequence)
- [x] context-dependent where-functions (`mdBuildTerm`: normalize/lambda/global)
- [x] generic driver (morphByRules + dataizeByRules, equivalent to morph/dataize')
- [x] allow-list validation (rewrite rules cannot use dataization-only functions)
- [x] swap executor (dataize now runs the YAML rules; corpus + pinned sequence confirm equivalence)

`make test` / `hlint` / `fourmolu` green so far. All five Phase 1 items are complete; full suite (`-Werror`), hlint and fourmolu are green. Ready for review.

Issue #760, opened 2026-06-04
Dataization and Morphing Rules Not Formalized in YAML


maxonfjvipon:
In the current version of [calculus-paper](https://github.com/objectionary/calculus-paper) we have 11 normalization rules which are generated by `phino explain` from yaml files. It gives us strict link between the paper and the `phino` as calculus implementation/rewriting tool.

We should have the same for dataization and morphing. Rules for dataization and morphing should be described in yaml files (we should figure out the format), phino should do dataization and morphing relying on these rules, phino should be able to explain them in latex in the same way it does it with normalization rules. This will give us an ability to paste them to the paper and be sure that there's no inconsistency between the paper and phino.

The actual human-readable version of dataization and morphing rules is presented [here](https://github.com/objectionary/phino/issues/169#issuecomment-3024656974) and in the comments in the phino source code in `Dataize.hs` module before `dataize` and `morph` functions.

maxonfjvipon:
Done — all three requirements are delivered and merged:

1. **YAML rule files** — `resources/morphing.yaml` and `resources/dataization.yaml` describe the 𝕄 and 𝔻 rules (ordered, `match → where → when → then`), alongside the normalization rules now under `resources/normalize/`.
2. **phino runs by interpreting them** — `phino dataize`/`morph` now reduce by interpreting these YAML rules; the hand-written executor was removed.
3. **phino explains them in LaTeX** — `phino explain --morph` and `phino explain --dataize` render the rule tables the same way `explain --normalize` does.

Merged PRs:
- #761 — formalize the rules in YAML + `explain --morph`/`--dataize`
- #770 — relax the `!t` tail matcher (prerequisite: numbers/atoms morph correctly)
- #764 — execution-driven driver (phino dataizes/morphs by the rules)
- #763 — README docs

A related inconsistency in the paper (the head/tail definition forbids application-led tails, contradicting how `Mphi`/`Mlambda` use `• t`) is tracked separately in objectionary/calculus-paper#27. Glyph/LaTeX polish of the explain output for paper-perfect rendering can be a small follow-up if needed.

Closing as complete.

maxonfjvipon:
@rultor release, tag is `0.0.75`

0crat:
[...]

PR #770, opened 2026-06-08
Relax '!t' tail matching to allow application-led tails

### LINKED ISSUE #765: Rule-driven morphing can't reduce atoms: '!t' tail is stricter than the paper's '• t' continuation
## Summary

While implementing the execution-driven part of #760 (making `phino` morph/dataize by **interpreting** `resources/morphing.yaml` / `resources/dataization.yaml` rather than the hand-written `Dataize.morph` / `dataize'`), the rule-driven relations turn out to be **equivalent to the hand-written executor on simple inputs but diverge on every real program that contains a number or an atom** (`5.plus(6)`, Fahrenheit, Factorial, …): they reduce to `nothing` instead of the expected value.

The proximate cause is the `!t` tail meta-variable (as matched by `Matcher.tailExpressions`) being **stricter than the morphing rules need**. Digging further, there is a likely **inconsistency in `calculus-paper`** between the formal *head/tail* definition and the way the `Mphi` / `Mlambda` rules use `• t`. Opening this so we can look at it together.

## Symptom

Pointing `dataize` at the rule-driven driver makes the characterization test fail with an exact diff:

```
5.plus(6):
  expected: [contextualize, Mphi, alpha, copy, dot, …, Mlambda, …, Mprim]   (the real reduction → 11)
  but got:  [contextualize, Mstuck]                                          (→ ⊥ → nothing)
```

`Fahrenheit`, `Factorial`, `Five` all → `nothing` the same way.

## Concrete example

Smallest reproducer — morph `Q.x(y ↦ ⟦⟧)` with universe `Q = ⟦ x ↦ ⟦⟧ ⟧`, i.e. a global dispatch that is **immediately applied**:

```
ExApplication (ExDispatch ExGlobal "x") (y ↦ ⟦⟧)
```

- The rule that should fire is **`Mphi`**: `𝕄(Q.x • t) → 𝕄(⟦⟧ • t)` with `t = (y ↦ ⟦⟧)`.
- `Mphi`'s pattern is `Φ.𝜏 * !t`. To match, `!t` must bind `t = (y ↦ ⟦⟧)` — but that continuation **starts with an application**, which `!t` forbids. So `Mphi` misses, `Mnmz` misses (the expression is already in normal form), and we fall through to `Mstuck` → `⊥`.

Every number is `Φ.number(<bytes>)`, so this hits all arithmetic: `5.plus(6)` is `Φ.number(arg).plus(6)`, and `Mphi` must fire on the inner `Φ.number` with `t = (arg).plus(6)` — again application-led.

## Where in the code

`src/Matcher.hs`, `tailExpressions` — the `ExApplication` branch only peels an application if the head-ward part already produced a **non-empty** tail led by a **dispatch**:

```haskell
ExApplication expr tau -> do
  (substs, tails@(t : _)) <- tailExpressionsReversed ptn' expr   -- requires non-empty tail
  if not (null tails) && isDispatch t then Just (...) else Nothing
```

So an application sitting directly on the matched head (empty intervening tail) → `Nothing`. This faithfully implements `!t`'s documented contract (README → *Meta variables*: *"`!t` … must start only with dispatch"*). The dispatch-led `MatcherSpec` cases (`Q * !t => Q.org(x->…)`) pass precisely because their tail starts with a dispatch.

The old `Dataize.withTail` never used `!t` — it recurses the spine transparently through both applications and dispatches (blocking only `ExApplication (ExFormation _) _`), so it resolves `Φ.number(arg)` fine. That is why the hand-written executor works and the rule-driven one does not.

## The deeper finding — likely inconsistency in `calculus-paper`

`sections/foundations.tex` (`tab:head-and-tail`) defines head/tail with **leading applications absorbed into the head, the tail starting at the first dispatch**:

| Expression | Head | Tail |
|---|---|---|
| `b₁(foo → b₂)` | `b₁(foo → b₂)` | — |
| `b₁.aa.bb(~0 → b₂)` | `b₁` | `aa.bb(~0 → b₂)` |
| `b₁(foo → b₂)(~0 → 42).print(~1 → 7)` | `b₁(foo → b₂)(~0 → 42)` | `print(~1 → 7)` |

By this definition a tail can never start with an application, so `Φ.number(arg)` splits as head `Φ`, tail `.number(arg)`.

But `sections/operators.tex` writes `Mphi` as `𝕄(Q.τ • t) → 𝕄(e • t)` and `Mlambda` as `𝕄(⟦B₁, λ↦f, B₂⟧ • t) → …`. For these to reduce a number, the continuation `t` after the redex `Q.τ` must be allowed to **start with an application** — contradicting the head/tail definition above. So `•`/tail is **overloaded**: the canonical *tail* (foundations) vs a *continuation after the redex* (the rules), and the two are incompatible.

In other words: the `t` in the morphing rules is a **continuation** (the spine following the redex), not the foundations *tail*. The implementation's `!t` was built to the foundations definition but is used where continuation semantics are required.

## Options

1. **Relax `!t`** (matcher only — not the parser): drop the dispatch-first restriction in the `ExApplication` branch so `!t` matches a continuation. Blast radius is small: `!t` is used only by the morphing rules; normalization patterns have no meta-tail (the rewriter is untouched), and the dispatch-led `MatcherSpec` cases still match identically. `⟦…λ…⟧(args)` stays correct because `Mnmz` precedes `Mlambda` and a formation-application is always reducible, so `Mnmz` pre-empts it. The README wording for `!t` would change. **(Recommended.)**
2. **Driver-side descent**: keep the matcher; have the morph driver resolve `Mphi` / `Mlambda` via the existing `withTail` / `phiDispatch` / `formation` (table-drive only the rest). No matcher change, but those two rules stop being purely rule-driven.
3. **Rethink the rule**: drop `• t` from the rule patterns and have the engine descend the spine (a congruence step) — diverges from the paper's compact formulation and makes the explained figure no longer match the executor.

Separately, it is worth an issue on `calculus-paper` about the `•`/tail inconsistency (define the morphing rules' `t` as a *continuation*, distinct from the foundations *tail* — or relax the tail definition, which would change rows 1 and 4 of `tab:head-and-tail`).

## Pointers

- WIP: branch `bug/#760/dataize-morph-driver`, draft PR #764.
- Code: `src/Matcher.hs` (`tailExpressions`), `resources/morphing.yaml` (`Mphi` / `Mlambda`), `src/Dataize.hs` (`withTail`, `morphByRules`, `dataizeByRules`).
- Paper: `objectionary/calculus-paper` — `sections/foundations.tex` (`tab:head-and-tail`), `sections/operators.tex` (morphing rules).

Part of #760.

maxonfjvipon:
Fixes #765.

`Matcher.tailExpressions` rejected any `!t` tail that **started with an application**: its `ExApplication` branch only peeled an application when the head-ward part had already produced a **non-empty** tail led by a **dispatch**. This made `!t` stricter than both the grammar (which parses `* !t` for any head) and the morphing rules' `• t` continuation — so `Q.x * !t` could not match `Q.x(y -> [[]])`. Since every number is `Φ.number(<bytes>)` and atoms are applications, this blocks rule-driven morphing of all arithmetic (the executor falls through to the stuck case).

## Fix

Make the `ExApplication` branch symmetric with the `ExDispatch` branch — the tail may now begin with an application:

```haskell
ExApplication expr tau -> do
  (substs, tails) <- tailExpressionsReversed ptn' expr
  Just (substs, TaApplication tau : tails)
```

Drops the now-unused `isDispatch` helper and the `LambdaCase` pragma, and updates the README wording for `!t`.

## Why it's safe

- `!t` is used only by the (not-yet-active) morphing rules; **no normalization rule uses a meta-tail**, so the rewriter is unaffected.
- The change is a strict superset: every existing dispatch-led `MatcherSpec` tail case still matches identically.
- Full suite green (1063 examples, `-Werror`), `hlint` and `fourmolu` clean, `markdownlint` clean.

## Test

Added a `MatcherSpec` case that fails before the fix and passes after:

```
Q.x * !t  =>  Q.x(y -> [[]])  =>  [(!t >> [(y -> [[]])])]
```

## Background

The corresponding paper inconsistency (the `foundations.tex` head/tail definition forbids application-led tails, contradicting how `operators.tex` uses `• t`) is filed as objectionary/calculus-paper#27. This unblocks the execution-driven driver in #760 (draft PR #764).

### DIFF
--- a/README.md
+++ b/README.md
@@ -399,8 +399,8 @@ This is the list of supported meta variables:
                 `nf:` guard is needed)
 * `!B` || `𝐵` - list of bindings
 * `!d` || `δ` - bytes in meta delta binding
-* `!t` - tail after expression, sequence of applications and/or dispatches,
-         must start only with dispatch
+* `!t` - tail after expression, a possibly empty sequence of applications
+         and/or dispatches
 * `!F` - function name in meta lambda binding
 
 Every meta variable may also be used with an integer index, like `!B1` or `𝜏0`.
--- a/src/Matcher.hs
+++ b/src/Matcher.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE LambdaCase #-}
-
 -- SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
 -- SPDX-License-Identifier: MIT
 
@@ -109,15 +107,8 @@ tailExpressions ptn tgt = case tailExpressionsReversed ptn tgt of
           (substs, tails) <- tailExpressionsReversed ptn' expr
           Just (substs, TaDispatch attr : tails)
         ExApplication expr tau -> do
-          (substs, tails@(t : _)) <- tailExpressionsReversed ptn' expr
-          if not (null tails) && isDispatch t
-            then Just (substs, TaApplication tau : tails)
-            else Nothing
-          where
-            isDispatch :: Tail -> Bool
-            isDispatch = \case
-              TaDispatch _ -> True
-              TaApplication _ -> False
+          (substs, tails) <- tailExpressionsReversed ptn' expr
+          Just (substs, TaApplication tau : tails)
         _ -> Just ([], [])
       substs -> Just (substs, [])
 
--- a/test/MatcherSpec.hs
+++ b/test/MatcherSpec.hs
@@ -409,6 +409,12 @@ spec = do
         , substs [[("t", MvTail [TaDispatch (AtLabel "org"), TaApplication (BiTau (AtLabel "x") (ExFormation [BiVoid AtRho]))])]]
         )
       ,
+        ( "Q.x * !t => Q.x(y -> [[]]) => [(!t >> [(y -> [[]])])]"
+        , ExMetaTail (ExDispatch ExGlobal (AtLabel "x")) "t"
+        , ExApplication (ExDispatch ExGlobal (AtLabel "x")) (BiTau (AtLabel "y") (ExFormation [BiVoid AtRho]))
+        , substs [[("t", MvTail [TaApplication (BiTau (AtLabel "y") (ExFormation [BiVoid AtRho]))])]]
+        )
+      ,
         ( "Q.!a * !t => Q.org.eolang(x -> [[]]) => [(!a >> org, !t >> [ .eolang, ( x -> [[ ]] ) ])]"
         , ExMetaTail (ExDispatch ExGlobal (AtMeta "a")) "t"
         , ExApplication (ExDispatch (ExDispatch ExGlobal (AtLabel "org")) (AtLabel "eolang")) (BiTau (AtLabel "x") (ExFormation [BiVoid AtRho]))


maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Great work on your contribution! 🎉 You've earned **+12 points** (+16 base, -4 for 23 hits-of-code). [Keep up the momentum](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) - your running score is now [+123](https://www.eolang.org/zerocracy/objectionary-vitals.html)! Don't forget to check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #775, opened 2026-06-08
Render data-bytes meta as \delta in explain output

yegor256:
Running `phino explain --dataize` printed the raw Unicode glyph δ for the data-bytes meta, both in the match clause and in the result column, instead of a proper LaTeX command. The rest of the table already uses LaTeX commands (for example the tau meta renders as `\tau`), so the stray δ stood out and would not typeset correctly.

The bytes carried by a delta binding were never passed through `toLaTeX`, so the byte meta head kept its Unicode form. This adds a `ToLaTeX` instance for bytes, threads the conversion through both the match and the result paths, and renders the byte meta as `\delta` to match the `\tau` convention.

One thing left untouched: the capital Δ attribute in the box rule's `if` condition is still emitted as Unicode. That is a separate symbol from the lowercase data meta this PR fixes, so I left it for a follow-up rather than widening the scope here.

### DIFF
--- a/src/CST.hs
+++ b/src/CST.hs
@@ -88,7 +88,7 @@ data META_HEAD
   | B -- 𝐵
   | B' -- B
   | D -- δ
-  | D' -- d
+  | D' -- \delta
   | TAIL -- t
   | F -- F
   deriving (Eq, Show)
--- a/src/LaTeX.hs
+++ b/src/LaTeX.hs
@@ -244,7 +244,8 @@ instance ToLaTeX BINDINGS where
   toLaTeX bds = bds
 
 instance ToLaTeX PAIR where
-  toLaTeX PA_DELTA{..} = PA_DELTA' bytes
+  toLaTeX PA_DELTA{..} = toLaTeX (PA_DELTA' bytes)
+  toLaTeX PA_DELTA'{..} = PA_DELTA' (toLaTeX bytes)
   toLaTeX PA_LAMBDA{..} = PA_LAMBDA' (piped func)
   toLaTeX PA_LAMBDA'{..} = PA_LAMBDA' (piped func)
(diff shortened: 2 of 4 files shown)


yegor256:
@maxonfjvipon let's merge? it fixes the bug

maxonfjvipon:
@rultor merge

yegor256:
@rultor release, tag is `0.0.0.76`

yegor256:
@rultor release, tag is `0.0.76`

0crat:
@yegor256 Thanks for your contribution! 🚀 Just a quick note about branch naming: we follow the [naming guidelines](https://www.yegor256.com/2014/04/15/github-guidelines.html) that recommend using ticket numbers as branch names. For this fix, "`774`" would be perfect instead of "`fix-delta-latex`". This adjustment cost -6 points, bringing your score to [+947](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep up the great work and remember to check your [Zerocracy account](https://www.zerocracy.com/) for updates!

0crat:
@yegor256 Hey! Nice work on your latest contribution 🎉 You snagged **+12 points** this time: started with **+16** as the base reward, but we had to dock **4 points** since you only hit **18** [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html) (our policy gives a small penalty for contributions under **40** HoC). No worries though - [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) those contributions flowing and you'll easily hit the sweet spot next time! Your running score is looking solid at **[+887](https://www.eolang.org/zerocracy/objectionary-vitals.html)**, and don't forget to peek at your **[Zerocracy account](https://www.zerocracy.com/)** for the full picture.

PR #781, opened 2026-06-08
Render Δ as \Delta in explain --dataize LaTeX output

### LINKED ISSUE #780: `explain --dataize` renders raw `Δ` instead of `\Delta` in LaTeX output
Running `phino explain --dataize` emits LaTeX in which the `Δ` attribute of the `box` rule's disjoint condition is printed as the raw Unicode glyph `Δ` rather than the LaTeX macro `\Delta`. The produced row reads `{ if $ [ Δ, L ] \cap ( B_1 \cup B_2 ) = \emptyset $ }`. The offending code is the `ToLaTeX ATTRIBUTE` instance in `src/LaTeX.hs:226-231`, whose catch-all `toLaTeX attr = attr` leaves `AT_DELTA` untouched.

This is wrong because every other token in that condition is converted to a LaTeX-safe form — `λ` becomes `L` and `δ` becomes `\delta` — but `AT_DELTA` falls through the catch-all and is rendered by `render DELTA = "Δ"` in `src/Render.hs:71-72`, leaking a raw Unicode character into math mode and producing `.tex` that does not compile portably.

The fix is to handle `AT_DELTA` in the `ToLaTeX ATTRIBUTE` instance, mirroring the existing `AT_LAMBDA` case: add a `DELTA'` constructor that renders as `\Delta` and return `AT_DELTA DELTA'` so the attribute is emitted as `\Delta`.

yegor256:
Running `phino explain --dataize` was leaking a raw Unicode `Δ` into the generated LaTeX. In the `box` rule's disjoint condition the row came out as `[ Δ, L ]` — the `λ` attribute was already converted to `L`, but the `Δ` attribute slipped through untouched and printed the literal glyph. This change makes it emit `\Delta` instead, so the row now reads `[ \Delta, L ]`.

The cause was the `ToLaTeX ATTRIBUTE` instance in `src/LaTeX.hs`: it handled `AT_LAMBDA` but had no case for `AT_DELTA`, so the attribute fell through to the catch-all and rendered via `render DELTA = "Δ"`. I added a `DELTA'` token that renders as `\Delta` and a matching `AT_DELTA` case, mirroring how `AT_LAMBDA` is already handled. A raw Unicode glyph in math mode doesn't compile portably, which is why this matters for the academic LaTeX output.

The existing `explains dataization rules` test in `test/CLISpec.hs` was updated to expect `\Delta`, and the full suite passes under `-Werror` along with hlint and fourmolu.

Closes #780

### DIFF
--- a/src/CST.hs
+++ b/src/CST.hs
@@ -46,7 +46,7 @@ data PHI = PHI | AT
 data RHO = RHO | CARET
   deriving (Eq, Show)
 
-data DELTA = DELTA
+data DELTA = DELTA | DELTA'
   deriving (Eq, Show)
 
 data XI = XI | DOLLAR
--- a/src/LaTeX.hs
+++ b/src/LaTeX.hs
@@ -227,6 +227,7 @@ instance ToLaTeX ATTRIBUTE where
   toLaTeX AT_LABEL{..} = AT_LABEL (piped label)
   toLaTeX AT_META{..} = AT_META (toLaTeX meta)
   toLaTeX AT_LAMBDA{} = AT_LAMBDA LAMBDA'
+  toLaTeX AT_DELTA{} = AT_DELTA DELTA'
   toLaTeX AT_REST{} = AT_REST DOTS'
   toLaTeX attr = attr
 
--- a/src/Render.hs
+++ b/src/Render.hs
@@ -70,6 +70,7 @@ instance Render RHO where
 
 instance Render DELTA where
   render DELTA = "Δ"
+  render DELTA' = "\\Delta"
 
 instance Render XI where
   render XI = "ξ"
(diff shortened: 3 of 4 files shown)


yegor256:
@maxonfjvipon FYI, let's merge?

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon **Great work on the review! 🎉** You've earned **+4 points** (base +12, -8 for missing comments, -4 for limited code coverage at 6 hits, +4 bonus applied). Your total score is now [+129](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep up the momentum and check your [Zerocracy account](https://www.zerocracy.com/) for updates!

0crat:
@yegor256 Thanks for the contribution! You've earned +12 points for this: +16 as a basis per our policy, but -4 for contributing only 6 hits-of-code (below our 40 minimum threshold). To maximize your bonus next time, aim for at least 40 hits-of-code to avoid the penalty while staying under 100 to prevent additional deductions. [Keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming! Your running score is [+899](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #786, opened 2026-06-10
Introduce `𝑘`/`!k` absolute meta-variable for the COPY rule (#785)

### LINKED ISSUE #785: Normalization Rules Invalid: Replace `\xifree` with Absolute Expressions
Let's introduce a new concept of absolute expression.

Formally, it's this:

"An normal form is absolute, denoted by $a$ and ranging over $\mathcal{A} \subseteq \mathcal{N}$, if it is either a) $Q$, b) formation, c) dispatch with an absolute subject, d) an application with absolute subject and argument."

Let's get rid of `\xifree` in normalization rules. Instead, let's use $a$ in the `COPY` rule.

yegor256:
Closes #785.

## What

Introduces the concept of an **absolute expression** and a dedicated meta-variable `𝑘`/`!k` to capture it, replacing the `\xifree` side-condition in the `COPY` normalization rule.

An expression is *absolute* (ranges over `𝒦 ⊆ 𝒩`) when it is **xi-free** *and* **in normal form**. Xi-free means it contains no `ξ` outside of a formation — it is `Φ`, a formation, a dispatch with an absolute subject, or an application with an absolute subject and argument.

## Approach

Mirrors the existing `𝑛`/`!n` normal-form meta-variable:

- New `𝑘`/`!k` meta-variable (parser, CST head, rendering, ASCII encoding) that binds only absolute sub-expressions — so no explicit guard is needed, just like `𝑛` needs no `nf:` guard.
- The constraint is wired through the same machinery as `𝑛`: a prefix-keyed key-space (`k`) collected by `kMetaNames`, applied during rule matching. Per @maxonfjvipon's note, the check runs **xi-free first** (cheap, structural, avoids the `ξ`-recursion the normal-form check could loop on) and **then is-nf**, matching `𝒦 ⊆ 𝒩`.
- `COPY` now reads `⟦ 𝐵1, 𝜏 ↦ ∅, 𝐵2 ⟧(𝜏 ↦ 𝑘)` → `⟦ 𝐵1, 𝜏 ↦ 𝑘, 𝐵2 ⟧` with **no `when:` clause** — the `\xifree` macro is gone.
- The set of absolute expressions renders as `\mathcal{K}` (𝒦).

`COPY` behaviour is **unchanged**: `𝑘` = NF ∧ xi-free is exactly what the old `𝑛` + `when: xi-free: 𝑛` enforced. The underlying `absolute:` condition is retained as the primitive the `𝑘` meta builds on (parallel to how `nf:` underlies `𝑛`), and is exercised by the `condition-packs/absolute-*.yaml` fixtures.

## Naming

`!a` was rejected because it already denotes the **attribute** meta (`!a || 𝜏`) and appears in expression position in existing parser tests (`p -> !a` ⇒ `$.a`). Per @yegor256, the meta is `k`/`𝑘` and the set is `K`/`𝒦`.

## Notes

⚠️ This environment has no GHC/Cabal toolchain, so I could not run `make test` / `make hlint` / `make fourmolu` locally — relying on CI.

https://claude.ai/code/session_01XDkkUWQv7ZZ4x8thamU8bQ


yegor256:
@maxonfjvipon WDYT?

maxonfjvipon (approved):

maxonfjvipon:
@yegor256 looks good to me

yegor256:
@maxonfjvipon wait a sec, I will modify it

maxonfjvipon:
@yegor256 if you want to introduce `a` meta variable in the syntax similar to `n` - make sure you check 'xi-free' before 'is-nf'

src/Rule.hs
```diff
@@ -367,12 +380,13 @@ matchExpressionBy matcher expr rule ctx =
           pure []
         else do
           inNf <- foldlM (\substs nm -> meetCondition (Y.NF (ExMeta nm)) substs ctx) matched (nfMetaNames ptn)
-          if null inNf
+          inK <- foldlM (\substs nm -> meetCondition (Y.Absolute (ExMeta nm)) substs ctx) inNf (kMetaNames ptn)
```
maxonfjvipon: @yegor256 here you check `isNF` condition twice: first above when constructing `inNF` and second inside `Y.Absolute` condition

src/Rule.hs
```diff
@@ -175,19 +175,25 @@ _nf (ExMeta meta) (Subst mp) ctx = case M.lookup meta mp of
   _ -> pure []
 _nf expr subst ctx = pure [subst | isNF expr ctx]
 
-_xiFree :: Expression -> Subst -> RuleContext -> IO [Subst]
-_xiFree (ExMeta meta) (Subst mp) ctx = case M.lookup meta mp of
-  Just (MvExpression expr) -> _xiFree expr (Subst mp) ctx
+-- An expression is xi-free when it contains no ξ outside of a formation:
+-- it is Φ, a formation, a dispatch with a xi-free subject, or an application
+-- with a xi-free subject and argument.
+xiFree :: Expression -> Bool
```
maxonfjvipon: @yegor256 let's move `xiFree` to the scope of `_absolute` function using `where` syntax

maxonfjvipon (changes requested):

yegor256:
@maxonfjvipon I renamed it to $k$, check again

src/Rule.hs
yegor256: Fixed in 36ffcd0. `_absolute` (the `Y.Absolute` condition) now does only the structural xi-free check; the normal-form check for `𝑘` metas is folded into the same `Y.NF` pass used for `𝑛` metas, so `isNF` is applied in a single place — xi-free first, then is-nf (`𝒦 ⊆ 𝒩`).

---

yegor256:
@maxonfjvipon should be good now, please take a look

0crat:
@maxonfjvipon **Great work on the review! 🎉** You've earned **+7 points** (+12 base, -5 for minimal comments). Your total score is now [+172](https://www.eolang.org/zerocracy/objectionary-vitals.html). Keep up the momentum and check your [Zerocracy account](https://www.zerocracy.com/) for updates!

0crat:
@yegor256 Thanks for the contribution! You've earned **+12 points** for this: **+16** as a basis, **+7.4** for 148 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), **-4** for exceeding 100 [hits-of-code](https://www.yegor256.com/2014/11/14/hits-of-code.html), and **-7.4** to cap the HoC bonus at 16 points maximum. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+949](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #788, opened 2026-06-10
Replace remaining "xi-free" naming with "absolute"

yegor256:
Follow-up to #786 (issue #785).

@yegor256 noted that "xi-free" terminology was still in the code and everything should be phrased as **absolute**.

To clarify: the `CO_XIFREE` *constructor* was already removed in #786 (it's `CO_ABSOLUTE` now). What remained was the structural helper `xiFree` (plus the `inXiFree` binding, comments, a log message, and README wording). That recursion is exactly the formal absolute definition — Φ, a formation, a dispatch with an absolute subject, or an application with an absolute subject and argument — so this renames it to `absolute` and drops the "xi-free" naming everywhere.

**Pure rename, no behavioural change.** The `𝑘` meta still binds only absolute normal forms (absolute structure checked first, then normal form), and `isNF` is still applied in a single place.

⚠️ No GHC/Cabal toolchain in this environment, so I couldn't run `make test`/`hlint`/`fourmolu` locally — relying on CI.

https://claude.ai/code/session_01XDkkUWQv7ZZ4x8thamU8bQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01XDkkUWQv7ZZ4x8thamU8bQ)_

### DIFF
--- a/src/Rule.hs
+++ b/src/Rule.hs
@@ -175,12 +175,12 @@ _nf (ExMeta meta) (Subst mp) ctx = case M.lookup meta mp of
   _ -> pure []
 _nf expr subst ctx = pure [subst | isNF expr ctx]
 
--- An expression is absolute (ranges over 𝒦 ⊆ 𝒩) when, being a normal form, it
--- is Φ, a formation, a dispatch with an absolute subject, or an application
--- with an absolute subject and argument (equivalently, no ξ leaks outside of a
--- formation). This checks that structure; the '𝑘' meta-variable applies this
--- check first (cheap, structural, rules out the recursion the normal-form check
--- could loop on) and the normal-form check second.
+-- Checks the structural part of absoluteness: an expression is absolute in
+-- structure when it is Φ, a formation, a dispatch with an absolute subject, or
+-- an application with an absolute subject and argument (equivalently, no ξ
+-- leaks outside of a formation). It does not enforce normal form on its own;
+-- the '𝑘' meta-variable combines this structural check (applied first, since it
+-- is cheap) with the normal-form check, so that 𝒦 ⊆ 𝒩.
 _absolute :: Expression -> Subst -> RuleContext -> IO [Subst]
 _absolute (ExMeta meta) (Subst mp) ctx = case M.lookup meta mp of
   Just (MvExpression expr) -> _absolute expr (Subst mp) ctx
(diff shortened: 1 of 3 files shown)


yegor256:
@maxonfjvipon better now?

0crat:
@yegor256 According to our [work policy](https://www.yegor256.com/2014/04/15/github-guidelines.html), your branch name "`claude/absolute-naming`" doesn't follow our naming convention, resulting in a **-6 point penalty** as specified in our guidelines. Branch names should match the ticket number you're solving - in this case, "`787`" would be the correct name. Your running score is [+897](https://www.eolang.org/zerocracy/objectionary-vitals.html); please check your [Zerocracy account](https://www.zerocracy.com/) and follow our established naming conventions going forward.

yegor256:
@maxonfjvipon what about now?

src/Rule.hs
```diff
@@ -175,24 +175,24 @@ _nf (ExMeta meta) (Subst mp) ctx = case M.lookup meta mp of
...
-_absolute expr subst _ = pure [subst | xiFree expr]
+_absolute expr subst ctx = pure [subst | absolute expr && isNF expr ctx]
   where
-    xiFree :: Expression -> Bool
-    xiFree (ExFormation _) = True
-    xiFree ExGlobal = True
-    xiFree (ExApplication e (BiTau _ te)) = xiFree e && xiFree te
-    xiFree (ExDispatch e _) = xiFree e
-    xiFree _ = False
+    absolute :: Expression -> Bool
```
maxonfjvipon: @yegor256 I think this function should named as `xiFree`

src/Rule.hs
```diff
@@ -378,17 +378,14 @@ matchExpressionBy matcher expr rule ctx =
...
           pure []
         else do
-          -- A '𝑘' meta-variable is absolute (𝒦 ⊆ 𝒩): check it is xi-free first
-          -- (cheap, structural), then fold its name into the same normal-form
-          -- check used for '𝑛' metas, so 'isNF' is applied in a single place.
-          inXiFree <- foldlM (\substs nm -> meetCondition (Y.Absolute (ExMeta nm)) substs ctx) matched (kMetaNames ptn)
-          inNf <- foldlM (\substs nm -> meetCondition (Y.NF (ExMeta nm)) substs ctx) inXiFree (nfMetaNames ptn ++ kMetaNames ptn)
-          if null inNf
+          absolute <- foldlM (\substs nm -> meetCondition (Y.Absolute (ExMeta nm)) substs ctx) matched (kMetaNames ptn)
+          normal <- foldlM (\substs nm -> meetCondition (Y.NF (ExMeta nm)) substs ctx) absolute (nfMetaNames ptn)
```
maxonfjvipon, line 381: @yegor256 here you would check `isNF` condition twice, first when building `absolute`, second when building `normal` using absolute

maxonfjvipon (changes requested):

src/Rule.hs
yegor256: Done in 6e203ef — renamed the inner structural predicate to `xiFree`. The recursion is exactly the structural "no ξ leaks outside a formation" property, so `xiFree` is the right name. The public `absolute:` condition and `_absolute` keep their names; only this helper became `xiFree`.

src/Rule.hs
yegor256: Fixed in 6e203ef — dropped `isNF` from `_absolute` so it is purely structural (`xiFree`), and the matcher normal-form fold now covers the `𝑘`-metas alongside the `𝑛`-metas, so `isNF` is applied once. The `𝑘` meta still ranges over 𝒦 ⊆ 𝒩 (structural `absolute:` ∩ `nf:`), so Rcopy is unchanged.

yegor256:
@maxonfjvipon fixed, please check again

maxonfjvipon (changes requested):
@yegor256 now this change does not make sense because it's just renaming which is not actually right. Absolute = xiFree + normal form. In this PR absolute is only xi-free

yegor256:
@maxonfjvipon take a look again. do you still need this PR? if not, just close it

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+226](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #789, opened 2026-06-11
Sweet printer emits NaN/Infinity double literals that the parser rejects


yegor256:
The sweet printer and the number parser disagree on how a `Φ.jeo.double` holding a special IEEE-754 value is rendered. In `src/Render.hs:197`, `render EX_NUMBER{..} = either (T.pack . show) (T.pack . show) num` prints a `Double` via Haskell's `show`, which yields the bare tokens `NaN`, `Infinity` and `-Infinity` for those values. The `number` parser in `src/Parser.hs:174` reads literals with `lexeme L.scientific`, which only accepts decimal/scientific numerals and has no case for those tokens.

The result is that phino cannot re-parse its own output. Disassembling a class that contains an `ldc Double.NaN` constant (for example `org.json.JSONObject`) and printing it produces `n0 ↦ NaN`; feeding that same program back into phino fails with `unexpected "Na", expecting binding`. A minimal reproduction: `phino rewrite --input=xmir --sweet` on an XMIR whose double has the bytes `7F-F8-00-00-00-00-00-00` prints `{⟦ x ↦ ⟦ φ ↦ Φ.jeo.double, n0 ↦ NaN ⟧ ⟧}`, and `phino rewrite --sweet` on that string then errors at column 33. This breaks any multi-pass pipeline (print, then re-read) on programs carrying NaN or infinite float constants.

The smallest fix is to make the two sides agree: either teach the `number` parser to accept the `NaN`/`Infinity`/`-Infinity` tokens that `show` already emits, or change the `EX_NUMBER` render branch to print those special values as something the existing parser can read. Aligning the printed form with the accepted grammar is enough; no broader number-handling change is needed.

yegor256:
@rultor release, tag is `0.0.78`

0crat:
@yegor256 Great work closing this issue! 🎉 You've earned **+12 points** (+8 base + 4 bonus for sub-12hr resolution). Your total score now stands at [+997](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep up the momentum and check your [Zerocracy account](https://www.zerocracy.com/) for updates!

PR #794, opened 2026-06-11
Preserve negative zero (-0.0) through parse/render pipeline

### LINKED ISSUE #793: Negative zero (-0.0) collapses to 0 in `btsToNum` (Misc.hs:324)
`btsToNum` in `src/Misc.hs:313-325` converts a number's 8 bytes to `Either Int Double`. For negative zero (bytes `80-00-00-00-00-00-00-00`), `wordToDouble` yields `-0.0`, which is neither NaN nor infinite, so `properFraction (-0.0)` returns `(0, 0.0)` and the `(n, 0.0) -> Left n` branch at `src/Misc.hs:324` stores `Left 0`. The sweet renderer (`src/Render.hs:197`, `either (show) (show) num`) then prints the bare `Int` `0`. So `phino rewrite --input=xmir --sweet` on `Φ.number(Φ.bytes(⟦ Δ ⤍ 80-00-00-00-00-00-00-00 ⟧))` emits `0`, and on reassembly the value becomes `+0.0` (`00-00-00-00-00-00-00-00`).

This is wrong because `-0.0` and `+0.0` are distinct IEEE-754 values (`Double.compare` separates them), so collapsing the sign silently changes semantics across any print-then-reparse pipeline. The number parser compounds it: `number` at `src/Parser.hs:174-186` reads the literal via `lexeme L.scientific` and `negate`, and `Scientific` has no negative zero, so even the literal `-0.0` parses to `+0.0`. The same NaN/Infinity guard added in #789 is what skips those specials while letting negative zero fall through. This surfaced downstream in JEO→phino→JEO round-tripping of `org.json.JSONObject`, whose `getNumber` returns `Double.valueOf(-0.0)`; after the round-trip the constant became `+0.0` and `JSONObjectTest.jsonObjectValues` failed.

The smallest fix is to keep negative zero as a `Double` rather than an `Int`: in `btsToNum`, guard the integral branch so that `val == 0 && isNegativeZero val` returns `Right val` (mirroring the existing NaN/Infinity case at `src/Misc.hs:321`), and in the `number` parser preserve the sign of a zero literal instead of discarding it through `Scientific`/`negate`.

yegor256:
Fixes #793.

## Problem

`-0.0` and `+0.0` are distinct IEEE-754 values, but `phino` collapsed the sign in two places:

1. **`btsToNum` (`src/Misc.hs`)** — for bytes `80-00-00-00-00-00-00-00`, `wordToDouble` yields `-0.0`, which is neither NaN nor infinite, so `properFraction (-0.0)` returns `(0, 0.0)` and the `(n, 0.0) -> Left n` branch stored `Left 0`. The sweet renderer then printed the bare `Int` `0`, and on reassembly the value became `+0.0`.

2. **`number` parser (`src/Parser.hs`)** — the literal sign was applied via `negate` on the `Scientific` value, but `Scientific` has no negative zero, so even `-0.0` parsed to `+0.0`.

This surfaced in JEO→phino→JEO round-tripping of `org.json.JSONObject`, where `getNumber` returns `Double.valueOf(-0.0)`.

## Fix

- In `btsToNum`, guard the integral branch with `isNegativeZero val` so `-0.0` is returned as `Right val` (mirroring the existing NaN/Infinity guard added in #789/#790).
- In the `number` parser, negate the `Double` (after `toRealFloat`) rather than the `Scientific`, which yields a distinct `-0.0`.

## Tests

- Added a `btsToNum` doctest for `80-00-00-00-00-00-00-00 -> Right (-0.0)`.
- Added parser cases for `-0`, `-0.0`, and `+0` in `ParserSpec`.

Note: I was unable to run the Haskell test suite in this environment (no GHC/Cabal toolchain present), so CI verification is needed.

https://claude.ai/code/session_01UMXZzubf56smTRG4naSqpD

---
_Generated by [Claude Code](https://claude.ai/code/session_01UMXZzubf56smTRG4naSqpD)_

### DIFF
--- a/src/Misc.hs
+++ b/src/Misc.hs
@@ -310,6 +310,8 @@ toHex w = [digit (w `shiftR` 4), digit (w .&. 0x0F)]
 -- Right Infinity
 -- >>> btsToNum (BtMany ["FF", "F0", "00", "00", "00", "00", "00", "00"])
 -- Right (-Infinity)
+-- >>> btsToNum (BtMany ["80", "00", "00", "00", "00", "00", "00", "00"])
+-- Right (-0.0)
 btsToNum :: Bytes -> Either Int Double
 btsToNum hx =
   let bytes = btsToWord8 hx
@@ -318,7 +320,7 @@ btsToNum hx =
         else
           let word = toWord64BE bytes
               val = wordToDouble word
-           in if isNaN val || isInfinite val
+           in if isNaN val || isInfinite val || isNegativeZero val
                 then Right val
                 else case properFraction val of
                   (n, 0.0) -> Left n
--- a/src/Parser.hs
+++ b/src/Parser.hs
@@ -177,11 +177,12 @@ number = do
   return
     ( DataNumber
         ( numToBts
-            ( toRealFloat
-                ( case sign of
-                    Just '-' -> negate unsigned
-                    _ -> unsigned
-                )
+            ( case sign of
+                -- Negate the Double rather than the Scientific so that a zero
+                -- literal preserves its sign: Scientific has no negative zero,
+                -- but negate on Double yields -0.0, a distinct IEEE-754 value.
+                Just '-' -> negate (toRealFloat unsigned)
+                _ -> toRealFloat unsigned
             )
         )
     )
--- a/test/ParserSpec.hs
+++ b/test/ParserSpec.hs
@@ -356,6 +356,9 @@ spec = do
     test
       parseNumber
       [ ("0", Just (DataNumber (BtMany ["00", "00", "00", "00", "00", "00", "00", "00"])))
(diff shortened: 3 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned **+4 points** for this: **+16** as a basis per our [work policy](https://www.yegor256.com/2014/11/14/hits-of-code.html); **-8** for the lack of code review (which our policy requires); **-4** for contributing fewer than 40 hits-of-code (you had 18). To maximize your bonus next time, ensure your contributions get reviewed and aim for at least 40 hits-of-code while [keeping quality high](https://www.yegor256.com/2018/03/06/speed-vs-quality.html). Your running score is [+1045](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too.

PR #798, opened 2026-06-11
Speed up rewrite loop detection with a hashed program digest (#797)

### LINKED ISSUE #797: rewrite does not finish on large programs: per-cycle whole-program loop-detection set (Rewriter.hs:177,190)
`phino rewrite` does not finish in practical time on the XMIR of a large class. With the `streams/*` ruleset it fails to rewrite a single ~4000-line class (apache/commons-io `IOUtils`, commit `cf5f9a2`) within 999 seconds, and stalls likewise on `FileUtils`/`PathUtils`; small programs are fine (json-java's ~30 small classes finish in ~50s total). The per-rule loop `_rewrite` in `src/Rewriter.hs:147-190` runs up to `_maxDepth` cycles, and several steps inside it are each O(program size): the loop-detection set at `Rewriter.hs:177` and `:190` does `Set.member`/`Set.insert` on the *whole* rewritten program `expr`, the match at `:161` rescans the entire program, and the debug render `printExpression expr` at `:184` serializes the whole program every cycle.

Because `_unique :: Set` holds entire program expressions, each `member`/`insert` hashes and compares the full program, and the set grows to one whole-program copy per cycle — so loop detection alone costs roughly O(maxDepth × programSize) per rule, which explodes on a large class while staying cheap on small ones. This makes whole-repo optimization through `hone-maven-plugin` infeasible: `--max-cycles` is already 1 and cannot be lowered, and the per-file budget is 999s.

The smallest fix is to stop storing whole expressions in `_unique`: key the loop-detection set on a cheap fixed-size digest of the program (e.g. a hash) instead of the full `Expression`, so `Set.member`/`Set.insert` at `Rewriter.hs:177,190` become O(1) rather than O(programSize). Guarding the `printExpression expr` argument at `:184` so it is built only when debug logging is on would remove a second per-cycle full-program pass.

yegor256:
Fixes #797.

## Problem

The per-rule rewriting loop in `Rewriter.hs` kept a `Set Expression` (`_unique`) of every rewritten program to detect cycles. Because the set held **whole** program expressions:

- each `Set.member` / `Set.insert` hashed and compared the entire program (`Rewriter.hs:177,190`), and
- the set grew by one full-program copy per cycle.

So loop detection alone cost roughly `O(maxDepth × programSize)` per rule, which exploded on large classes (e.g. apache/commons-io `IOUtils`/`FileUtils`/`PathUtils` with the `streams/*` ruleset) while staying cheap on small ones.

## Fix (dirty-by-hash with a slow re-check)

Per the issue and the request on the ticket: do a **fast, dirty comparison by hash**, and whenever the hash says two programs are equal, **re-compare the old, slow way** to confirm. This is both faster and still sound.

- Added `hashExpression :: Expression -> Int` to `AST.hs` — a cheap fixed-size FNV-1a-style digest over the AST.
- Replaced `_unique :: Set Expression` with `Seen = Map Int [Expression]`, keyed by the digest, storing the full expressions that produced each digest.
- `seenMember` does an `O(1)` digest lookup; the exact structural `(==)` runs **only** on a digest match, so genuine loops are still detected exactly and digest collisions can never cause a false positive.

Loop detection is now `O(programSize)` per cycle for hashing plus `O(log n)` on the digest, instead of `O(maxDepth × programSize)`.

## Verification

- `cabal build` clean with `-Werror`
- `cabal test spec` — 1065 examples, 0 failures (incl. the existing "with looping rules" loop-detection test)
- `hlint` and `fourmolu --mode check` both clean on the changed files

---
_Generated by [Claude Code](https://claude.ai/code/session_01LNki8aBHp2noZNPj6g11Ja)_

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE LambdaCase #-}
 
 -- SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
 -- SPDX-License-Identifier: MIT
@@ -6,6 +7,8 @@
 -- This module represents AST tree for parsed phi-calculus program
 module AST where
 
+import Data.Bits (xor)
+import Data.List (foldl')
 import Data.Text (Text)
 import qualified Data.Text as T
 import GHC.Generics (Generic)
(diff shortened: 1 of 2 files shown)


yegor256:
@maxonfjvipon good?

0crat:
@yegor256 **Great contribution! 🎉** You've earned **+4 points**: +16 base reward, -4 for code size (118 >= 100 lines), -8 for missing code review. Please [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) the excellent work coming! Your total score is now [+1109](https://www.eolang.org/zerocracy/objectionary-vitals.html) — check your [Zerocracy account](https://www.zerocracy.com/) for updates.

PR #803, opened 2026-06-15
Use a primitive 𝑝 meta-variable in the Mprim morphing rule

### LINKED ISSUE #802: `explain --morph` renders Mprim with redundant `e ∈ P` condition instead of a primitive `p` meta
Running `phino explain --morph` renders the Mprim morphing rule with the side condition `if $ e \in \mathcal{P} $`. The rule lives in `resources/morphing.yaml:6-11`, where `match` and `then` use the ordinary `𝑒` meta and a `when: primitive: 𝑒` clause restricts it to primitives; that clause is turned into LaTeX by `src/Render.hs:258`.

This is awkward notation. By convention the letter `p` already denotes a primitive, so the rule reads cleaner as `𝕄(p) ⟿ p` with no side condition at all — the `e ∈ P` guard merely repeats what a `p`-named variable would already imply, and that is how the rule is written on paper.

The smallest fix is to add a `𝑝` primitive-constrained meta-variable next to the existing `𝑛`/`𝑘` kinds in `exMetaHead` (`src/CST.hs:291`), make `src/Matcher.hs` bind such a meta only when the target is primitive, and rewrite Mprim to `match: 𝑝` / `then: 𝑝` without the `when` condition.

yegor256:
Running `phino explain --morph` printed the Mprim rule as `𝕄(e) ⟿ e if e ∈ 𝒫`. The letter `p` already means "a primitive" in the calculus, so spelling out the `e ∈ 𝒫` side condition was just noise. This rewrites the rule to `𝕄(p) ⟿ p` with no condition at all.

To keep the rule honest rather than only cosmetic, `𝑝` is a real meta-variable kind now, sitting next to the existing `𝑛` and `𝑘` kinds. It parses (`!p` or `𝑝`), renders to `p` in LaTeX, and — the important part — the matcher binds it only when the target is actually a primitive. So Mprim still fires exactly when it used to, just without the explicit `when: primitive` guard. The shared `primitive` check moved into `AST` so the matcher and the condition code use the same definition.

The dataization `norm` rule keeps its `e ∉ 𝒫` form on purpose: that one is a negated condition where the `p` convention does not read naturally, so it stays out of scope here.

Closes #802

### DIFF
--- a/resources/morphing.yaml
+++ b/resources/morphing.yaml
@@ -4,11 +4,9 @@
 # Morphing 𝕄 — applied top-to-bottom, first matching clause wins.
 
 - name: Mprim
-  description: '𝕄(e) ⟿ e if e ∈ 𝒫'
-  match: 𝑒
-  when:
-    primitive: 𝑒
-  then: 𝑒
+  description: '𝕄(p) ⟿ p'
+  match: 𝑝
+  then: 𝑝
 
 - name: Mnmz
   description: '𝕄(e) ⟿ 𝕄(𝒩(e)) if e ∉ NF'
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -128,6 +128,17 @@ hashExpression = goExpr fnvOffset
       AtDelta -> step h 26
       AtMeta t -> hashText (step h 27) t
 
+-- A primitive is the termination ⊥ or a formation without a λ binding.
+primitive :: Expression -> Bool
+primitive ExTermination = True
+primitive (ExFormation bds) = not (any lambda bds)
+  where
+    lambda :: Binding -> Bool
+    lambda (BiLambda _) = True
+    lambda (BiMetaLambda _) = True
+    lambda _ = False
+primitive _ = False
+
 countNodes :: Expression -> Int
 countNodes (ExFormation bds) = 1 + sum (map nodesInBinding bds) + length bds
   where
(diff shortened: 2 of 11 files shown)


yegor256:
@maxonfjvipon mind taking a look when you get a chance? Happy to explain any of the meta-variable plumbing if it is unclear.

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the review! You've earned **+4 points** for this contribution. According to our [policy](https://www.yegor256.com/2014/11/14/hits-of-code.html), you received +12 points as the base reward for authoring a code review, but -8 points were deducted since no comments were posted during the review process. Your running score is [+200](https://www.eolang.org/zerocracy/objectionary-vitals.html) - keep engaging with reviews through comments to maximize your bonus potential (don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too)! 💪

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+852](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #805, opened 2026-06-15
Drop the redundant alpha guard from the miss rule

### LINKED ISSUE #804: MISS rule carries a redundant `not: alpha: 𝜏` guard because the matcher lets 𝜏 bind alpha
The MISS normalization rule (`resources/normalize/miss.yaml:8-14`) guards its `when` with `not: alpha: 𝜏`, asserting that 𝜏 must not be one of α0, α1, …. By definition 𝜏 ranges only over φ, ρ, and labels (lowercase identifiers) — never an alpha index — so on the face of it the guard restates something that can never happen.

The guard is nonetheless load-bearing, which is the real defect: `matchAttribute` (`src/Matcher.hs:68`) lets a 𝜏 meta-variable bind to *any* attribute, including `AtAlpha`. So MISS's pattern `⟦𝐵⟧(𝜏 ↦ 𝑒)` matches positional applications too, and the `alpha` rule actually relies on the same permissiveness (its `index: 𝜏2` reads an `AtAlpha`). Without the guard, MISS races the `alpha` rule on positional applications: under `--shuffle` (or any order where MISS precedes `alpha`) a resolvable `⟦x↦∅⟧(α0↦e)` collapses to ⊥ instead of being resolved. In other words the matcher does not honour 𝜏's definition, and the guard papers over that.

Suggested fix: make 𝜏 strict in `matchAttribute` (a 𝜏 meta never matches `AtAlpha`), introduce a dedicated alpha-only meta `𝛼` for the `alpha` rule's index lookup, drop MISS's redundant guard, and add a DC-like rule `dca` (`⊥(𝛼 ↦ 𝑒) → ⊥`) to cover dead positional applications. This mirrors #802, which introduced the primitive-constrained `𝑝` meta to drop a redundant side condition from Mprim.

yegor256:
The `miss` normalization rule carried a `not: alpha: 𝜏` guard saying that 𝜏 must not be one of α0, α1, …. That just restates 𝜏's own definition — it only ranges over φ, ρ, and labels — so instead of keeping the guard, this makes the matcher actually honour the definition: a 𝜏 meta-variable no longer binds an alpha attribute. A dedicated alpha-only meta `𝛼` takes over where an alpha is genuinely meant (the `alpha` rule's index lookup), and the guard is dropped.

The guard wasn't cosmetic, which is the real point. `matchAttribute` let a 𝜏 meta bind to any attribute including an alpha index, so `miss` would match positional applications and race the `alpha` rule — under `--shuffle` a resolvable `⟦x↦∅⟧(α0↦e)` could collapse to ⊥ depending on rule order. With 𝜏 strict, `miss` can't touch positional applications at all, so normalization stays order-independent.

Because `dc` (`⊥(𝜏↦𝑒)→⊥`) also relied on 𝜏 matching alpha, a sibling rule `dca` (`⊥(𝛼↦𝑒)→⊥`) now covers dead positional applications. Rendering of `𝛼` is wired through CST/Render/LaTeX/Encoding so `explain` shows `\alpha`. This follows the same shape as #802, which added the primitive-constrained `𝑝` meta to drop a side condition from Mprim. All tests pass and hlint/fourmolu are clean.

Closes #804


yegor256:
@maxonfjvipon take a look. now `!h` is "any alpha"

regression is caused by adding new DCA rule -- now normalization is slower

resources/normalize/alpha.yaml
```diff
@@ -2,9 +2,9 @@
 # SPDX-License-Identifier: MIT
 ---
 name: alpha
-pattern: ⟦𝐵1, 𝜏1 ↦ ∅, 𝐵2⟧(𝜏2 ↦ 𝑒)
+pattern: ⟦𝐵1, 𝜏1 ↦ ∅, 𝐵2⟧(𝜂 ↦ 𝑒)
```
maxonfjvipon, line 5: @yegor256 `𝜏1` can be changed to just `𝜏`

src/Matcher.hs
```diff
@@ -65,7 +65,9 @@ combineMany :: [Subst] -> [Subst] -> [Subst]
 combineMany xs xy = catMaybes [combine x y | x <- xs, y <- xy]
 
 matchAttribute :: Attribute -> Attribute -> [Subst]
-matchAttribute (AtMeta meta) tgt = [substSingle meta (MvAttribute tgt)]
+matchAttribute (AtMeta meta) tgt
```
maxonfjvipon: @yegor256 this code does not make sense - in both cases it returns `substSingle meta (MvAttribute tgt)`

resources/normalize/dca.yaml
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+name: dca
```
maxonfjvipon, line 4: @yegor256 so this rule is the same as `dc` but with `alpha`?

src/Parser.hs
```diff
@@ -332,10 +332,12 @@ attribute =
 -- 3. rho
 -- 4. phi
 -- 5. alpha
+-- 6. eta (alpha-index meta)
 fullAttribute :: Parser Attribute
 fullAttribute =
   choice
-    [ attribute
+    [ try attribute
+    , AtMeta <$> meta' 'h' "𝜂"
```
maxonfjvipon: @yegor256 let's move it to the `attribute` function

maxonfjvipon (changes requested):
@yegor256 I think this is the point where we should introduce new type to AST for alpha-attributes instead of trying to build crutches using current AST. 
Since alphas can be used only in application arguments, maybe it would be better to make a special type for them together with special meta type

resources/normalize/dca.yaml
yegor256: @maxonfjvipon yes, exactly

yegor256:
@maxonfjvipon good point — the name-prefix trick was the weak part. Pushed `AtMetaAlpha`, a dedicated `Attribute` constructor for the `𝜂` alpha-index meta, so the kind now lives in the type instead of being re-derived from the `'h'` prefix at match/render time.

What that removed:

- `matchAttribute` no longer does `"h" \`isPrefixOf\` meta` — it pattern-matches `AtMetaAlpha` directly (`𝜏`/`AtMeta` stays non-alpha, `𝜂`/`AtMetaAlpha` is alpha-only).
- The `attrMetaHead` prefix sniffer in `CST.hs` is gone; `toCST` maps the constructor straight to the `ETA` head.
- Condition handling (`alpha`, `index`, `eq`, `where`) goes through a small `metaName :: Attribute -> Maybe Text` helper, so `𝜂` behaves uniformly with `𝜏` everywhere.

Parser, rendering and the rule files are unchanged — `𝜂`/`!h`/`\eta` still parse and print the same; only the AST representation behind them is now typed. All 1074 tests, hlint and fourmolu are green.

On the fuller idea — a separate type confined to application arguments — that's a bigger structural change (it touches `BiTau`/application parsing across the whole AST). This commit gives the alpha meta its own type without that blast radius; happy to track the deeper split as its own issue if you'd like to go that far. Note the same `isPrefixOf` shape still lives in the `𝑝` primitive-expression meta from #802, which we could clean up the same way in a follow-up.

The `rewrite/normalize` regression is unrelated to this refactor — it's the extra `dca` rule being tried each pass, as flagged above; if anything, dropping the per-match `isPrefixOf` scan makes the matcher a touch cheaper.

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+207](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #818, opened 2026-06-18
Drop head/tail from morphing rules

### LINKED ISSUE #814: Refactor Morphing Rules to Exclude Head/Tail for Efficiency
Morphing rules may be rewritten without head/tail:

```
M(p) --> p

M([L>F]) --> M(n)  where n := F()

M(n.tau)  --> M(n).tau

M(n(tau -> n))  --> M(n)(tau -> n)

M(Q)  --> n    if Q -> n
```

We can define `M()` as only accepting normal form, not any exception possible.

Functions (in atoms) may only return normal forms.

Morphing rules may be applied in any other, until the stop applying.

yegor256:
This drops the `* t` head/tail construct from the morphing rules. The `lambda` and `root` rules now match their bare head, and `morph` carries the trailing dispatches and applications as a function instead, re-attaching them when a head rule fires. So `prim` and `nmz` judge the whole expression while `lambda`/`root` fire on the head, which keeps the evaluation order right — an atom buried under a tail fires only once its arguments have filled the voids.

The motivation is efficiency: the matcher no longer splits every expression into head and tail hunting for a match. A global dispatch `Φ.τ` stays a normal form that `morph` resolves, so recursive top-level objects still normalize without looping.

Behavior is unchanged. The full suite (1062 examples) passes, including the reduction-label-sequence check, and fourmolu and hlint are clean. There's also a small chore commit that ignores the built `phino` binary, which was tripping the reuse license check.

Closes #814


yegor256:
@maxonfjvipon let's merge?

PR #820, opened 2026-06-18
Restrict dataization and morphing to normal-form input

### LINKED ISSUE #819: D() and M() Functions Fail to Restrict to Normal Form Input Only
Let's tighten our rules.

yegor256:
This makes 𝔻 and 𝕄 accept only normal forms. The morphing 'nmz' rule, which used to quietly normalize whatever you fed it, is gone, and the catch-all clauses ('stuck', 'norm', 'none') now match the normal-form meta 'n'. Since the matcher rejects an 'n' bind whose value isn't already in normal form, a non-NF expression simply matches no clause — the functions never reduce it.

Normalization still happens, but now at the points where a rule actually builds a fresh expression: 'box' normalizes its contextualized body, 'lambda' and 'root' normalize their products, and the two dataize entry points normalize before they start reducing. Atom results are reduced to a normal form as well, so the built-in functions hand back normal forms too. The upshot is that the YAML specs now read the way the calculus is meant to work: 𝔻(n) and 𝕄(n), with 𝒩 spelled out wherever a new term appears.

One test that dataized a non-NF dispatch straight through D was dropped (that path is covered by the φ-wrapped case and the end-to-end Five/Located cases), and the reduction-trace fixtures shifted to the new step order. Full suite, hlint, fourmolu and coverage all pass.

Closes #819


yegor256:
@maxonfjvipon let's merge?

src/Dataize.hs
```diff
@@ -189,14 +185,40 @@ leadsTo ((prog, _) :| rest) rule expr DataizeContext{..} = do
...
+  (rewrittens, _) <- rewrite prog' normalizationRules (rewriteContext ctx)
+  let (rw :| rws) = NE.reverse rewrittens
+      seq' = rw :| rws <> NE.tail seq
+  expr' <- locatedExpression _locator (fst rw)
+  pure (expr', seq')
+
+-- Reduce 'expr' to its normal form, discarding the intermediate steps. Used to
+-- keep atom results in normal form before they re-enter morphing.
+toNormalForm :: Expression -> DataizeContext -> IO Expression
+toNormalForm expr ctx@DataizeContext{..} = fst <$> normalized expr ((_program, Nothing) :| []) ctx
```
maxonfjvipon: @yegor256 why do we need 2 functions for normalization?

src/Dataize.hs
```diff
@@ -114,33 +114,21 @@ morph (expr, seq) ctx@DataizeContext{..} = do
...
-        MtDisabled
-        Nothing
-        _saveStep
 
 dataize :: DataizeContext -> IO Dataized
 dataize ctx@DataizeContext{..} = do
   expr <- locatedExpression _locator _program
-  (maybeBytes, seq) <- dataize' (expr, (_program, Nothing) :| []) ctx
-  pure (maybeBytes, reverse seq)
+  (normal, seq) <- normalized expr ((_program, Nothing) :| []) ctx
```
maxonfjvipon: @yegor256 I'm not sure if we need normalization here. This is the input function, it must accept expression in normal form and should validate it, but not normalize

maxonfjvipon (changes requested):

yegor256:
The failing benchmark check is a false alarm, not a real regression.

`rewrite/normalize` is a tiny micro-bench (~22ms) and swings a lot between runs. Running the *exact same* binary four times in a row on my machine gives 21.6, 22.4, 24.4 and 24.7 ms — a ~14% spread on identical code, so a +40% blip on a shared CI runner is well within the noise. A same-machine A/B against master landed the head sample right inside master's own variance band.

This also can't be a real regression by construction: the PR doesn't touch the normalization path. `rewrite/normalize` exercises `rewrite prog normalizationRules`, and `Rewriter`, the `normalize/*.yaml` rules and `normalizationRules` are all unchanged here — the diff lives entirely in dataization/morphing. So it's safe to merge from my side; a re-run of the bench job should come back green.

src/Dataize.hs
yegor256: Good point — removed it. `dataize` no longer normalizes its input; it just hands the located expression to `dataize'`. NF is now enforced by the rules themselves: the catch-all clauses match the `𝑛` meta, so a non-NF, non-formation input matches no clause and yields ⊥ instead of being silently normalized.

I kept it as match-level validation rather than a hard `isNF`-or-throw at the boundary, because `box` matches φ-formations structurally and reduces their contextualized body (its own product, with the `𝒩` already in the rule). A throw would also reject a formation whose φ-body isn't in NF yet, which `box` currently handles. If you'd prefer it to reject any non-NF input outright, I'm happy to add the assertion — that just pushes normalization to the caller/CLI and I'd adjust the one `--sequence` test that feeds a non-NF formation.

src/Dataize.hs
yegor256: Collapsed into one — dropped `toNormalForm`. Everything now reduces through `normalized`; the atom-result path calls it directly and just keeps the resulting expression (`TeExpression . fst <$> normalized …`). The only reason it returns the step sequence at all is so `box`/`lambda`/`root` can splice their reduction steps into the `--sequence` chain; callers that don't need the chain (atoms) ignore it.

yegor256:
@maxonfjvipon take a look again please

src/Dataize.hs
```diff
@@ -250,7 +266,7 @@ _lambda ctx [ArgExpression expr] subst = do
     ExFormation bds -> do
       resolved <- formation bds ctx
       case resolved of
-        Just obj -> pure (TeExpression obj)
+        Just obj -> TeExpression . fst <$> normalized obj ((ctx._program, Nothing) :| []) ctx
```
maxonfjvipon: @yegor256 why do we need `normalized` here. Normalization should be called outside of 'lambda' function call.

src/Dataize.hs
maxonfjvipon, line 280: @yegor256 you've changed the description of `root` rule: `𝕄(Φ.τ * t) ⟿ 𝕄(𝒩(e * t))`
There's a normalization called, but I don't see it here

maxonfjvipon (changes requested):

src/Dataize.hs
yegor256: Agreed — removed it. `lambda()` now just applies the atom and returns the result; `_lambda` is back to `pure (TeExpression obj)`. The normalization is done by the rule outcome instead (`then: morph: { normalize: 'e * t' }`, i.e. `𝕄(𝒩(F(...) * t))`), so it happens outside the function call, same as `global()`/`root`. It was also redundant, since that outcome already wraps the result in `𝒩`. All tests still pass.

src/Dataize.hs
yegor256: Right — `global()` (this function) deliberately doesn't normalize; it only resolves the Φ-dispatch and returns the bound expression. The `𝒩` from the `root` description is applied by the rule's outcome in `morphing.yaml`:

```yaml
- name: root
  match: 'Φ.𝜏 * !t'
  where:
    - meta: 𝑒
      function: global
      args: [𝜏]
  then:
    morph:
      normalize: '𝑒 * !t'    # ← 𝕄(𝒩(e * t))
```

That `normalize:` outcome is handled in `morph`'s `MaNormalize` branch, which runs the rewriter over `e * t` and splices the steps in. With the `lambda()` change above, `lambda` and `root` are now symmetric: the builder resolves, the rule normalizes.

yegor256:
@maxonfjvipon let's try again

maxonfjvipon (approved):

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+248](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (155 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+519](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #826, opened 2026-06-22
Drop head/tail from morphing for a recursive 𝕄/𝔼/𝔻 scheme

### LINKED ISSUE #814: Refactor Morphing Rules to Exclude Head/Tail for Efficiency

yegor256:
This swaps the head/tail mechanism in the morphing function for a recursive rule set. Morphing now walks a normal form down to its primitive one operation at a time — peeling each dispatch and application and resolving the universe Φ through normalization — instead of matching an arbitrary trailing `* t` chain in a single step. A new evaluation relation 𝔼 fires atoms, so 𝕄 hands back a saturated λ-formation untouched and 𝔻 fires it through a new `fire` rule.

The `* t` construct existed only to serve those two morphing rules, so it is gone everywhere: the AST node, the parser, the CST and its renderers, the matcher, and the builder. The net diff removes more than it adds.

Dataization is unchanged where it counts — every byte-result fixture (`5.plus(6)`, Fahrenheit, Factorial, and the located cases) produces the same bytes as before. The visible differences are in `explain --morph`/`--dataize` output and in the ρ-context now carried by a `Q`-dispatch (`Q.x` morphs to `⟦ρ ↦ Q⟧` rather than a bare `⟦⟧`), and the regenerated goldens reflect that. Tests, hlint, fourmolu, and coverage all pass.

Closes #814


resources/morphing.yaml
```diff
@@ -1,37 +1,57 @@
...
   then:
     morph:
-      normalize: '𝑒 * !t'
+      normalize: '𝑒.𝜏'
+
+- name: dispatch
+  description: '𝕄(n.τ) ⟿ 𝕄(𝒩(𝕄(n).τ))'
+  match: '𝑛.𝜏'
+  then:
+    morph-head: '𝑛.𝜏'
```
maxonfjvipon: @yegor256 I don't like the idea with this special `morph-head`. It would be better to do something like this instead:
```
match: n.t
where:
  - meta: e
    function: morph
    args: n
  - meta: n1
    function: normalize
    args: e.t
then:
  morph: n1

# OR

match: n.t
where:
  - meta: e
    function: morph
    args: n
then:
  morph:
    normalize: e.t

```

yegor256:
@maxonfjvipon wdyt?

yegor256:
@maxonfjvipon now, our rules are the following (no head+tail anymore):

```
Evaluation:

𝔼(⟦…,λ⤍F,…⟧)  ⟿  F(⟦…,λ⤍F,…⟧)

Dataization 𝔻 (dataization.yaml) — top-to-bottom, first match wins:

delta  𝔻(⟦…,Δ⤍δ,…⟧) ⟿ δ
box    𝔻(⟦…,φ↦e,…⟧)  ⟿ 𝔻(𝒩(𝒞(e)))   if [Δ,λ] ∩ rest = ∅
fire   𝔻(⟦…,λ⤍F,…⟧)  ⟿ 𝔻(𝒩(𝔼(…)))
none   𝔻(⟦B⟧)        ⟿ nothing
bott   𝔻(⊥)          ⟿ nothing
norm   𝔻(n)          ⟿ 𝔻(𝕄(n))

Morphing 𝕄 (morphing.yaml) — walks a normal form to its primitive one op at a time:

prim          𝕄(⟦𝐵⟧)       ⟿ ⟦𝐵⟧
lambda        𝕄(⟦…λ⤍F…⟧.τ) ⟿ 𝕄(𝒩(𝔼(⟦…λ⤍F…⟧).τ))
dispatch      𝕄(n.τ)       ⟿ 𝕄(𝒩(𝕄(n).τ))
application   𝕄(n(τ↦e))    ⟿ 𝕄(𝒩(𝕄(n)(τ↦e)))
applicationa  𝕄(n(η↦e))    ⟿ 𝕄(𝒩(𝕄(n)(η↦e)))
root          𝕄(Φ)         ⟿ 𝕄(𝒩(e)), e := global()
stuck         𝕄(n)         ⟿ ⊥
```

We also don't have primitives anymore.

maxonfjvipon (changes requested):

resources/morphing.yaml
yegor256: Good call, done in fb09469. The dispatch/application/applicationa rules now bind `morph(n)` in a `where` and reuse the plain `morph: { normalize: ... }` outcome, so `morph-head` and its special driver branch are gone. Went with your second variant since it folds the normalize into the existing outcome and only needs the one new `morph` function.

One heads-up: a `where` function returns just a term, not a step sequence, so the head-morph sub-steps no longer get spliced into the reduction chain and the explain/label output is shorter. The final bytes are unchanged. If you'd rather keep those intermediate steps visible in the chain, that needs a different approach, let me know.

yegor256:
@maxonfjvipon check again please

resources/morphing.yaml
```diff
@@ -1,37 +1,75 @@
...
   then:
     morph:
-      normalize: '𝑒 * !t'
+      normalize: '𝑒.𝜏'
+
+- name: dispatch
+  description: '𝕄(n.τ) ⟿ 𝕄(𝒩(𝕄(n).τ))'
+  match: '𝑛.𝜏'
+  where:
+    - meta: 𝑒1
```
maxonfjvipon: @yegor256 we can use just `𝑒` everywhere where it's possible, not `𝑒1`

src/Dataize.hs
```diff
@@ -122,7 +106,9 @@ morph (expr, seq) ctx@DataizeContext{..} = do
       built <- buildExpressionThrows arg subst
       labelled <- leadsTo seq name built ctx
       (expr', seq') <- normalized built labelled ctx
-      morph (expr', seq') ctx
+      if expr' == expr
+        then pure (ExTermination, seq')
```
maxonfjvipon: @yegor256 I'm not sure this is correct, at least this logic is absent in yaml rules

maxonfjvipon (changes requested):

resources/morphing.yaml
yegor256: Done in e9ffb75. `dispatch` now uses `𝑒`, like `lambda` and `root`. I kept `𝑒1` in `application` and `applicationa`: `𝑒` is the only unconstrained expression meta (`𝑛` forces a normal form, `𝑘` forces absolute), and in those two rules `𝑒` already names the argument, so the morphed head needs its own name.

src/Dataize.hs
yegor256: This one is load-bearing — I removed it to check, and the suite hangs. It fires when `𝒩(𝕄(n)…)` comes back equal to the input: the dispatch/application cannot reduce any further, so morphing is stuck and the answer is `⊥`, exactly what the `stuck` rule yields. The reason it is not a yaml rule is that it is a post-reduction check — the result is only known after the `then`/`𝒩` step runs, whereas `when`/`having`/`where` all run before it, so no clause can express it. I expanded the comment in e9ffb75 to spell that out. Happy to revisit if you see a cleaner way to fold it into the rules.

yegor256: You were right, and my earlier reply here was wrong — it does belong in the rules. I instrumented the guard and ran the whole suite: it fires exactly once, only for the `root` rule on a self-referential program where `global()` returns Q itself. It was never a general dispatch/application thing. In 85c6a2c the `root` rule carries it as a `when` (`e ≠ Q`): a universe that resolves to Q is skipped, so `stuck` produces the `⊥`, and the whole thing shows up in `explain --morph`. The Haskell guard is deleted. dispatch and application never reach a fixpoint, so they stay as they were and their step chains are unchanged.

yegor256:
@maxonfjvipon one more time please

maxonfjvipon (approved):

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+255](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (496 >= 100); -8 for way too many hits-of-code (496 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+523](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #830, opened 2026-06-22
Replace the index() condition function with an αi construct

### LINKED ISSUE #828: `index` function redundancy with `αi` function
Let's get rid of it and use `αi` instead.

yegor256:
This drops the `index` condition function (which rendered as `\indexof`) and replaces it with an `α𝑖` argument that captures a positional index straight into a meta variable. The alpha normalization rule now matches `α𝑖 ↦ e` and guards it with a bare `i = |B̄₁|`, where before it matched `𝜂 ↦ e` and wrote `\indexof{𝜂} = |B̄₁|`. So the rule explains as `\alpha_{i} -> e` with `if $ i = ... $`.

The `index` function only ever existed to pull the index back out of a matched α attribute, so it was redundant. Binding the index at the point the argument is matched lets us reference it directly in conditions and removes the auxiliary function from the matcher, the YAML/textual condition DSLs and the LaTeX rendering. The branch also carries an earlier commit that was sitting on master refactoring the alpha pattern; the two belong together since this one updates the same rule and its expected output. Closes #828.

One thing for the reviewer: while writing the test packs I ran into a latent parser quirk — a unicode meta followed by a space and an ASCII `->` loses the `-` to the meta-name suffix, so `(𝜂 -> e)` fails to parse while `(𝜂 ↦ e)` works. It predates this change and affects every unicode meta, so I left it alone and used `↦` in the packs; I'll file it separately.


yegor256:
@maxonfjvipon this is shorter and nicer in LaTeX, please check

resources/normalize/alpha.yaml
```diff
@@ -2,9 +2,9 @@
 # SPDX-License-Identifier: MIT
 ---
 name: alpha
-pattern: ⟦𝐵1, 𝜏1 ↦ ∅, 𝐵2⟧(𝜂 ↦ 𝑒)
-result: ⟦𝐵1, 𝜏1 ↦ ∅, 𝐵2⟧(𝜏1 ↦ 𝑒)
+pattern: ⟦𝐵1, 𝜏 ↦ ∅, 𝐵2⟧(α𝑖 ↦ 𝑒)
+result: ⟦𝐵1, 𝜏 ↦ ∅, 𝐵2⟧(𝜏 ↦ 𝑒)
 when:
   eq:
-    - index: 𝜂
```
maxonfjvipon, line 9: @yegor256 let's remove `𝜂` from syntax then and rename everything with 'alpha-meta', 'meta-idx' to just 'meta'

maxonfjvipon (changes requested):

resources/normalize/alpha.yaml
yegor256: Done in 5e0edf8 — `𝜂`/`!h` is removed from the grammar entirely, and every rule (morphing's `applicationa`, `dca`, and the condition packs) now matches positional arguments with `α𝑖`, even where the index isn't read.

I reclaimed the freed name, so `AlIndex` → `AlMeta`: there's now a single alpha-meta construct. The CST nodes `AT_ALPHA_META` and `IDX_META` stay as-is, since `AT_META` is already the attribute meta (`𝜏`/`!a`) and `META` is the record type — happy to pick specific names there if you'd like.

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

yegor256:
@maxonfjvipon I'll fix it now

yegor256:
@rultor merge

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+233](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (162 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+473](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #837, opened 2026-06-23
Rename meta-variable syntax and model alpha as a dedicated CST type

### LINKED ISSUE #835: Incorrect Use of `!t` and `a` Meta Variables in Attributes and Alpha Attributes
After #830 and #818 we freed the `!t` meta which was used for tail. Let's use `!t` instead of `!a` for attributes, since it's closer to `\tau`, then let's use `a` (without `!`) instead of `~` for alpha attributes with number or meta index. Also let's add pretty alternative for `!F` for meta functions

maxonfjvipon:
Closes #835.

## Syntax changes
- attribute meta `!a` → `!t` (closer to `τ`)
- positional/alpha argument `~` → `a` (e.g. `a0`, `a!i`)
- function meta gains the pretty unicode form `𝐹` alongside `!F`; the built-in morphing/dataization rules now use `𝐹`

The unicode forms (`𝜏`, `α𝑖`) are unchanged, so the built-in `resources/` rules needed no syntax migration beyond adopting `𝐹`.

## Refactor (from review of the change)
- model AST `Alpha` as a dedicated CST `ALPHA` type with its own `PA_ALPHA` pair, instead of overloading `ATTRIBUTE` (which previously carried `AT_ALPHA`/`AT_ALPHA_META` constructors that had no AST `Attribute` counterpart)
- as a result, an alpha index now renders as `\alpha_{n}` in LaTeX, matching the existing `\alpha_{i}` meta form (it previously fell through the LaTeX catch-all as raw `~n`/`an`)

## Notes
- backward-incompatible: rule files using `!a` or `~` must migrate to `!t`/`a`
- in application-argument position `a0` parses as an alpha index; elsewhere it is a label

## Tests
- migrated all test-resource packs, specs and the README
- added coverage for `𝐹` (parse + sweet render), ASCII alpha-meta `a!i`, and the LaTeX alpha rendering
- `make test` (with `-Werror`): 1047 examples, 0 failures; fourmolu and hlint clean


maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (594 >= 100); -8 for way too many hits-of-code (594 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+265](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #838, opened 2026-06-23
bug(#824): stop randomString looping on literal-only patterns

### LINKED ISSUE #824: `randomString` loops forever on second call with literal-only pattern (`src/Random.hs`)
`randomString` from `src/Random.hs` enters an infinite loop on the second call when its pattern contains no `%d` or `%x` placeholder.

The function keeps an unbounded `IORef (Set String)` of every value it has ever returned (lines 14-16), and `regenerate` (lines 33-40) recurses with the same pattern until it produces a value not already in that set. When the pattern is a plain literal like `"hello"`, or any string whose `%`-escapes are all unknown (for example `"%q"`), `generate` is deterministic: it always yields the same output. The first call stores that output in the set, and the next call with the same pattern recurses forever looking for a fresh value that cannot exist.

This is reachable from the `random-string` built-in function exposed in `src/Functions.hs:143-148`, so user rules that invoke `random-string` with a constant or with only unknown `%`-codes will hang the rewriter. The test suite hides the bug because each `describe` block calls a unique literal once per process, so the second collision never occurs.

A small fix is to detect, in `generate`, whether the pattern contains any randomized placeholder (`%d`, `%x`); when it does not, skip the uniqueness loop and return the pattern unchanged. The same change would also stop the unbounded `strings` set from growing for literal-only patterns. A separate but related concern is that the global `strings` set leaks for the lifetime of the process; capping it or scoping it per run would be worth a follow-up.

The location to change is `src/Random.hs`, around lines 18-45.

maxonfjvipon:
Closes #824.

## Problem

`randomString` (`src/Random.hs`) always routed through `regenerate`, which recurses until it produces a value not already in the global `strings` set. For a pattern with no `%d`/`%x` placeholder (a plain literal like `"hello"`, or one whose `%`-escapes are all unknown like `"%q"`), `generate` is deterministic — it always yields the same output. The first call caches it, and the second call with the same pattern loops forever looking for a fresh value that cannot exist. This is reachable via the `random-string` built-in, so a user rule with a constant argument hangs the rewriter.

The existing suite hid the bug because each `describe` block used a unique literal only once per process, so the second collision never happened.

## Fix

Skip the uniqueness loop when the pattern contains no randomized placeholder, returning `generate pat` directly. This also stops the global `strings` set from growing for literal-only patterns.

## Test

Added a timeout-guarded test that calls `randomString` twice with the same literal pattern. Verified it fails (timeout → `Nothing`) against the unpatched code and passes with the fix.

## Follow-up

The global `strings` set still leaks for the lifetime of the process; capping or scoping it per run is worth a separate issue, as noted in #824.

### DIFF
--- a/src/Random.hs
+++ b/src/Random.hs
@@ -40,6 +40,11 @@ regenerate pat set = do
       pure next
 
 randomString :: String -> IO String
-randomString pat = do
-  set <- readIORef strings
-  regenerate pat set
+randomString pat
+  | randomized pat = readIORef strings >>= regenerate pat
+  | otherwise = generate pat
+  where
+    randomized :: String -> Bool
+    randomized [] = False
+    randomized ('%' : ch : rest) = ch == 'd' || ch == 'x' || randomized rest
+    randomized (_ : rest) = randomized rest
--- a/test/RandomSpec.hs
+++ b/test/RandomSpec.hs
@@ -12,6 +12,7 @@ module RandomSpec where
 import Data.Char (isDigit, isHexDigit)
 import Data.Set qualified as Set
 import Random (randomString)
+import System.Timeout (timeout)
 import Test.Hspec (Spec, describe, it, shouldBe, shouldSatisfy)
 
 spec :: Spec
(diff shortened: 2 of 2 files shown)


maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+277](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #857, opened 2026-06-24
bug(#856): make morphing `dispatch` disjoint from `lambda`

### LINKED ISSUE #856: Morphing `lambda` and `dispatch` rules overlap, making rule order load-bearing
In `resources/morphing.yaml` the `lambda` rule (`match: ⟦𝐵1, λ ⤍ 𝐹, 𝐵2⟧.𝜏`) is a strict special case of the `dispatch` rule (`match: 𝑛.𝜏`): a λ-formation dispatch matches both. Morphing applies rules top-to-bottom with first-match-wins (`firstMatch` in `src/Dataize.hs:81-87`), so `lambda` only fires because it is listed before `dispatch`. The two rules overlap, and their relative order is an undocumented contract.

This is fragile because the overlap makes morphing order-dependent for no structural reason: swap the two rules and every λ-formation dispatch silently routes through `dispatch` instead of `lambda`. The rules should be mutually exclusive so their order cannot change behavior.

The fix is to add a `when` guard to `dispatch` that holds only when the head 𝑛 is not a λ-bearing formation, making it the exact complement of `lambda`. The condition vocabulary already supports this — `not` composed with `matches`, and `disjoint` already recognizes `λ` via `AtLambda`/`BiLambda` (`src/Rule.hs:224`) — so no new primitive is needed. With the guard in place `lambda` and `dispatch` become disjoint and order-independent; the `explain` snapshot at `test/CLISpec.hs:1061` and a test asserting a λ-formation dispatch still morphs identically would need to follow.

yegor256:
Fixes #856.

## Problem

In `resources/morphing.yaml` the `lambda` rule (`⟦𝐵1, λ ⤍ 𝐹, 𝐵2⟧.τ`) is a
strict special case of the `dispatch` rule (`𝑛.τ`): a λ-formation dispatch
matches **both**. Morphing applies rules top-to-bottom with first-match-wins,
so `lambda` only fires because it is listed before `dispatch`. The overlap made
the relative order of the two rules an undocumented, load-bearing contract —
swapping them would silently route every λ-formation dispatch through
`dispatch` instead of `lambda`.

## Fix

Add a `when` guard to `dispatch` so it becomes the exact complement of
`lambda`: it now holds only when the head `𝑛` is **not** a formation. The
guard is expressed with a new first-class `binding` condition:

```yaml
- name: dispatch
  match: '𝑛.𝜏'
  where:
    - meta: 𝑒
      function: morph
      args:
        - 𝑛
  when:
    not:
      binding: 𝑛
  then:
    morph:
      normalize: '𝑒.𝜏'
```

`binding 𝑛` holds when `𝑛` resolves to a formation (an abstraction `⟦…⟧`);
`not (binding 𝑛)` therefore holds for any non-formation head. Because *every*
formation head — λ-bearing or not — is excluded, `dispatch` and `lambda`
(together with `prim`) become mutually exclusive and order-independent.

Choosing `not (binding 𝑛)` over the narrower `λ ∉ bindings(𝑛)` keeps the rule
robust: it does not depend on enumerating the head's bindings and naturally
handles arbitrary heads (`Φ`, `ξ`, a dispatch chain, …), which are simply not
formations and so still morph.

## Changes

- `resources/morphing.yaml` — guard `dispatch` with `not (binding 𝑛)`; document
  the disjointness invariant in the header comment.
- `src/Yaml.hs` — add the `IsBinding` constructor to `Condition`, parse the
  `binding` key, and register it in `validateYamlObject`.
- `src/Condition.hs` — parse `binding( … )` in the textual condition grammar.
- `src/Rule.hs` — evaluate the condition: `_isBinding` resolves a meta and holds
  when the bound expression is an `ExFormation`.
- `src/CST.hs`, `src/Render.hs`, `src/Encoding.hs`, `src/LaTeX.hs`,
  `src/Lining.hs`, `src/Sugar.hs` — thread the new `CO_BINDING` CST node through
  the rendering pipeline (`binding( … )` in text, recursive transforms for
  ASCII / LaTeX / single-line / salty).
- `README.md` — document the `binding` condition.

## Tests

- `test/ConditionSpec.hs` — parse/round-trip cases for `not(binding(!n))`.
- `test/CLISpec.hs` — updated the `explain --morph` snapshot so `dispatch` now
  shows its `not( binding( n ) )` guard.
- `test/DataizeSpec.hs` — new `morphing 'dispatch' is disjoint from 'lambda'`
  group: `dispatch` does **not** fire on a λ-bearing formation dispatch, **does**
  fire on a non-formation (`ξ`) dispatch, and a chained `⟦λ ⤍ F⟧.a.b.c` dispatch
  drills down through `dispatch → dispatch → lambda` so the base λ-formation
  reaches `lambda` (proven by the atom-firing error on the undefined `F`).

## Note

This branch is merged up to current `master`, which had independently added an
optional rule `label` field, a `gt` condition, and a `⊥`-reduction rule. Merge
conflicts in `src/Yaml.hs` (combine both new condition keys), `test/CLISpec.hs`
(master's new morphing LaTeX style + this branch's guard), and the `asRule` test
helper (updated for the new 9-field `Yaml.Rule` constructor) were resolved. The
net diff above is exactly the #856 feature changes — no master code is altered.

https://claude.ai/code/session_01JJDDzLCbdQH6sKCMGFaDPV

---
_Generated by [Claude Code](https://claude.ai/code/session_01JJDDzLCbdQH6sKCMGFaDPV)_


yegor256:
@maxonfjvipon take it look. this should help us make morphing rules order independent

resources/morphing.yaml
```diff
@@ -34,6 +39,9 @@
       function: morph
       args:
         - 𝑛
+  when:
+    not:
+      binding: 𝑛
```
maxonfjvipon: @yegor256 binding is 'X -> Y', maybe it would be better to rename it to 'formation'?

maxonfjvipon (changes requested):

yegor256:
@maxonfjvipon fixed, look again

README.md
```diff
@@ -310,6 +310,10 @@ Condition:
   | part-of:             # returns True if given expression is attached to any
       - Expression'      # attribute in ginve bindings
       - BiMeta'
+  | formation: Expression' # returns True if given expression is a formation
```
maxonfjvipon: @yegor256 let's indent all the comments to the left so they are tabulated in the same way

maxonfjvipon (changes requested):

resources/morphing.yaml
yegor256: Good point — renamed to `formation` throughout: the YAML key (`formation:`), the AST/CST (`IsFormation`/`CO_FORMATION`), the condition evaluator (`_isFormation`), the textual grammar (`formation(...)`), and the LaTeX macros (`\phinoIsFormation` / `\phinoNotFormation`). Done in 176e1ab.

---

README.md
yegor256: Done in 59db016. Moved `Expression'` onto its own line (like the `matches`/`part-of` entries) and aligned every comment at the standard column, so all the `#` columns now line up with the rest of the grammar block.

---

maxonfjvipon (approved):

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+358](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (100 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+949](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #859, opened 2026-06-24
fix(#858): fail loudly when no morphing/dataization rule matches

### LINKED ISSUE #858: Morphing silently returns input unchanged when no YAML rule matches, instead of failing
The morphing function `morph` (`src/Dataize.hs:74-79`) tries the YAML rules in order and, when none matches (`firstMatch` returns `Nothing`), silently returns its input unchanged: `Nothing -> pure (expr, seq)`. This branch is behavior that exists only in Haskell — it is not expressed by any rule in `resources/morphing.yaml`. The identical pattern sits in `dataize'` (`src/Dataize.hs:130`), which returns "no data" on no-match.

This is wrong because the YAML rule set is meant to be the single source of truth for 𝕄, and the `morphing.yaml` rules already end with a catch-all `stuck` (`match: 𝑛`) that matches every expression — so the `Nothing` branch is unreachable in correct operation and is pure dead code. Worse, it is a silent fallback: if the rule set ever loses its catch-all (an edit, a future refactor), `morph` would quietly return its argument unchanged — a no-op the rules never sanctioned — instead of surfacing that the rule set is incomplete. That violates fail-fast: a reachable "no rule matched" state is an invariant violation, not a default to swallow.

The fix is to make the no-match branch raise instead of returning: replace `Nothing -> pure (expr, seq)` at `src/Dataize.hs:79` with `Nothing -> throwIO (userError "no morphing rule matched")` (`throwIO` is already imported). The sibling at `src/Dataize.hs:130` should do the same so dataization fails loudly rather than reporting empty data.

yegor256:
Closes #858.

## Problem

`morph` (`src/Dataize.hs:79`) and `dataize'` (`src/Dataize.hs:130`) silently returned their input unchanged / "no data" when `firstMatch` returned `Nothing`. This is behavior that exists only in Haskell — it is not expressed by any rule in `resources/morphing.yaml` or `resources/dataization.yaml`.

Both rule sets already end with a catch-all rule (`match: 𝑛`) that matches every expression, so the `Nothing` branch is unreachable dead code in correct operation. As a silent fallback it also violates fail-fast: if the rule set ever loses its catch-all, these functions would quietly return a no-op the rules never sanctioned instead of surfacing that the rule set is incomplete.

## Fix

- `src/Dataize.hs:79`: `Nothing -> throwIO (userError "no morphing rule matched")`
- `src/Dataize.hs:130`: `Nothing -> throwIO (userError "no dataization rule matched")`

`throwIO` was already imported. A reachable "no rule matched" state is now treated as the invariant violation it is.

## Testing

The local environment has no Haskell toolchain (`cabal`/`ghc` unavailable), so `make test` could not be run here — CI will exercise the suite. The change touches only an unreachable branch, and no existing test relies on the old silent-fallback behavior.

https://claude.ai/code/session_016z3NG3LvWzE7oCANgv8K4k

---
_Generated by [Claude Code](https://claude.ai/code/session_016z3NG3LvWzE7oCANgv8K4k)_

### DIFF
--- a/src/Dataize.hs
+++ b/src/Dataize.hs
@@ -76,7 +76,7 @@ morph (expr, seq) ctx@DataizeContext{..} = do
   matched <- firstMatch Y.morphingRules
   case matched of
     Just (rule, subst) -> apply rule.then_ rule.name subst
-    Nothing -> pure (expr, seq)
+    Nothing -> throwIO (userError "no morphing rule matched")
   where
     firstMatch :: [Y.MorphRule] -> IO (Maybe (Y.MorphRule, Subst))
     firstMatch [] = pure Nothing
@@ -127,7 +127,7 @@ dataize' (expr, seq) ctx = do
   matched <- firstMatch Y.dataizationRules
   case matched of
     Just (rule, subst) -> apply rule subst
-    Nothing -> pure (Nothing, NE.toList seq)
+    Nothing -> throwIO (userError "no dataization rule matched")
   where
     firstMatch :: [Y.DataizeRule] -> IO (Maybe (Y.DataizeRule, Subst))
     firstMatch [] = pure Nothing


yegor256:
@maxonfjvipon merging?

yegor256:
The failing `regression-check` is the known-flaky `rewrite/normalize` benchmark, and this PR's change cannot be the cause:

- This PR only edits `morph`/`dataize'` in `src/Dataize.hs` (the no-match branch, formerly dead code). The `rewrite/normalize` benchmark runs `rewrite prog normalizationRules` (`benchmark/Main.hs:93`), which exercises `Rewriter.hs` — and `Rewriter.hs` does not import `Dataize` (only `CLI/Runners.hs` does, via the `Deps.hs` inversion). So the executed code on the normalize path is unchanged by this PR.
- `rewrite/normalize` is the fastest benchmark (~44–57 ms base) with ~19% std dev, so it's the most noise-prone. Across both CI runs head landed ~65–70 ms while base bounced 43–57 ms — a same-runner ordering artifact (head always runs second), not a code effect. The merge-base commit was itself titled *"Re-trigger CI (flaky rewrite/normalize benchmark)"*.

I held off on a third empty re-trigger commit since the noise isn't reliably clearing. This red check appears safe to merge through; happy to re-trigger again if preferred.

---

maxonfjvipon (approved):

maxonfjvipon:

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (4) hits-of-code; +4 to give you at least something.  Your running score is [+313](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (4) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+727](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #861, opened 2026-06-24
Split catch-all `stuck` morphing rule into `⊥`/`ξ`/`Φ` rules

### LINKED ISSUE #860: Catch-all `stuck` morphing rule is order-dependent; split into fine-grained `⊥`/`ξ`/`Φ` rules
In `resources/morphing.yaml:79` the `stuck` rule (`match: 𝑛`, `then: ⊥`) matches *every* expression. It works only because it is listed last: it is a position-dependent sink, and like the `lambda`/`dispatch` overlap in #856 it makes morphing order load-bearing. We want all morphing rules to be order-independent, and the way to get there is to simplify the catch-alls into fine-grained, mutually-exclusive rules that match concrete shapes rather than "anything".

The `𝑛` match is the worst offender: besides depending on its position, it silently turns any genuinely-unexpected normal form into `⊥` instead of failing loudly — the opposite of the direction set by #858 (commit `e8750bf`, "fail loudly when no morphing/dataization rule matches"). Instrumenting `apply` in `src/Dataize.hs:94` across the whole suite shows `stuck` only ever receives three concrete inputs: `⊥` (`ExTermination`), `ξ` (`ExXi`), and `Φ` (`ExRoot`, on the `root` rule's `when` fall-through when the universe is `Φ` itself).

The fix is to replace `stuck` with three literal rules — `dead` (`⊥ ⟿ ⊥`), `xi` (`ξ ⟿ ⊥`), and `globe` (`Φ ⟿ ⊥`, kept after `root`) — so morphing no longer has an "anything" sink and any other unexpected input fails loudly. The matcher already resolves these literals (`src/Matcher.hs:107-109`); the `explain --morph` snapshot at `test/CLISpec.hs:1061` and the `morph` cases at `test/DataizeSpec.hs:56-58` already pin the three outcomes. Sibling effort to #856.

yegor256:
The `stuck` rule in `resources/morphing.yaml` matched `𝑛` — every expression — and produced `⊥`. It only behaved because it sat last in the list, which made the whole morphing relation order-dependent and quietly swallowed any unexpected normal form as `⊥`. This drops it in favour of three literal rules: `dead` (`⊥ ⟿ ⊥`), `xi` (`ξ ⟿ ⊥`) and `globe` (`Φ ⟿ ⊥`, kept right after `root` so it only catches the self-referential-universe fall-through).

The goal is to make morphing rules order-independent by matching concrete shapes instead of "anything". Instrumenting the suite showed those three are the only inputs `stuck` ever received, so behaviour for valid programs is unchanged; anything genuinely unexpected now fails loudly, in line with #858. The `morph` cases in `test/DataizeSpec.hs` already pin the three outcomes and the `explain --morph` snapshot in `test/CLISpec.hs` is updated to list the new rules.

This is a step toward the broader order-independence effort tracked in #856. Closes #860.


yegor256:
@maxonfjvipon let's merge?

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

maxonfjvipon:

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+317](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (34) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+731](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #876, opened 2026-06-25
bug(#875): restore ~ as the ASCII alpha prefix

### LINKED ISSUE #875: ASCII alpha prefix 'a' shadows attribute labels a0, a1, … in argument position (regression from #835)
Since #835 (PR #837) renamed the ASCII alpha prefix from `~` to `a`, an attribute whose label is `a` followed only by digits (`a0`, `a1`, `a2`, …) is silently swallowed as a positional alpha in application-argument position and can no longer be passed as a named argument:

```
$ echo '{[[ ex -> [[ x -> ? ]] ( a1 -> 42 ) ]]}' | phino rewrite
Φ ↦ ⟦ ex ↦ ⟦ x ↦ ∅, ρ ↦ ∅ ⟧( α1 ↦ ... ), ρ ↦ ∅ ⟧
```

The argument `a1 ↦ 42` becomes positional `α1 ↦ 42` rather than a binding to the attribute named `a1`. The same label stays a normal label everywhere else — `[[ a1 -> 42 ]]` and `$.a1` both keep it — so only the argument slot is affected, and the reinterpretation is silent.

The cause is that `argument` (`src/Parser.hs:357`) tries `alpha` before the `attribute` label, and `alpha` (`src/Parser.hs:344`) now accepts the prefix `a`, which is itself a legal label-start character (`label'`, `src/Parser.hs:83`). The old `~` prefix could never collide because `~` cannot start a label, and the Unicode `α` is safe for the same reason. The rename traded a disjoint sigil for one that overlaps the label namespace, so the `a[0-9]+` slice of attribute names is now unreachable as a named argument.

The fix is to restore `~` as the ASCII alpha prefix — revert `symbol "a"` back to `symbol "~"` at `src/Parser.hs:344` (and the matching index-meta example to `~!i`) — so positional arguments use a sigil that can never be a label and `a0`/`a1`/… are labels again.

yegor256:
Fixes #875.

## Problem

Since #835 (PR #837) the ASCII alpha (positional argument) prefix was renamed from `~` to `a`. But `a` is itself a legal label-start character (`label'`), so it overlaps the label namespace.

In application-argument position, `argument` (`src/Parser.hs`) tries `alpha` before the `attribute` label. With `a` as the alpha prefix, an attribute whose label is `a` followed only by digits (`a0`, `a1`, …) was silently swallowed as a positional alpha and could no longer be passed as a named argument:

```
$ echo '{[[ ex -> [[ x -> ? ]] ( a1 -> 42 ) ]]}' | phino rewrite
Φ ↦ ⟦ ex ↦ ⟦ x ↦ ∅ ⟧( α1 ↦ ... ), … ⟧   # a1 became positional α1
```

The same label stayed a normal label everywhere else (`[[ a1 -> 42 ]]`, `$.a1`), so only the argument slot was affected, and the reinterpretation was silent. The old `~` prefix could never collide because `~` cannot start a label, and the Unicode `α` is safe for the same reason.

## Fix

Restore `~` as the ASCII alpha prefix:

- `src/Parser.hs`: `symbol "a"` → `symbol "~"` in `alpha`.
- `src/Render.hs`: `render ALPHA' = "a"` → `"~"`, so ASCII-encoded output round-trips back through the parser.
- Updated comments/docs (`src/CST.hs`, `src/Sugar.hs`, `README.md`).

`a0`/`a1`/… are now labels again everywhere, including the argument slot, while positional alphas use `~0`/`~!i` (or the Unicode `α0`/`α𝑖`).

## Tests

- Migrated all affected test packs (parser, condition, cst, rewriter) and specs from the `a`-prefix back to `~` where they encode positional alphas. Unicode `α` fixtures are untouched.
- Restored the `~0`/`~1`/`~123` → not-an-attribute coverage in `parse attribute`.
- Added a regression in `ParserSpec`: `Q.foo(a1 -> Q.y)` now parses `a1` as a named binding (`ArTau (AtLabel "a1")`), not a positional alpha.

> Note: no Haskell toolchain is available in this environment, so the changes were verified by reasoning through the parser/renderer and each affected fixture rather than by running `make test` locally. CI will confirm.

https://claude.ai/code/session_019wzmZyHRbpehazp3ex2noa

---
_Generated by [Claude Code](https://claude.ai/code/session_019wzmZyHRbpehazp3ex2noa)_


yegor256:
@maxonfjvipon let's merge?

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+329](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+867](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #874, opened 2026-06-25
Reduce out-of-range positional application ⟦𝐵⟧(α𝑖 ↦ 𝑒) to ⊥ (#870)

yegor256:
Fixes #870.

## Problem

An application of a formation to a positional argument whose index is at or beyond the formation's domain never reduced, so the term was wrongly treated as a normal form:

```console
$ printf 'Φ ↦ ⟦ x ↦ ∅ ⟧(α2 ↦ ⟦⟧)\n' | phino rewrite --normalize --flat
Φ ↦ ⟦ x ↦ ∅, ρ ↦ ∅ ⟧( α2 ↦ ⟦ ρ ↦ ∅ ⟧ )    # stuck; only x (0) and ρ (1) exist
```

`alpha` resolves the in-range case (`𝑖 = domain(𝐵1)` for a void slot), but nothing handled an index past the last attribute. Like `miss`/`over` send a misapplied *named* attribute to `⊥`, an out-of-range *positional* application is a dead term, not a normal form.

## Fix

- Add a `gt` ("greater-than") comparator to the condition vocabulary — the only ordering comparator was `eq`:
  - `src/Yaml.hs`: `Gt` constructor, parser branch, and `validateYamlObject` entry.
  - `src/Condition.hs`: `gt(...)` parser branch.
  - `src/Rule.hs`: `_gt` evaluation, dispatch in `meetCondition'`, and the shared `numToInt` helper lifted to module scope (reused by both `_eq` and `_gt`).
  - `src/CST.hs` / `src/Render.hs`: render `gt`/`not(gt)` as `>` / `\not>` for `phino explain`.
- Add the `amiss` normalization rule (`resources/normalize/amiss.yaml`) matching `⟦𝐵⟧(α𝑖 ↦ 𝑒)` → `⊥`, guarded by `not(gt(domain: 𝐵, 𝑖))`.

The guard is `𝑖 ≥ domain(𝐵)`, not a strict `𝑖 > domain(𝐵)`: `α1` on a single-void formation legitimately binds the auto-added `ρ` at index 1, so the boundary `𝑖 = domain(𝐵)` is itself out of range.

## Verification

```console
$ printf 'Φ ↦ ⟦ x ↦ ∅ ⟧(α2 ↦ ⟦⟧)\n' | phino rewrite --normalize --flat
Φ ↦ ⊥
[...]

Issue #871, opened 2026-06-25
LaTeX terminals \rho, \xi, \alpha_i rendered without \phiTerminal{} wrapper


yegor256:
When rendering to LaTeX (`--output=latex`), the phi-calculus terminal symbols come out bare. A program like `{Q.x.^.y}` produces `\rho`, the `$` terminal produces `\xi{}`, and `{Q.x(42)}` produces `\alpha_{0}` — none of them wrapped in the `\phiTerminal{}` macro. They should be emitted as `\phiTerminal{\rho}`, `\phiTerminal{\xi}`, and `\phiTerminal{\alpha_i}` respectively.

The bare strings originate in three places. `src/LaTeX.hs:246` sets `toLaTeX AT_RHO{} = AT_RHO RHO'`, which `src/Render.hs:70` renders as `\rho`. `src/LaTeX.hs:237` sets `toLaTeX EX_XI{} = EX_XI XI'`, which `src/Render.hs:79` renders as `\xi{}`. The alpha label is built at `src/LaTeX.hs:273` as `AT_LABEL ("\\alpha_{" <> subscript <> "}")`. Because the terminals are not wrapped, the generated LaTeX cannot be styled or redefined through the `\phiTerminal` macro the way the rest of the notation is.

A minimal fix: change `render RHO'` to `\phiTerminal{\rho}` and `render XI'` to `\phiTerminal{\xi}` in `src/Render.hs` (lines 70 and 79), and wrap the alpha label at `src/LaTeX.hs:273` as `AT_LABEL ("\\phiTerminal{\\alpha_{" <> subscript <> "}}")`. The `RHO'`/`XI'` constructors are produced only on the LaTeX path, so the change stays contained to LaTeX output.

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+783](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #877, opened 2026-06-25
bug: fix master CI breakage from merge ordering

yegor256:
Master CI is red ([run 28171915100](https://github.com/objectionary/phino/actions/runs/28171915100)) with 3 test failures. All three are merge-ordering breakages: #874 ("Reduce out-of-range positional application … to ⊥") landed alongside #875 (restore `~` as the ASCII alpha prefix) and #871 (wrap LaTeX terminals in `\phiTerminal{}`), but #874's new test fixtures were written against the pre-merge state.

## Failures and fixes

**1 & 2 — `Rule/check conditions/{gt-domain,not-gt}.yaml`** ("List of matched substitutions is empty which is not expected")

These packs wrote positional arguments as `a0` / `a3`. With the `~`/`α` alpha prefix restored, `a0` parses as a plain label attribute, so the pattern `[[ !B ]](α𝑖 ↦ Q)` no longer matched and the substitution list came back empty — before the `gt` condition was even evaluated. Switched to `α0` / `α3`, matching how the `e-ami` rewriter pack writes `α2` against the same pattern.

**3 — `CLI/explain/explains normalization rules`**

The expectation rendered the `amiss` rule's `α𝑖` as bare `\alpha_{i}`, but terminals are now wrapped, so the real output is `\phiTerminal{\alpha_{i}}`. Updated the expectation to match — consistent with the `alpha` and `dca` rules in the same expected block, which already use `\phiTerminal{\alpha_{i}}`.

## Verification

The Haskell toolchain isn't available in this environment, so I relied on CI to verify. The fixes match the failure output exactly and are internally consistent with fixtures that already pass (`e-ami` → `α2`; `alpha`/`dca` → `\phiTerminal{\alpha_{i}}`). No source/behavior changes — only test fixtures and one expected string.

---
_Generated by [Claude Code](https://claude.ai/code/session_01M9csBxJu36f5MY7xff2JW8)_

### DIFF
--- a/test-resources/condition-packs/gt-domain.yaml
+++ b/test-resources/condition-packs/gt-domain.yaml
@@ -1,7 +1,7 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
 # SPDX-License-Identifier: MIT
 ---
-expression: '[[ x -> $, w -> $ ]](a0 -> Q)'
+expression: '[[ x -> $, w -> $ ]](α0 -> Q)'
 pattern: '[[ !B ]](α𝑖 ↦ Q)'
 condition:
   gt:
--- a/test-resources/condition-packs/not-gt.yaml
+++ b/test-resources/condition-packs/not-gt.yaml
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MIT
 ---
 failure: true
-expression: '[[ x -> $, w -> $ ]](a3 -> Q)'
+expression: '[[ x -> $, w -> $ ]](α3 -> Q)'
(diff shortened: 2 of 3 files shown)


yegor256:
@maxonfjvipon good?

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+871](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #882, opened 2026-06-25
Make morphing 𝕄 binary with the global universe Q as immutable second argument

### LINKED ISSUE #879: Morphing 𝕄 is unary; the global universe Q should be its second argument, threaded through the rules
The morphing function 𝕄 is implemented as a unary relation: `morph :: Morphed -> DataizeContext -> IO Morphed` (src/Dataize.hs:74) takes a single expression and returns one. The global universe Q is never passed as an argument; instead the `root` rule (resources/morphing.yaml:65-78) fetches it implicitly through `function: global`, and `_global` (src/Dataize.hs:263-265) reads it from the mutable `DataizeContext._program` field. Formally 𝕄 should be binary — 𝕄(e, Q) — with Q the fixed global universe.

Because Q is read from the mutable `_program` instead of being threaded as a parameter, it can drift. `_dataize` (src/Dataize.hs:197-205), the synthetic dataization used inside atoms, replaces the program with `ctx{_program = prog}` (line 201), where `prog` prepends a fresh synthetic `BiTau` binding to the root formation. Any morphing performed under that context resolves Φ (the `root` rule's `global()`) against the augmented universe, not the true global Q; likewise `_morph` (src/Dataize.hs:271-276) recurses with whatever `_program` happens to be set. The global is implicit and silently mutable, which the binary definition would forbid.

Make 𝕄 binary: thread Q explicitly through `morph` (an added Q argument, or an immutable context field distinct from the mutable `_program`). The `root` rule then becomes 𝕄(Φ, Q) ⟿ 𝕄(𝒩(Q), Q), substituting the passed-in Q instead of calling `global()`; the `e ≠ Φ` guard and the `globe` ⊥-fallback stay. Every other rule (`prim`, `lambda`, `dispatch`, `application`, `applicationa`, `dead`, `xi`, `globe`) just forwards Q unchanged to its recursive 𝕄 calls, and the YAML descriptions plus the `explain --morph` rendering should show the Q argument.

yegor256:
Fixes #879.

## Problem

The morphing function 𝕄 was implemented as unary (`morph :: Morphed -> DataizeContext -> IO Morphed`). The global universe Q was never threaded as an argument; the `root` rule fetched it implicitly via `function: global`, and `_global` read it from the **mutable** `DataizeContext._program` field.

Because Q was read from `_program`, it could **drift**: `_dataize` (the synthetic dataization used inside atoms) replaces the program with `ctx{_program = prog}`, prepending a fresh synthetic `BiTau` binding to the root formation. Any morphing performed under that context resolved Φ (the `root` rule's `global()`) against the *augmented* universe, not the true global Q.

## Fix

Formally 𝕄 is binary — 𝕄(e, Q) — with Q the fixed global universe. This threads Q through an **immutable context field distinct from the mutable `_program`** (one of the two approaches suggested in the issue):

- Added `DataizeContext._universe :: Program`, set once when the context is built and never reassigned.
- `_global` (the `global()` build-term used by the `root` rule) now reads `_universe` instead of `_program`, so 𝕄(Φ, Q) always expands to the true global universe even when morphing runs under the augmented program installed by `_dataize`. The `e ≠ Φ` guard and the `globe` ⊥-fallback are unchanged.
- Every morphing rule forwards the same immutable Q; `_dataize` augments only `_program`, leaving Q untouched.

## Presentation

- `morphing.yaml` descriptions (and header), and the `dataization.yaml` `norm` description now show the binary form 𝕄(e, Q).
- `explain --morph` / `explain --dataize` rendering shows Q as the second argument: `\phinoMorph` becomes a binary macro `\mathbb{M}( #1, #2 )`, and the `morph` build-term in `where` clauses forwards Q.
- Updated `CLISpec` expectations, README examples, and the `latex.yml` mock macro.

https://claude.ai/code/session_01DYkvH9GLNHxb6WbGJovym4

---
_Generated by [Claude Code](https://claude.ai/code/session_01DYkvH9GLNHxb6WbGJovym4)_


yegor256:
@maxonfjvipon let's merge?

src/Dataize.hs
```diff
@@ -35,9 +35,15 @@ type Dataizable = (Expression, NonEmpty Rewritten)
...
 
+-- '_universe' is the fixed global universe, the second argument 'e' of the binary
+-- morphing function 𝕄(n, e) (Φ, rendered Q, is just its locator). Unlike
+-- '_program', which the synthetic '_dataize' and '_morph' helpers replace with
+-- augmented formations, '_universe' is set once when the context is built and
+-- never reassigned, so the universe cannot drift.
 data DataizeContext = DataizeContext
   { _locator :: Expression
   , _program :: Program
+  , _universe :: Program
```
maxonfjvipon: @yegor256 why do we need both `_program` and `_universe`? `_program` is changed during dataization, `_universe` is not and it isn't really correct because `_universe` may become invalid, we just don't have a proper tests for it

src/Rule.hs
```diff
@@ -343,19 +343,30 @@ kMetaNames :: Expression -> [T.Text]
...
 
-matchExpressionBy :: MatchExpressionFunc -> Expression -> Y.Rule -> RuleContext -> IO [Subst]
-matchExpressionBy matcher expr rule ctx =
+-- Like 'matchExpressionWithRule'' but seeds matching with the given
+-- substitutions, letting a caller inject pre-bound meta-variables. The morphing
+-- driver passes the global universe bound to 'e', the second argument of
+-- 𝕄(n, e), so the 'root' rule reads it directly instead of through a 'global()'
+-- build-term function.
+matchExpressionWithRuleSeeded' :: [Subst] -> Expression -> Y.Rule -> RuleContext -> IO [Subst]
+matchExpressionWithRuleSeeded' = matchExpressionBy matchExpression'
```
maxonfjvipon: @yegor256 instead of introducing new function, let's just change the signature, implementation and usages of `matchExpressionWithRule'` so it has [Subst] as first argument

resources/morphing.yaml
```diff
@@ -1,73 +1,73 @@
...
   then:
     morph:
-      normalize: '𝑒1(𝜏 ↦ 𝑒)'
+      normalize: '𝑒1(𝜏 ↦ 𝑒2)'
 
 - name: applicationa
-  description: '𝕄(n(α_i ↦ e)) ⟿ 𝕄(𝒩(𝕄(n)(α_i ↦ e)))'
-  match: '𝑛(α𝑖 ↦ 𝑒)'
+  description: '𝕄(n(α_i ↦ e_2), e) ⟿ 𝕄(𝒩(𝕄(n, e)(α_i ↦ e_2)), e)'
+  match: '𝑛(α𝑖 ↦ 𝑒2)'
```
maxonfjvipon: @yegor256 It's more native to use `e1` is match, and `e2` in where, so indexes go more natively. Let's do it everywhere

resources/dataization.yaml
```diff
@@ -50,7 +50,7 @@
   then: '--'
 
 - name: norm
-  description: '𝔻(n) ⟿ 𝔻(𝕄(n)) otherwise'
+  description: '𝔻(n) ⟿ 𝔻(𝕄(n, e)) otherwise'
```
maxonfjvipon: @yegor256 maybe we can introduce some special char for expression which is attached to Q? So we don't mess it up with regular expressions. WDYT?

maxonfjvipon (changes requested):

resources/dataization.yaml
yegor256: @maxonfjvipon it's just an expression. We can't use anything else, just `e`.

src/Dataize.hs
yegor256: `_universe` is the fixed global universe Q — the second argument of the binary 𝕄(n, e). `_program` is the working program, which `_dataize` augments with a synthetic τ-binding while dataizing inside atoms. They have to differ: if the `root` rule resolved Φ against the mutated `_program`, it would morph against that augmented universe instead of the true Q — which is exactly the drift this PR (#879) fixes. By the formal model Q is fixed for the whole dataization, so `_universe` is set once and never reassigned; that immutability is the point, not an accident.

Agreed the test gap is fair — I can add a focused test that augments `_program` (as `_dataize` does) and checks that morphing Φ still resolves to the original universe rather than the augmented one. Want me to add it here?

---

maxonfjvipon: @yegor256 Q can't be fixed because program is rewritten by normalization rules

src/Rule.hs
yegor256: Done in 8c5d6b7. Dropped `matchExpressionWithRuleSeeded'` and gave `matchExpressionWithRule'` a leading `[Subst]` seed argument instead. Updated its two usages: `dataize'` passes `[substEmpty]` (no seed) and `morph` passes the universe seed.

---

resources/morphing.yaml
yegor256: Done in 8c5d6b7. `e1` is now the match-bound expression and `e2` the where-computed one, applied consistently across `lambda`, `dispatch`, `application` and `applicationa` (and reflected in the `explain --morph` test). `n` stays un-indexed since there is only one per rule.

---

src/Dataize.hs
yegor256: Resolved in f880e55, going further than keeping the field: per @yegor256's direction, nothing global is fixed in `DataizeContext` anymore. Both `_program` and `_universe` are gone. 𝔻 and 𝕄 now receive the working program and the universe Q as explicit arguments, threaded on to the atoms. `_dataize` augments only the working program and forwards Q unchanged, so morphing Φ still resolves to the true universe (the drift fix), now expressed purely through arguments rather than a stored field. So the "may become invalid" concern no longer applies — there's nothing stored to go stale.

---

resources/morphing.yaml
```diff
@@ -1,73 +1,73 @@
...
-      normalize: '𝑒.𝜏'
+      normalize: '𝑒2.𝜏'
 
 - name: dispatch
-  description: '𝕄(n.τ) ⟿ 𝕄(𝒩(𝕄(n).τ))'
+  description: '𝕄(n.τ, e) ⟿ 𝕄(𝒩(𝕄(n, e).τ), e)'
   match: '𝑛.𝜏'
   where:
-    - meta: 𝑒
+    - meta: 𝑒2
```
maxonfjvipon: @yegor256 why e2, e1 can be used

yegor256:
@maxonfjvipon take a look again. we want to get rid of the global state (Q). it should be passed as argument to D() and M().

src/Dataize.hs
```diff
@@ -65,23 +69,32 @@ formation bds ctx = do
...
+-- 𝕄(n, e): besides the term 'n' it takes the fixed global universe 'e' ('univ')
+-- as an argument and seeds it into the substitution as the 'e' meta that the
+-- 'root' rule substitutes. The working program 'prog' is threaded for the
+-- normalization context. It is driven by the ordered rules from 'morphing.yaml':
+-- the first matching rule's 'then' outcome either stops at a formation
+-- ('MoStop') or keeps morphing ('MoMorph'), always forwarding the same universe.
+-- When the morphed argument is a normalization ('MaNormalize', as in the
+-- 'lambda' and 'root' rules), the rewriter runs over the rule's product and its
+-- individual steps are spliced into the chain before morphing continues.
+morph :: Morphed -> Program -> Expression -> DataizeContext -> IO Morphed
```
maxonfjvipon: @yegor256 why do we need `Program` + `Expression`? Why can't we use single `Program` which is universe and pass it everywhere?

maxonfjvipon (changes requested):

src/Dataize.hs
yegor256: Done in 6c0ebf5. There's no longer a `Program` threaded anywhere — 𝕄(n, e) and 𝔻(n, e) now take two **expressions**: the term and the universe `e`. The working program needed for normalization is taken from the head of the step chain (it gets re-embedded at the locator anyway, so it's equivalent), and `_dataize` augments the universe expression to build its local program while forwarding the universe unchanged. `DataizeContext` keeps only configuration.

---

resources/morphing.yaml
yegor256: Done in 6b426ca. `lambda` and `dispatch` now use `e1` for their where-computed meta (there's no match-bound expression in those rules, so `e1` is free). `e2` is used only where `e1` is already taken by a match-bound expression — i.e. `application`/`applicationa` (and `box` in dataization.yaml).

---

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (4) comments.  Your running score is [+351](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (398 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+953](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #886, opened 2026-06-26
Write morphing and dataization rules as inference rules

### LINKED ISSUE #885: Suboptimal YAML Formatting of Dataization and Morphing Rules in `morphing.yaml`
Instead of the current format of the `application` rule in the `morphing.yaml` file, I suggest the following: 

```
- name: application
  label: 'app'
  match: '𝑛(𝜏 ↦ 𝑒)'
  n-result: '𝑛3'
  e-result: 'e'
  premises:
  - n-result: '𝑛1'
    e-result: 'e'
    morph: '𝑛'
  - n-result:  '𝑛2'
    normalize: '𝑛1(𝜏 ↦ 𝑒)'
  - n-result:  '𝑛3'
    e-result: 'e'
    morph: '𝑛2'
```

This should give us this formula in LaTeX:

```
\begin{phinoInference}
  \phinoName{application}
  \phinoLabel{app}
  \phinoPremise{\phinoMorph{n}{n1}}
  \phinoPremise{\phinoNormalize{n1(\tau -> e)}{n2}}
  \phinoPremise{\phinoMorph{n2}{n3}}
  \phinoConclusion{\phinoMorph{n(\tau -> e)}{n3}}
\end{phinoInference}
```

Similar way, we should restructure all dataization and morphing rules.

yegor256:
Each rule in `morphing.yaml` and `dataization.yaml` is now written as an inference rule: a `match`, an `e-result`/`n-result`/`d-result` conclusion, an optional `when` guard, and an ordered list of `premises` (each a `morph`/`normalize`/`lambda`/`contextualize`/`dataize` judgment). `explain --morph` and `explain --dataize` build `\phinoInference` blocks from that structure instead of the old flat `\phinoMorphingRule` rows.

The previous `match`/`where`/`then` shape was hard to read and did not match how the rules are presented on paper. The engine still needs the operational form, so it reconstructs the `where`/`then` steps from the premises internally — dataization values and the per-step reduction traces are byte-for-byte unchanged, only the YAML and the LaTeX output differ.

Two small departures from the sketch in the issue: the application argument keeps the meta name `e1` (a bare `e` would collide with the seeded universe meta), and `\phinoMorph`/`\phinoDataize` are ternary with the universe as the middle argument so the judgments carry it explicitly.

Closes #885.


yegor256:
@maxonfjvipon take a look

resources/dataization.yaml
```diff
@@ -5,55 +5,72 @@
...
+  label: box
   match: ⟦𝐵1, φ ↦ 𝑒1, 𝐵2⟧
-  where:
-    - meta: 𝑒2
-      function: contextualize
-      args:
-        - 𝑒1
-        - ⟦𝐵1, φ ↦ 𝑒1, 𝐵2⟧
+  e-match: 𝑒
+  d-result: δ1
```
maxonfjvipon: @yegor256 why `δ1`, why not `δ`

resources/dataization.yaml
```diff
@@ -5,55 +5,72 @@
...
-      function: lambda
-      args:
-        - ⟦𝐵1, λ ⤍ 𝐹, 𝐵2⟧
-  then:
-    dataize:
-      normalize: 𝑒1
+  e-match: 𝑒
+  d-result: δ1
+  premises:
+    - n-result: 𝑛1
```
maxonfjvipon: @yegor256 why `n1`, why not `n`

src/Yaml.hs
```diff
@@ -331,13 +365,71 @@ instance FromJSON DataizeRule where
...
-            <*> o .: "then"
+            <*> o .:? "premises" .!= []
       )
 
+-- Recover the operational ('where', 'then') form of a morphing rule from its
+-- inference-rule premises and conclusion. The conclusion meta is produced by a
+-- trailing 'morph' premise; a 'morph(normalize(_))' continuation folds the
+-- normalize and morph premises into 'then', leaving the rest as 'where'
+-- side-computations. A literal conclusion is a terminal result.
+morphReduction :: MorphRule -> (Maybe [Extra], MorphOutcome)
```
maxonfjvipon: @yegor256 what's the actual point of having these morph and dataize reductions?

maxonfjvipon (changes requested):

resources/dataization.yaml
yegor256: Good point — there is only one bytes meta in this rule, so the subscript bought nothing. Renamed `δ1` → `δ` here, and likewise in `fire` and `norm`. Fixed in 8fe61ab.

resources/dataization.yaml
yegor256: Agreed — `fire` does not bind `n` in its `match`, so the first derived term can just be `n`. The chain now reads `n` (𝔼) → `n1` (𝒩) → `δ` (𝔻). I kept the numbered `n1…n3` only in the rules whose `match` already binds the bare `n` (`dispatch`, `application`, `applicationa`, `norm`). Fixed in 8fe61ab.

src/Yaml.hs
yegor256: These bridge the two views of a rule. The YAML is now declarative — premises plus a conclusion — but the morphing/dataization engine in `Dataize.hs` still runs an operational shape: match the term, build the `where` metas, then take one `then` reduction step. `morphReduction`/`dataizeReduction` translate the premises and conclusion into that `(where, then)` pair, so the existing engine drives the new format unchanged and the step traces stay byte-identical (the side-computation premises stay silent, only the continuation splices its sub-steps; that is why the trace tests still pass untouched).

The alternative is to make the driver consume premises directly, but that just moves the same where/then split into `Dataize` and risks changing the `--steps` output. Happy to go that way instead if you would prefer the driver to evaluate premises natively — let me know.

maxonfjvipon: @yegor256 it would be better to adapt the engine so it works natively with this new premises approach. We don't need this morph outcome since it does not used and parsed. What will it require? We still will need some sort of a bridge but we don't need morph outcome in that case. Premises made rules simpler so it seems that building new substituions from premises should be easier now. WDYT?

maxonfjvipon (changes requested):

src/Yaml.hs
yegor256: Done in 2bf4066. The engine now consumes the premises directly: `morph`/`dataize'` match the conclusion term, check the guard, fold the side-computation premises into the substitution, then run the spine that builds the conclusion straight off the premise list. `MorphOutcome`/`DataizeOutcome`/`MorphArg`/`DataizeArg` and `morphReduction`/`dataizeReduction` are gone — `Yaml.hs` is now just the data model plus parsers (−92 lines there).

The only bridge left is `asRule`, which turns a rule into a bare `Y.Rule` (just `match` + `when`) for matching — no more `where`/`having` smuggling. Side-computations still run through the same isolated `morph`/`lambda`/`contextualize` build-term functions, so step traces and dataization values are byte-identical; the whole suite (1065 examples) passes untouched. What moved is *where* the premises are evaluated — out of the matcher and into the driver, as you suggested.

yegor256:
@maxonfjvipon I reworked the engine to consume the premises natively, as you suggested (`2bf4066`).

`morph` and `dataize'` now match the conclusion term, check the guard, fold the side-computation premises into the substitution, and then run the spine that produces the conclusion — directly off the premise list. The `MorphOutcome`/`DataizeOutcome`/`MorphArg`/`DataizeArg` types and the `morphReduction`/`dataizeReduction` reductions are gone, so `Yaml.hs` is now just the data model plus parsers (−92 lines there).

The one bridge that remains is `asRule`, which builds a bare `Y.Rule` from just `match` + `when` so the existing matcher can still find and guard a rule — no more `where`/`having` smuggling. Side-computations (`morph`/`lambda`/`contextualize` of an earlier term) keep running through the same isolated build-term functions they always did, so step traces and dataization values come out byte-identical; the whole suite (1065 examples) passes untouched and CI is green.

Mind taking another look?

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the review! You've earned +15 points for this: +12 as a basis; +8 for the 838 hits-of-code that you reviewed; -5 for very few (3) comments.  Your running score is [+409](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (838 >= 100); -8 for way too many hits-of-code (838 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+866](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #894, opened 2026-06-27
fix(#891): emit set-grouping parentheses as \lparen/\rparen

### LINKED ISSUE #891: `explain` emits set-grouping parentheses as literal `(` `)` instead of `\lparen`/`\rparen`
Running `phino explain --dataize` prints the `box` rule's disjoint `when` condition as `{ [ \Delta, L ] \cap ( B_1 \cup B_2 ) = \emptyset }`, where the parentheses grouping the `\cup` set are emitted as the literal characters `(` and `)`. The offending code is `render CO_DISJOINT` in `src/Render.hs:269-274`, which hardcodes `" ] \\cap ( "` and `" ) = \\emptyset"`.

This is wrong because parentheses are not part of phi syntax: in the typeset condition they are LaTeX grouping delimiters and should be emitted as the macros `\lparen` and `\rparen`, the same way every other phi token in the condition is rendered through a macro rather than a raw character. The same literal `(`/`)` also appears in the sibling cases of the `Render CONDITION` instance — the nested `CO_LOGIC` wrap (`src/Render.hs:259`), `matches`/`part-of` (`src/Render.hs:266-267`) and `renderFunc` (`src/Render.hs:278`). This is independent of #890, which concerns the `\Delta`→`D` glyph in the same row.

The fix is to replace the literal `( ` and ` )` with `\lparen ` and ` \rparen` in `render CO_DISJOINT` (`src/Render.hs:269-274`), and the matching literals in the other cases of that instance.

yegor256:
Fixes #891.

## Problem

`phino explain --dataize` printed the `box` rule's disjoint `when` condition as `{ [ \Delta, L ] \cap ( B_1 \cup B_2 ) = \emptyset }`, where the grouping parentheses were emitted as the literal characters `(` and `)`. Parentheses are not part of phi syntax — in the typeset condition they are LaTeX grouping delimiters and should be emitted through macros like every other phi token.

## Fix

In `src/Render.hs`, replaced the literal `(`/`)` with the `\lparen`/`\rparen` macros in every case of the `Render CONDITION` instance (and its helper) that emitted them:

- `render CO_DISJOINT` (the case named in the issue)
- the nested `CO_LOGIC` wrap (`renderWrapped`)
- `render CO_MATCHES` and `render CO_PART_OF`
- `renderFunc`

Updated the `explain --dataize` expectation in `test/CLISpec.hs` to match the new output.

https://claude.ai/code/session_01SDyHs9upTBmbrhmVkKyWQn

---
_Generated by [Claude Code](https://claude.ai/code/session_01SDyHs9upTBmbrhmVkKyWQn)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+950](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #903, opened 2026-06-28
fix(#901): shuffle morphing rules before first-match walk

### LINKED ISSUE #901: Morphing walks rules in fixed YAML order; shuffle to verify order-independence (src/Dataize.hs:87)
In `src/Dataize.hs:87` the `morph` function selects a morphing rule via `firstMatch Y.morphingRules`, walking the rules in their fixed `resources/morphing.yaml` declaration order and taking the first whose `match`/`e-match` succeeds (`src/Dataize.hs:92-98`). Together with the recursive 𝕄 function this is the only path morphing takes to a formation, so the rule order is load-bearing by convention.

Issues #856 and #860 already established that we want every morphing rule to be order-independent, and they removed the known overlaps (`lambda`/`dispatch`, the `stuck` catch-all) so that order no longer changes behavior. Nothing, however, exercises that invariant: while `firstMatch` always walks the YAML order, a newly-introduced overlap would silently re-establish a hidden order dependence and the suite would stay green. Normalization avoids this fragility by applying its rules repeatedly until none match — the fixpoint loop in `rewrite` (`src/Rewriter.hs:215-239`) — instead of relying on a fixed first-match position.

The fix is to shuffle `Y.morphingRules` (seeded, for reproducibility) before the `firstMatch` walk in `morph` (`src/Dataize.hs:87`), so each morphing step chooses among the matching rules in a random order, mirroring normalization's "apply until they stop matching" approach. If morphing is genuinely order-independent the suite stays deterministic; if any hidden overlap remains it now surfaces as a nondeterministic failure.

yegor256:
Closes #901.

## Problem

`morph` (`src/Dataize.hs`) selected a morphing rule via `firstMatch Y.morphingRules`, walking the rules in their fixed `resources/morphing.yaml` declaration order and taking the first whose `match`/`e-match` succeeded. Together with the recursive 𝕄 function, this is the only path morphing takes to a formation, so the rule order was load-bearing by convention.

Issues #856 and #860 already made every morphing rule order-independent — they removed the known overlaps (`lambda`/`dispatch`, the `stuck` catch-all) so order no longer changes behavior. But nothing exercised that invariant: while `firstMatch` always walked the YAML order, a newly-introduced overlap would silently re-establish a hidden order dependence and the suite would stay green.

## Change

Shuffle `Y.morphingRules` before the `firstMatch` walk in `morph`, so each morphing step chooses among the matching rules in a random order — mirroring normalization's "apply until they stop matching" fixpoint approach (`Rewriter.rewrite`). This reuses the existing `Misc.shuffle` (seeded Fisher-Yates) already in the module's imports.

- If morphing is genuinely order-independent (one rule matches at any term), the result and step trace are unchanged and the suite stays deterministic.
- If any hidden overlap remains, it now surfaces as a nondeterministic failure instead of a silently order-dependent green suite.

The morphing rules are mutually exclusive by construction (`root` vs `globe` via `e-match`/`when`, `lambda` vs `dispatch` via the `not (formation 𝑛)` guard), so shuffling does not change observable behavior under correct conditions.

Doc comments on `morph` and the `morphing.yaml` header were updated to describe the shuffle-then-first-match behavior.

## Notes

No Haskell toolchain is available in this environment, so `make test` / `make hlint` / `make fourmolu` were not run locally — CI will exercise them.

---
_Generated by [Claude Code](https://claude.ai/code/session_01WUFPnnWLoRQC3GRKWS8gwd)_


yegor256:
@maxonfjvipon WDYT?

src/Dataize.hs
```diff
@@ -74,17 +74,23 @@ formation bds univ ctx = do
...
+-- is built, always forwarding the same universe. The 'morph' premise that
+-- produces the conclusion is the spine: when
 -- its argument comes from a 'normalize' premise, the rewriter runs over that
 -- argument and its individual steps (alpha, copy, dot, …) are spliced into the
 -- chain before morphing continues. Every other premise is a side-computation
 -- evaluated in isolation by 'sidePremise', its own steps discarded.
 morph :: Morphed -> Expression -> DataizeContext -> IO Morphed
 morph (expr, seq) univ ctx = do
-  matched <- firstMatch Y.morphingRules
+  rules <- shuffle Y.morphingRules
```
maxonfjvipon: @yegor256 let's make it optional with `--shuffle` flag

maxonfjvipon (changes requested):

src/Dataize.hs
yegor256: Done. Shuffling is now opt-in behind a `--shuffle` flag on the `dataize` command (mirroring the existing flag on `rewrite`/`explain`), threaded into `DataizeContext` as `_shuffle`. By default `morph` walks the rules in declaration order; it shuffles only when the flag is set. I also added a `DataizeSpec` test that morphs several inputs 100 times under a shuffling context and asserts the outcome is always the single expected form, proving the rules may be applied in any order with the same result.

---

yegor256:
@maxonfjvipon check again please

yegor256:
@maxonfjvipon the branch is clean, take a look

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

yegor256:
@rultor status

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1172](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+453](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #904, opened 2026-06-28
fix(#902): guard the `none` dataization rule with a disjoint condition

### LINKED ISSUE #902: `none` rule in dataization.yaml matches any formation without a guard, defeating order-independence (resources/dataization.yaml:57)
The `none` rule in `resources/dataization.yaml:57` matches any formation `⟦𝐵⟧` and yields empty bytes `--`, with no `when` guard. It is correct only because 𝔻 walks the rules top-to-bottom and stops at the first match (`firstMatch Y.dataizationRules`, `src/Dataize.hs:153`), so `none` happens to sit after `delta` (line 19), `box` (line 25), and `fire` (line 44).

This makes the rule's correctness depend entirely on its position in the file. `delta`, `box`, and `fire` each match a more specific formation — one carrying a Δ, φ, or λ binding — yet `none` matches every formation, those included. If the dataization rules were reordered or shuffled (the order-independence goal already pursued for morphing in #901), `none` would shadow the three specific rules and wrongly return `--` for a formation that should dataize through its Δ, φ, or λ. The `box` rule already avoids this trap: it guards itself with `when: disjoint: [[Δ, λ], [𝐵1, 𝐵2]]` so its match stays unambiguous regardless of order.

Add the same kind of guard to `none`: `when: disjoint: [[Δ, λ, φ], [𝐵]]`, asserting that the matched bindings 𝐵 carry none of Δ, λ, or φ. The `disjoint` condition (`src/Rule.hs:233`) already supports exactly this — it holds when none of the listed attributes is present in the union of the named binding metas. With the guard in place `none` becomes self-describing and the dataization rules can be applied in any order.

yegor256:
Fixes #902.

## Problem

The `none` rule in `resources/dataization.yaml` matched any formation `⟦𝐵⟧` and yielded empty bytes `--` with no `when` guard. It was correct only because 𝔻 walks rules top-to-bottom and stops at the first match (`firstMatch Y.dataizationRules`, `src/Dataize.hs:153`), so `none` happened to sit after `delta`, `box`, and `fire`. Reordering the rules would let `none` shadow those three specific clauses and wrongly return `--` for a formation that should dataize through its Δ, φ, or λ.

## Change

Added the same kind of guard `box` already uses:

```yaml
when:
  disjoint:
    - [Δ, λ, φ]
    - [𝐵]
```

This asserts the matched bindings `𝐵` carry none of Δ, λ, or φ. The `disjoint` condition (`src/Rule.hs:233`) holds when none of the listed attributes is present in the union of the named binding metas. With the guard in place, `none` is self-describing and the dataization rules can be applied in any order. The empty-formation case (`⟦⟧ ⇒ --`) still matches, since an empty binding set carries none of Δ, λ, φ.

https://claude.ai/code/session_01R9gh4HKJb1EkBzLNCeajb2

---
_Generated by [Claude Code](https://claude.ai/code/session_01R9gh4HKJb1EkBzLNCeajb2)_

### DIFF
--- a/resources/dataization.yaml
+++ b/resources/dataization.yaml
@@ -59,6 +59,10 @@
   match: ⟦𝐵⟧
   e-match: 𝑒
   d-result: '--'
+  when:
+    disjoint:
+      - [Δ, λ, φ]
+      - [𝐵]
 
 - name: bott
   label: bott
--- a/test/CLISpec.hs
+++ b/test/CLISpec.hs
@@ -1138,6 +1138,7 @@ spec = do
             , "\\begin{phinoInference}"
             , "  \\phinoName{none}"
             , "  \\phinoLabel{none}"
+            , "  \\phinoCondition{ [ D, L, @ ] \\cap \\lparen B \\rparen = \\emptyset }"
             , "  \\phinoConclusion{ \\phinoDataize{ [[ B ]] }{ e }{ -- } }"
             , "\\end{phinoInference}"
             , "\\begin{phinoInference}"


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (5) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1030](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #906, opened 2026-06-28
fix(#905): guard the `norm` dataization rule so clause order is not load-bearing

### LINKED ISSUE #905: `norm` dataization rule matches any term, making rule order load-bearing (resources/dataization.yaml:73)
The `norm` dataization rule (`resources/dataization.yaml:73-82`) matches the bare meta `𝑛`, which unifies with any expression. Rule selection in `dataize'` walks the rules in their fixed YAML order and takes the first match (`firstMatch Y.dataizationRules`, `src/Dataize.hs:153`), so `norm` only behaves correctly because it is declared last and catches whatever the earlier clauses left over.

This makes the declaration order load-bearing. `norm` overlaps every preceding clause: `delta`, `box`, `fire` and `none` (all formations) and `bott` (`⊥`). Move it above any of them and it would morph-then-dataize a formation or a termination that the specific rule is supposed to handle, silently changing behavior. Morphing already removed exactly this kind of overlap for `lambda`/`dispatch` so their order stops mattering (see the header comment in `resources/morphing.yaml:24-27`), and #901 tracks the same order-independence invariant on the morphing side. The dataization `norm` rule is the remaining hole.

The fix is to add a `when` guard so `norm` fires only when `𝑛` is neither a formation nor a termination — reuse the existing `formation` condition function that `dispatch` already uses (`resources/morphing.yaml:53-55`, implemented as `_isFormation` in `src/Rule.hs:200`), combined with a check that `𝑛` is not `⊥`/`T`. With that guard the dataization clauses become disjoint and apply in any order.

yegor256:
Closes #905.

## Problem

The `norm` dataization rule (`resources/dataization.yaml`) matches the bare meta `𝑛`, which unifies with any expression. Rule selection in `dataize'` (`src/Dataize.hs`) walks the rules in fixed YAML order and takes the first match, so `norm` only behaved correctly because it is declared last and catches whatever the earlier clauses left over. This made the declaration order load-bearing: `norm` overlaps every preceding clause — `delta`, `box`, `fire`, `none` (all formations) and `bott` (`⊥`).

## Fix

Add a `when` guard so `norm` fires only when `𝑛` is neither a formation nor the termination `⊥`:

```yaml
when:
  and:
    - not:
        formation: 𝑛
    - not:
        eq:
          - 𝑛
          - ⊥
```

- `not (formation 𝑛)` reuses the existing `_isFormation` condition (`src/Rule.hs`), carving out `delta`, `box`, `fire` and `none`.
- `not (𝑛 = ⊥)` carves out `bott`. Since `T` and `⊥` both parse to `ExTermination`, the single check covers both.

With this guard the dataization clauses are disjoint and apply in any order — mirroring what morphing already did for `lambda`/`dispatch`. The header comment in `resources/dataization.yaml` is updated to document the invariant.

## Tests

Added a `dataization 'norm' is disjoint from the specific clauses` block in `test/DataizeSpec.hs` (mirroring the existing `dispatch`/`lambda` disjointness test) asserting `norm` does not fire on a formation or on `⊥`, but still fires on a non-formation, non-termination normal form.

https://claude.ai/code/session_01XoLXw5ve6cwjPBYymS2NXV

---
_Generated by [Claude Code](https://claude.ai/code/session_01XoLXw5ve6cwjPBYymS2NXV)_

### DIFF
--- a/resources/dataization.yaml
+++ b/resources/dataization.yaml
@@ -15,6 +15,14 @@
 # 𝔻(n, e); it is always the e meta. The single bytes result is named δ; derived
 # terms are 𝑛, 𝑛1, … in premise order, skipping the bare 𝑛 only when 'match'
 # already binds it.
+#
+# The clauses are disjoint, so their relative order does not change behavior.
+# 'norm' matches the bare meta 𝑛, which unifies with any expression, so it is
+# guarded to fire only when 𝑛 is neither a formation ('not (formation 𝑛)',
+# carving out 'delta', 'box', 'fire' and 'none') nor the termination ⊥
+# ('not (𝑛 = ⊥)', carving out 'bott'). Without the guard 'norm' would behave
+# correctly only by being declared last; with it the clauses no longer rely on
+# their order.
 
 - name: delta
   label: \Delta
@@ -75,6 +83,14 @@
   match: 𝑛
   e-match: 𝑒
   d-result: δ
+  when:
+    and:
+      - not:
+          formation: 𝑛
+      - not:
+          eq:
+            - 𝑛
+            - ⊥
   premises:
     - n-result: 𝑛1
       morph: 𝑛
--- a/test/CLISpec.hs
+++ b/test/CLISpec.hs
@@ -1149,6 +1149,7 @@ spec = do
             , "\\begin{phinoInference}"
             , "  \\phinoName{norm}"
             , "  \\phinoLabel{norm}"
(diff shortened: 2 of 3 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (38) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1034](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #915, opened 2026-06-28
feat(#912): migrate normalization rules to the premise-conclusion schema

### LINKED ISSUE #912: Migrate normalization rules to the premise-conclusion schema used by 𝕄/𝒞/𝔻
We now carry two rule schemas. Normalization rules in `resources/normalize/*.yaml` use `name`/`pattern`/`result` with optional `when`/`where`/`having` (parsed into `Yaml.Rule`, src/Yaml.hs:201), while morphing, contextualization and dataization use the inference-rule form `match` + `e-match`/`c-match` + `n-result`/`c-result`/`d-result` + `premises` (`Yaml.MorphRule`/`ContextualizeRule`/`DataizeRule`, src/Yaml.hs:226-395). The split also forks the LaTeX backend: normalize rules render through `explainRule`/`trrule` (src/LaTeX.hs:356-453), the other three through the shared `inference`/`premiseToLatex` helpers (src/LaTeX.hs:374-427).

Migrating normalization to the premise-conclusion schema would give one uniform rule format across all five reduction systems and one uniform LaTeX inference-rule output, letting `explainRule`/`trrule` collapse into `inference`. The mapping is clean: conditionless rules (`copy`, `dc`, `dd`, `dca`, `stay`, `null`) become premise-less axioms; `when`/`having` become the side condition above the line; and `dot.yaml`'s `where` (a `contextualize` build-term call) becomes a genuine premise, 𝒞(𝑛, formation) ⟿ 𝑒.

One caveat to settle before coding: normalization is a rewriting relation — the engine searches for any redex anywhere and iterates to a fixpoint (`isNF`, `matchesAnyNormalizationRule` in src/Rule.hs:42-73) — whereas 𝕄/𝒞/𝔻 are structural-recursive judgments. So this migration touches the rule schema and LaTeX only, not the rewriting engine, and the issue should decide whether to unify the conclusion's relation symbol (a one-step rewrite ↦) with the recursive ⟿ of 𝕄/𝒞/𝔻 or keep normalization's relation distinct while sharing the inference-rule layout. Concrete fix scope: add the new normalize schema and parser in src/Yaml.hs, update consumers (src/Rule.hs, src/Rewriter.hs), and fold the two LaTeX paths into `inference`.

Maybe we can even merge all rules into `normalize.yaml` single file.

yegor256:
Closes #912.

Normalization rules now use the same premise-conclusion schema as 𝕄/𝒞/𝔻, and `explain` emits one uniform inference-rule LaTeX layout across all five reduction systems — `explainRule`/`trrule` collapse into the shared `inference` helper.

## What changed

- **`src/Yaml.hs`** — `FromJSON Rule` is now a custom parser that accepts the premise-conclusion keys (`match` + `premises`) **as well as** the legacy keys (`pattern` + `where`). A `premises` list desugars (`premiseToExtra`) into the build-term `where` extras the engine already runs, so the rewriting engine itself is untouched. `verb`/`verbArgs` moved here from `Dataize` so both the desugaring and the dataizer share them.
- **`src/LaTeX.hs`** — `explainRule` folds into `inference`: `when`/`having` become the side condition above the line, `where` extensions become premise judgments (so `dot`'s `contextualize` shows as 𝒞(n, formation) ⟿ e), and the conclusion is the one-step rewrite 𝒩(match) ⟿ result. `trrule` and `extraArgumentsToLatex` are removed.
- **`resources/normalize/*.yaml`** — migrated to `match`; `dot.yaml`'s `where` contextualize call is now a genuine `premises` entry.
- **`README.md`** — the `explain --normalize` example now shows the inference-rule output (consistent with the already-updated `--contextualize` example).
- **`test/CLISpec.hs`** — `explain` expectations updated to the `\begin{phinoInference}` format.

## Design decisions (the issue left two open)

These were the two forks the issue raised; I picked defensible defaults and call them out here so they can be revisited in review:

1. **Scope — additive, not a hard rename.** The `Yaml.Rule` schema is *also* the general rewriting-rule schema used by the user-facing `--rule` feature and ~50+ test packs (`condition-packs/`, `rewriter-packs/`, `cli/`). A hard rename would break all of those. Instead the parser accepts both schemas, the built-in `resources/normalize/*.yaml` files are migrated to the new form, and nothing else breaks. Because `premises` desugar to the existing `where` extras, `src/Rule.hs` and `src/Rewriter.hs` needed no changes — the engine keeps consuming `where` extras unchanged. (Note: `where`/`Extra` is strictly more general than `premises` — it supports arbitrary build-term functions like `sed`/`concat` — which is why it stays the engine mechanism rather than being replaced.)
2. **Relation symbol — shared 𝒩 macro, no new symbol.** The conclusion renders through `\phinoNormalize{lhs}{rhs}`, the same macro the 𝒩 premises of the other systems already use, so the inference layout is shared without inventing a separate one-step relation symbol. This keeps the relation in the 𝒩 family and needs no new `.sty` macro. Happy to switch to a distinct ↦ macro or unify with ⟿ if preferred.

## Verification

- `cabal test --ghc-options=-Werror` → **1073 examples, 0 failures**
- `fourmolu --mode check` and `hlint` clean on all changed files
- Manually ran `phino explain --normalize` (inference output incl. the `dot` premise) and a `rewrite --normalize` (the `dot` rule fires correctly end-to-end)

https://claude.ai/code/session_01NZ7UYvDx9A7XhFyc4NSPy9

---
_Generated by [Claude Code](https://claude.ai/code/session_01NZ7UYvDx9A7XhFyc4NSPy9)_


yegor256:
@maxonfjvipon good?

src/LaTeX.hs
```diff
@@ -371,6 +376,29 @@ explainRule rule =
...
     joinedConditions (Just first) (Just second) = Just (Y.And [first, second])
 
+-- Render a 'where' extension as a premise judgment. The five reduction verbs
+-- render through their inference-rule macros; any other build-term function
+-- (a user rule's 'sed', 'concat', …) keeps the 'meta ≔ func(args)' assignment
+-- form, since it has no inference-rule reading.
+extraToLatex :: Y.Extra -> String
+extraToLatex extra = fromMaybe assignment judgment
+  where
+    assignment = renderToLatex (extraToCST extra) defaultLatexContext
```
maxonfjvipon, line 388: @yegor256 Let's add function types to all of the functions

maxonfjvipon (changes requested):
@yegor256 As I understand we still have some sort of a bridge? Are we planning to get rid of it? If so let's add pdd todo

src/LaTeX.hs
yegor256: Added type signatures to the `extraToLatex` helpers. Also went further per @yegor256's note: dropped the legacy `pattern`/`where` schema entirely — the parser now accepts only `match`/`premises`, and all rule YAML (built-in normalization rules plus the test/`--rule` packs) is migrated. A premise is either a reduction judgment (verb-keyed) or a general build-term call (`meta`/`function`/`args`), so `sed`/`concat`/etc. still work and the rewriting engine is unchanged.

---

yegor256:
@maxonfjvipon here's a summary of what this PR now does, after folding in your review feedback.

### Goal (#912)
Migrate normalization rules to the same premise-conclusion schema used by 𝕄/𝒞/𝔻, and collapse the forked LaTeX backend so all five reduction systems render through one inference-rule layout.

### What changed
- **One rule schema.** `Yaml.Rule` now uses `match` + `premises` (was `pattern` + `where`). The legacy keys are **gone** — the parser accepts only the new schema. Per @yegor256's note, this removes the old machinery entirely rather than keeping a dual schema.
- **Premises subsume the old `where`.** A premise is either:
  - a reduction judgment, keyed by its verb (`contextualize` / `morph` / `normalize` / `evaluate` / `dataize`), or
  - a general build-term call (`meta` / `function` / `args`).

  Both desugar to the same `Extra` the rewriting engine already consumes, so `sed` / `concat` / `size` / `random-tau` / etc. keep working and **the rewriting engine is untouched** — only the schema and the LaTeX backend changed.
- **Unified LaTeX.** `explainRule`/`trrule` are gone; normalization rules render through the shared `inference` helper, emitting `\begin{phinoNormalizationInference}` consistent with the `phino{Morphing,Dataization,Contextualization}Inference` environments from #918. `when`/`having` become the side condition, premises become judgments (so `dot`'s `contextualize` shows as 𝒞(n, formation) ⟿ e), and the conclusion is 𝒩(match) ⟿ result.
- **All `Y.Rule` YAML migrated:** built-in `resources/normalize/*`, plus `test-resources/{yaml-packs,yaml-typos,cli}/*` and the inline `custom:` rules in `rewriter-packs/custom/*`. `condition-packs` use a different schema (`ConditionPack`) and were intentionally left alone.
- **README** rule example + schema docs updated to `match`/`premises`.
- **Review fix:** added type signatures to the `extraToLatex` helpers.

### One open call worth your eyes
The normalization conclusion reuses the existing `\phinoNormalize{lhs}{rhs}` macro (the same one the 𝒩 premises of the other systems use) rather than introducing a distinct one-step `↦` relation symbol — sharing the layout without a new macro. Easy to switch to a separate symbol if you'd prefer the one-step relation be visually distinct from the recursive ⟿.

### CI
Green locally on GHC 9.8.4: `cabal test --ghc-options=-Werror` (1073 examples, 0 failures), `fourmolu` clean, `hlint` 0 hints. (The latex smoke test and hlint both needed small follow-ups — the latter because dropping the `pattern` field name let hlint fully parse the module and surface latent `Eta reduce` hints on the `phino*` `printf` helpers, now point-free.)

---

maxonfjvipon (commented):
@yegor256 What about ExtraArgument type? It seems it's not parsed anymore and used only for bridge from new premises front end to old when+where front end? If so, it would be better to get rid of it. Also what about `having`? How this new premises approach resolves it? We need it for `hone-maven-plugin`

src/LaTeX.hs
yegor256: Good questions — two clarifications:

**`ExtraArgument` is still parsed and load-bearing**, so we can't just drop it. The general build-term premise form — `{ meta, function, args }`, used by `sed` / `concat` / `size` / `random-tau` / `dataize` / etc. — is parsed straight into `Extra` (`FromJSON Extra`/`FromJSON ExtraArgument`), and the rewriting engine consumes `[Extra]` directly in `extraSubstitutions`. So it isn't only a bridge; it's the engine's representation for build-term premises.

The **real bridge** is `premiseToExtra`: the verb-keyed premises (`contextualize:`/`morph:`/…) are parsed as the typed `Operation`/`Premise` (shared with 𝕄/𝒞/𝔻) and then desugared into `Extra`. So there are two premise representations and one converter between them. Fully removing it means having the engine consume `Premise` directly and retiring `Extra`/`ExtraArgument` — a larger change than this PR, so I've added a pdd todo at `premiseToExtra` (`@todo #912:60min …`) to track it. Happy to do it here instead if you'd prefer.

**`having` is untouched** — still parsed (`o .:? "having"`) and applied as the post-premises side condition in `Rule.hs` (`meetMaybeCondition rule.having` after the premises run). The new schema only renamed `pattern→match` and `where→premises`; `when`/`having` are unchanged, so `hone-maven-plugin` keeps working. It's covered after the migration by [`part-of-after.yaml`](https://github.com/objectionary/phino/blob/claude/phino-issue-912-fp1u0i/test-resources/rewriter-packs/custom/part-of-after.yaml), which uses `match` + a `concat` premise + a `having: { not: { part-of: … } }` guard.

---

maxonfjvipon:
@yegor256 can we close it?

0crat:
@maxonfjvipon Thanks for the review! You've earned +13 points for this: +12 as a basis; +5 for the 553 hits-of-code that you reviewed; -5 for very few (1) comments.  Your running score is [+746](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #929, opened 2026-06-30
fix(#928): drop state subscript when a rule threads a single state

### LINKED ISSUE #928: explain --morph labels a single state as s_1 instead of s in src/LaTeX.hs
Running `phino explain --morph` renders rules that thread only one state with a subscripted state name. For example the `mg` rule prints `\phinoMorph{ Q }{ Q }{ s_1 }{ T }{ s_1 }`, where both the incoming and outgoing state are the same `s_1`. The same happens for `none`, `end`, `xi`, `dead`, `mf` and the lambda-labelled rule — every rule whose premises never thread a new state.

This is wrong because the subscript exists only to distinguish multiple states threaded through 𝕄/𝔻/𝔼 premises (`s_1`, `s_2`, …). When a rule has no state-changing premises there is exactly one state, so the subscript carries no information and should read as a bare `s` — the conclusion ought to be `\phinoMorph{ Q }{ Q }{ s }{ T }{ s }`. The cause is `stateName` in `src/LaTeX.hs:419-420`, which unconditionally emits `"s_" ++ show n`; `explainMorphRule` (line 385) then passes index 1 for both the incoming and the final state when `final == 1` (from `premisesToLatex`, line 427).

Drop the subscript when a rule uses a single state: have `stateName` (or its callers in `explainMorphRule`/`explainDataizeRule`) emit `s` instead of `s_1` whenever the rule's final state index is 1, keeping the subscripted form only when more than one state is threaded.

yegor256:
Closes #928.

## Problem

`phino explain --morph` (and `--dataize`) rendered the conclusion of every rule whose premises never thread a new state with a subscripted state name, e.g. the `mg` rule printed `\phinoMorph{ Q }{ Q }{ s_1 }{ T }{ s_1 }`. The subscript only exists to distinguish the several states threaded through 𝕄/𝔻/𝔼 premises (`s_1`, `s_2`, …). When a rule has a single state, the subscript carries no information and should read as a bare `s`.

This affected `none`, `end`, `xi`, `dead`, `mf`, `mg`, `delta` and the lambda-labelled rule.

## Fix

In `src/LaTeX.hs`, added a `conclusionStateName` helper used by `explainMorphRule` and `explainDataizeRule`. When the rule's final state index is `1` (a single threaded state) it emits a bare `s`; otherwise it keeps the subscripted form (`s_1` … `s_final`) shared with the premises. The premise rendering (`premiseToLatex`) is unchanged, since state-rendering premises only ever appear when more than one state is threaded.

Updated the expected outputs in `test/CLISpec.hs` for the single-state morphing rules (`mf`, `dead`, `xi`, `mg`) and dataization rules (`delta`, `none`, `end`).

https://claude.ai/code/session_01HBquczUbG74PohVvuuNUtR

---
_Generated by [Claude Code](https://claude.ai/code/session_01HBquczUbG74PohVvuuNUtR)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1232](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #935, opened 2026-06-30
fix(#934): escape condition-set commas as \char44{}

### LINKED ISSUE #934: explain --dataize emits raw comma in condition sets instead of \char44{}
`phino explain --dataize` renders the `box` rule condition as `\phinoCondition{ [ D, L ] \cap \lparen B_1 \cup B_2 \rparen = \emptyset }`. The comma separating the set elements comes from the hard-coded `T.intercalate ", "` in `render CO_DISJOINT` at `src/Render.hs:272`, and the same literal sits in `render ST_ATTRIBUTES` at `src/Render.hs:230`. The `none` rule is affected the same way, printing `[ D, L, @ ]`.

This is wrong because, inside `\phinoCondition`, the bracketed list is a meta-language set rather than a phi-calculus expression, so its separator must be the LaTeX-safe `\char44{}` and the set should read `[ D \char44{} L ]`. This mirrors the escapes already present in the `ToLaTeX Text` instance at `src/LaTeX.hs:313-317` (`\char36{}`, `\char64{}`, …), which never reach the comma.

A concrete fix is to change the `", "` separator in those two `render` clauses to `" \char44{} "`; condition sets render only through the `explain` LaTeX path, so plain phi output stays unaffected. Note that formation binding commas — for example the `[[ B_1, L> F, B_2 ]]` shown in the `fire` rule — are genuine phi-calculus syntax and must stay plain. The `[ D, L ]` and `[ D, L, @ ]` expectations in `test/CLISpec.hs` should be updated to the escaped form.

yegor256:
Closes #934.

`phino explain --dataize` was rendering condition sets such as the `box` and `none` rules with a raw comma separator (`[ D, L ]`, `[ D, L, @ ]`). Inside `\phinoCondition`, the bracketed list is a meta-language set rather than a phi-calculus expression, so its separator must be the LaTeX-safe `\char44{}` — mirroring the escapes already applied in the `ToLaTeX Text` instance.

## Changes

- `src/Render.hs`: change the `T.intercalate ", "` separator to `T.intercalate " \char44{} "` in both `render ST_ATTRIBUTES` and `render CO_DISJOINT`. Condition sets render only through the `explain` LaTeX path, so plain phi output is unaffected. Formation binding commas (e.g. `[[ B_1, L> F, B_2 ]]`) are genuine phi-calculus syntax and remain plain — they go through a different code path.
- `test/CLISpec.hs`: update the `[ D, L ]` and `[ D, L, @ ]` expectations to the escaped `[ D \char44{} L ]` / `[ D \char44{} L \char44{} @ ]` form.

## Note

No Haskell toolchain is available in this remote environment, so the test suite was not run locally. The changes are exact, minimal substitutions matching the issue's prescription.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wb7hdddpriZF4yP1EKvmT2)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1328](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #937, opened 2026-06-30
fix(#936): add `dl` normalization rule collapsing λ+Δ formations to ⊥

yegor256:
Closes #936.

## Problem

The normalization rule set in `resources/normalize/` had no rule to terminate a formation that holds **both** a λ binding and a Δ binding. Such an object is simultaneously an atom (λ) and data (Δ), which is contradictory and should normalize to ⊥. Today such a formation is left untouched by the rewriter.

This was inconsistent with the rest of the calculus, which already assumes the two are mutually exclusive — the `box` dataization rule guards itself with `disjoint: [[Δ, λ], [𝐵1, 𝐵2]]` (`resources/dataization.yaml`) — but nothing enforced that invariant during normalization, unlike the other ill-formed shapes (`amiss`, `miss`, `over`, `stop`).

## Change

- Add `resources/normalize/dl.yaml`, mirroring `stop.yaml`: pattern `⟦𝐵⟧`, result `⊥`, and a `when: and:` of `in: [Δ, 𝐵]` and `in: [λ, 𝐵]`. The directory is embedded via `embedDir "resources/normalize"` (`src/Yaml.hs`), so the file is picked up automatically.
- Add `test-resources/rewriter-packs/normalize/e-dl.yaml` — a formation `⟦ Δ ⤍ 42-, λ ⤍ Fn ⟧` normalizes to `⊥`.
- Update the `explains normalization rules` assertion in `test/CLISpec.hs` to include the new `dl` rule (rendered alphabetically between `dd` and `dot`).

## Verification

Full test suite passes: `1079 examples, 0 failures`. Confirmed `phino explain --normalize` emits the new rule in the expected LaTeX form.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01F5ZV3UFHRCPNgWcAj9cgMz)_

PR #938, opened 2026-06-30
perf(#936): match `dl` on the rare λ asset to avoid normalize regression

### LINKED ISSUE #936: Missing normalization rule `dl`: ⟦𝐵⟧ holding both λ and Δ should reduce to ⊥
The normalization rule set in `resources/normalize/` has no rule that terminates a formation holding both a λ binding and a Δ binding. An object that is simultaneously an atom (λ) and data (Δ) is contradictory and should normalize to ⊥, but today such a formation is left untouched by the rewriter.

This is inconsistent with the rest of the calculus, which already assumes the two are mutually exclusive: the `box` dataization rule guards itself with `disjoint: [[Δ, λ], [𝐵1, 𝐵2]]` (`resources/dataization.yaml:39-42`). Nothing enforces that invariant during normalization, so a malformed `⟦Δ ⤍ ..., λ ⤍ ...⟧` slips through instead of collapsing to termination like other ill-formed shapes do (`amiss`, `miss`, `over`, `stop`).

Add a new `resources/normalize/dl.yaml`, mirroring `stop.yaml`, with pattern `⟦𝐵⟧`, result `⊥`, and a `when: and:` of `in: [Δ, 𝐵]` and `in: [λ, 𝐵]`. The directory is embedded via `embedDir "resources/normalize"` (`src/Yaml.hs:215`), so the file is picked up automatically.

yegor256:
Follow-up to #937 (which fixed #936). That PR merged with a failing `regression-check`: the `dl` rule regressed the `rewrite/normalize` benchmark by +24.9%, past the +20% gate. This PR keeps the exact same behavior while removing that cost.

## Root cause

`dl` shipped with a bare formation pattern, `⟦𝐵⟧`. Every other normalization rule anchors on a dispatch (`⟦𝐵⟧.𝜏`) or application (`⟦𝐵⟧(…)`) node, which are comparatively rare. A bare formation pattern matches **every** formation node in the tree (~9,948 in the bench fixture) and fires its IO `when` condition on each, every normalization cycle — that's the regression. It's inherent to matching bare formations, not a logic bug.

## Change

Re-anchor the pattern on the **λ** binding and check that Δ sits among the remaining bindings:

```yaml
pattern: ⟦𝐵1, λ ⤍ 𝑓, 𝐵2⟧
result: ⊥
when:
  or:
    - in:
        - Δ
        - 𝐵1
    - in:
        - Δ
        - 𝐵2
```

λ bindings are far rarer than formations (3 vs ~9,948 in the fixture), so the rule produces a handful of candidate matches per cycle instead of one per formation. This mirrors the `box`/`delta`/`fire` dataization rules' λ/Δ idioms.

The `or` guard keeps the rule **order-independent** — both `⟦Δ ⤍ …, λ ⤍ …⟧` and `⟦λ ⤍ …, Δ ⤍ …⟧` still collapse to ⊥. It also reads cleanly in the `explain --normalize` LaTeX output (`D \in B_1 \;\text{or}\; D \in B_2`). (An earlier candidate that put both assets in the pattern was cheaper still but order-dependent, so it would silently miss λ-before-Δ; rejected for correctness.)

## Benchmark (CI `regression-check`, A/B vs merge-base)

`rewrite/normalize`: base 73,749 µs → head 56,975 µs = **−22.7% (improved)**. All other benchmarks flat (±2%). The base here is `master`, which already contains the merged bare-`⟦𝐵⟧` version, so this PR effectively reverses the regression #937 introduced and returns `rewrite/normalize` to roughly its pre-`dl` baseline.

## Tests

- `test/CLISpec.hs` — updated the `explain --normalize` assertion to the new `dl` rendering.
- `test-resources/rewriter-packs/normalize/e-dl.yaml` — existing Δ-before-λ case (unchanged).
- `test-resources/rewriter-packs/normalize/e-dl2.yaml` — **new** λ-before-Δ case, locking in order-independence.

Full suite: `1080 examples, 0 failures`.


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1336](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #945, opened 2026-06-30
fix(dataize): `none` rule delegates to `end` instead of duplicating `--`

### LINKED ISSUE #942: `none` dataization rule duplicates `end`'s `--` instead of delegating to it
The `none` dataization rule in `resources/dataization.yaml:67-74` produces the
empty-bytes result `--` directly via `d-result: '--'`, exactly as the `end` rule
does at `resources/dataization.yaml:76-79`. The terminal "no bytes" value is thus
hard-coded in two clauses, even though `none` is conceptually a special case that
should bottom out in `end`.

This duplication is a smell: the meaning of "an empty formation dataizes to
nothing" is split across two rules, so any future change to how termination
yields bytes has to be made in both places, and the `none` clause no longer
reads as "this reduces to ⊥". The intent is that `none` should not emit `--`
itself but delegate to `end` (which matches `⊥`).

Concretely, `none` should become a premise-conclusion rule:

- conclusion: 𝔻(⟦𝐵⟧, 𝑒, 𝑠) → ⟨δ, 𝑠⟩
- condition: [Δ, λ, φ] ∩ [𝐵] = ∅  (unchanged)
- premise: 𝔻(⊥, 𝑒, 𝑠) → ⟨δ, 𝑠⟩

i.e. change `d-result: '--'` to `d-result: δ` and add a `dataize: ⊥` premise that
binds δ, leaving the `when` block as-is. Note this requires a `dataize` premise to
accept the literal `⊥` as its argument (today the `box`/`fire`/`norm` premises pass
a meta such as `𝑛1`); `⊥` is already recognized in `norm`'s `eq` guard and as
`end`'s `match`, so the matcher knows the symbol, but `Dataize.hs`'s premise
handling may need to allow a literal term there.

yegor256:
Closes #942.

## Problem

The `none` dataization rule (`resources/dataization.yaml`) produced the
empty-bytes result `--` directly via `d-result: '--'`, exactly as the `end`
rule does. The terminal "no bytes" value was thus hard-coded in two clauses,
even though `none` is conceptually a special case that should bottom out in
`end` (which matches `⊥`).

## Change

`none` is now a premise-conclusion rule that delegates to `end`:

- **conclusion**: 𝔻(⟦𝐵⟧, 𝑒, 𝑠) → ⟨δ, 𝑠⟩
- **condition**: `[Δ, λ, φ] ∩ [𝐵] = ∅` (unchanged)
- **premise**: 𝔻(⊥, 𝑒, 𝑠) → ⟨δ, 𝑠⟩

i.e. `d-result: '--'` became `d-result: δ` plus a `dataize: ⊥` premise binding
δ. An empty formation now reduces `𝔻(⟦𝐵⟧) → 𝔻(⊥)` and `end` yields the empty
bytes, so the meaning of "an empty formation dataizes to nothing" lives in one
place.

The existing `_` branch of `dataize'`'s `reduce` already builds a literal
premise argument via `buildExpressionThrows`, so the matcher needs no new
support to accept the literal `⊥`. That branch now labels the
side-computation-free transition by the conclusion's own verb (`dataize`)
through a small `labelOr` helper, so every reduction step in the trace still
carries a defining label.

## Tests

- Updated the `explain --dataize` LaTeX golden in `CLISpec` to show the new
  premise on `none`.
- Added a `DataizeSpec` case asserting that an empty formation reduces through
  a single labelled `dataize` step before `end`.
- Full suite green under `make test` (`cabal test --ghc-options=-Werror`,
  1080 examples, 0 failures); `fourmolu --mode check` and `hlint` clean on the
  touched files.

---
_Generated by [Claude Code](https://claude.ai/code/session_01E5SCjmQhisEYajxtYGpowu)_


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (27) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1344](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #949, opened 2026-06-30
fix(#947): express mg morphing rule as a premise that morphs ⊥

### LINKED ISSUE #947: Express the `mg` morphing rule as a premise that morphs `⊥`, not a literal
In `resources/morphing.yaml:115` the `mg` rule short-circuits to a hardcoded literal — it fires only on `𝕄(Φ, Φ)` (the head is `Φ` and the universe is pinned to `Φ`) and sets `n-result: ⊥` directly, without an ordered premise:

```yaml
- name: mg
  match: Φ
  e-match: Φ
  n-result: ⊥
```

This is the same odd shape as the `xi` rule in #946. Every other 𝕄 rule that yields a derived term states the judgment that produces it through an ordered premise binding `𝑛` (see the header comment, lines 16–24), and #912 is migrating the remaining rules onto that premise-conclusion schema. `mg` instead names the answer rather than the reduction that reaches it. Since the `dead` rule already gives `𝕄(⊥, e, s) → (⊥, s)`, morphing `⊥` and binding its result is behaviorally identical to the literal — so this is a formal alignment, not a behavior change.

The fix is to replace the literal `n-result: ⊥` with `n-result: 𝑛` and a single `morph: ⊥` premise, so the rule reads as the inference `𝕄(⊥, Φ, s) → (𝑛, s)` over `𝕄(Φ, Φ, s) → (𝑛, s)`:

```yaml
- name: mg
  match: Φ
  e-match: Φ
  n-result: 𝑛
  premises:
    - n-result: 𝑛
      morph: ⊥
```

yegor256:
Closes #947.

## Problem

The `mg` rule in `resources/morphing.yaml` fired on 𝕄(Φ, Φ) and assigned its conclusion directly to the literal `⊥`, instead of following the premise-conclusion inference schema used by every other 𝕄 rule.

## Fix

Refactor `mg` to derive its result through a premise rather than naming the literal:

```yaml
- name: mg
  match: Φ
  e-match: Φ
  n-result: 𝑛
  premises:
    - n-result: 𝑛
      morph: ⊥
```

This expresses the rule as the inference "𝕄(Φ, Φ, s) → (𝑛, s)" derived from "𝕄(⊥, Φ, s) → (𝑛, s)", mirroring the way dataization's `none` rule delegates to `end` (#945). Behavior is unchanged: the premise morphs `⊥`, which matches the `dead` rule and returns `⊥`, so the final morphed term is identical.

## Tests

Updated the `explain --morph` LaTeX expectation in `test/CLISpec.hs` so the `mg` inference now renders its `morph ⊥` premise above the conclusion line.

---
_Generated by [Claude Code](https://claude.ai/code/session_01LAB6P8hvGyr8a89BT9AnBw)_

### DIFF
--- a/resources/morphing.yaml
+++ b/resources/morphing.yaml
@@ -115,4 +115,7 @@
 - name: mg
   match: Φ
   e-match: Φ
-  n-result: ⊥
+  n-result: 𝑛
+  premises:
+    - n-result: 𝑛
+      morph: ⊥
--- a/test/CLISpec.hs
+++ b/test/CLISpec.hs
@@ -1101,7 +1101,8 @@ spec = do
             , "\\end{phinoMorphingInference}"
             , "\\begin{phinoMorphingInference}"
             , "  \\phinoName{mg}"
-            , "  \\phinoConclusion{ \\phinoMorph{ Q }{ Q }{ s }{ T }{ s } }"
+            , "  \\phinoPremise{ \\phinoMorph{ T }{ e }{ s_1 }{ n }{ s_2 } }"
+            , "  \\phinoConclusion{ \\phinoMorph{ Q }{ Q }{ s_1 }{ n }{ s_2 } }"
             , "\\end{phinoMorphingInference}"
             ]
         ]


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1352](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #956, opened 2026-07-01
fix(#955): drop dataization `end` rule so 𝔻(⊥) fails instead of yielding empty bytes

### LINKED ISSUE #955: Dataization `end` rule maps terminator ⊥ to empty bytes instead of failing
The `end` rule in `resources/dataization.yaml:79-82` matches the terminator `⊥` and concludes `d-result: '--'`, so `𝔻(⊥)` returns empty bytes. The `none` rule (`resources/dataization.yaml:67-77`) reaches the same outcome by dataizing `⊥`.

This is wrong because `⊥` (the terminator `T`) is defined as an expression that *signals an error*, whereas `--` is a legitimate datum: `⟦Δ ⤍ --⟧` is a real zero-length byte array. The `end` rule conflates "the computation reached a dead end" with "the computation produced empty data", so a program that fails becomes indistinguishable from one that genuinely yields empty bytes. It also contradicts the calculus paper, where `𝔻` is a *partial* function (`⇀`): `⊥` should lie outside its domain rather than map to a value.

The fix is to drop the `end` rule so `𝔻(⊥)` matches no clause and dataization stops at `⊥`. `src/Dataize.hs:184` already throws on an unmatched term (`no dataization rule matched`); sharpen that message to name the terminator. Since `none` already routes through `dataize: ⊥`, a data-less formation then fails through the same path, which is consistent — such a formation has nothing to dataize.

yegor256:
Fixes #955.

## Problem

The `end` rule in `resources/dataization.yaml` matched the terminator `⊥` and concluded `d-result: '--'`, so `𝔻(⊥)` returned empty bytes. This conflated two distinct outcomes:

- **the computation reached a dead end** (`⊥`, the terminator `T`, which *signals an error*), and
- **the computation produced empty data** (`--`, a legitimate zero-length byte array `⟦Δ ⤍ --⟧`).

A program that fails became indistinguishable from one that genuinely yields empty bytes. It also contradicted the calculus paper, where `𝔻` is a *partial* function (`⇀`): `⊥` should lie outside its domain rather than map to a value.

## Fix

- **`resources/dataization.yaml`** — remove the `end` rule so `𝔻(⊥)` matches no clause. The `norm` guard (`not (𝑛 = ⊥)`) still keeps `norm` from matching `⊥`, so dataization simply stops at `⊥`.
- **`src/Dataize.hs`** — the unmatched-term path (`no dataization rule matched`) now names the terminator: `dataization reached the terminator ⊥, which signals an error and cannot be dataized`. Updated the surrounding doc comments (`𝔻` is now partial, not total; `asNumber`'s empty-bytes note).
- Since `none` already routes a data-less formation `⟦𝐵⟧` through `dataize: ⊥`, such a formation now fails through the same path — consistent, as it has nothing to dataize.

## Tests

- `test/DataizeSpec.hs` — replaced the `T => --` and `⟦⟧ => --` success cases with a `fails to dataize the terminator` block asserting both `⊥` and a data-less formation throw; the empty-formation label-sequence test now expects a throw.
- `test/CLISpec.hs` — the empty-object case now expects failure with `terminator ⊥`; removed the `end` block from the `explain --dataize` expected output.

---
_Generated by [Claude Code](https://claude.ai/code/session_01T1gTXUxAwcMnkNCfbH45vD)_

### DIFF
--- a/.github/workflows/latex.yml
+++ b/.github/workflows/latex.yml
@@ -59,7 +59,7 @@ jobs:
             printf '\\begin{document}\n'
             "${e}" explain --normalize
             printf '{[[ x -> [[ y -> ? ]].y ]]}' | "${e}" rewrite --normalize --output=latex --sequence
-            printf '{[[ x -> [[ y -> ? ]].y ]]}' | "${e}" dataize --output=latex --sequence --quiet
+            printf '{[[ @ -> [[ x -> [[ D> 01-, y -> ? ]](y -> [[ ]]) ]].x ]]}' | "${e}" dataize --output=latex --sequence --quiet
             printf '\\end{document}\n'
           ) > article.tex
           pdflatex -shell-escape -interaction=errorstopmode -halt-on-error article
--- a/resources/dataization.yaml
+++ b/resources/dataization.yaml
@@ -22,9 +22,11 @@
 # 'norm' matches the bare meta 𝑛, which unifies with any expression, so it is
 # guarded to fire only when 𝑛 is neither a formation ('not (formation 𝑛)',
 # carving out 'delta', 'box', 'fire' and 'none') nor the termination ⊥
-# ('not (𝑛 = ⊥)', carving out 'end'). Without the guard 'norm' would behave
-# correctly only by being declared last; with it the clauses no longer rely on
-# their order.
+# ('not (𝑛 = ⊥)'). 𝔻 is partial: ⊥ (the terminator T) signals an error and
+# lies outside its domain, so it deliberately matches no clause and dataization
+# stops there — there is no 'end' rule mapping ⊥ to empty bytes (see #955).
+# Without the guard 'norm' would behave correctly only by being declared last;
+# with it the clauses no longer rely on their order.
 
 - name: delta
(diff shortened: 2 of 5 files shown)


yegor256:
@maxonfjvipon good?

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1349](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+493](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #968, opened 2026-07-03
fix(#967): DOT excludes dispatched attribute from contextualization context

### LINKED ISSUE #967: DOT normalization loops forever on self-referential objects instead of reducing to ⊥
The `dot` rule in `resources/normalize/dot.yaml:12` passes the whole formation `⟦𝐵1, 𝜏 ↦ 𝑛, 𝐵2⟧` — including the dispatched binding `𝜏 ↦ 𝑛` — as the contextualization context for `ξ`. When the body `𝑛` refers back to the dispatched attribute through `ξ` (for example `⟦ a ↦ ξ ⟧.a` or `⟦ a ↦ ξ.a ⟧.a`), contextualization rebuilds a `⟦…, 𝜏 ↦ 𝑛, …⟧.𝜏` term that DOT immediately matches again, so normalization never terminates. Mutual cases like `⟦ a ↦ ξ.b, b ↦ ξ.a ⟧.a` diverge the same way.

This makes the whole normalization function partial. Normalization is meant to be the syntactic, (near-)total layer, with productive recursion living in morphing/dataization. A self-referential object has no normal form and should reduce to ⊥ (termination), not spin.

Change the context argument on `resources/normalize/dot.yaml:12` from `⟦𝐵1, 𝜏 ↦ 𝑛, 𝐵2⟧` to `⟦𝐵1, 𝐵2⟧`, so `ξ` inside `𝑛` no longer sees the dispatched attribute; a self-reference then dispatches on a formation without `𝜏` and collapses to ⊥ via `stop`/`null`/`dd`. `result` on line 6 still binds the full formation to `ρ`, so sibling and φ-decoration references are unaffected — only the self-`ξ` path narrows. This removes the DOT self-dispatch divergence class (not divergence in general). The expected outputs in `test-resources/rewriter-packs/basic/dot-xi.yaml` and the `dot-paper-*` packs, plus the contextualization prose in `resources/contextualization.yaml`, will need to be updated to match.

yegor256:
Fixes #967.

## Problem

The `dot` rule in `resources/normalize/dot.yaml` passed the whole formation `⟦𝐵1, 𝜏 ↦ 𝑛, 𝐵2⟧` — including the dispatched binding `𝜏 ↦ 𝑛` — as the contextualization context for `ξ`. When the dispatched body `𝑛` referred back to the dispatched attribute through `ξ` (e.g. `⟦ a ↦ ξ ⟧.a`, `⟦ a ↦ ξ.a ⟧.a`, or mutual cases like `⟦ a ↦ ξ.b, b ↦ ξ.a ⟧.a`), contextualization rebuilt a `⟦…, 𝜏 ↦ 𝑛, …⟧.𝜏` term that DOT immediately matched again, so normalization never terminated — making the whole normalization function partial.

## Fix

Narrow the contextualization context on `dot.yaml` from `⟦𝐵1, 𝜏 ↦ 𝑛, 𝐵2⟧` to `⟦𝐵1, 𝐵2⟧`, so a self-referential `ξ` inside `𝑛` no longer sees the dispatched attribute. A self-reference then dispatches on a formation without `𝜏` and collapses to ⊥ via `stop`/`null`/`dd`, so normalization terminates. `result` still binds the full formation to `ρ`, so sibling and φ-decoration references are unaffected — only the self-`ξ` path narrows. This removes the DOT self-dispatch divergence class (not divergence in general).

## Changes

- `resources/normalize/dot.yaml` — the one-line context change, plus explanatory prose.
- `resources/contextualization.yaml` — clarify that the caller supplies the context and how DOT chooses it.
- Refreshed expected outputs that depended on the old context: `dot-xi`, `dot-recursive`, `dot-paper-e2`, `dot-paper-e4-1`, `dot-paper-e5-1`, and the full-normalization packs `e-cpy`, `e-fnk`, `e-phi`, `e-rep`, `e-twi`.
- Updated the `explain`/LaTeX expectations in `test/CLISpec.hs` (the dot rule rendering and the two `--meet-prefix`/`--compress` sequences, plus the mutual-recursion limit example).
- Added regression packs `e-self.yaml` (`⟦ a ↦ ξ.a ⟧.a → ⊥`) and `e-mut.yaml` (`⟦ a ↦ ξ.b, b ↦ ξ.a ⟧.a → ⊥`) covering the previously-diverging cases.

## Verification

Built with GHC 9.6.7 and ran the full suite: **1083 examples, 0 failures**.

https://claude.ai/code/session_01Bem4P8xAzuGYwRkAfpVEKL

---
_Generated by [Claude Code](https://claude.ai/code/session_01Bem4P8xAzuGYwRkAfpVEKL)_


yegor256:
@maxonfjvipon agree?

test-resources/rewriter-packs/normalize/e-self.yaml
```diff
@@ -0,0 +1,12 @@
...
+---
+# A self-referential object dispatched on its own attribute used to loop
+# forever in DOT (see #967); now the dispatched attribute is excluded from the
+# contextualization context, so the self-ξ collapses to ⊥ and normalization
+# terminates.
+normalize: true
+input: |
+  ⟦ a ↦ ξ.a ⟧.a
+output: |-
+  ⊥
```
maxonfjvipon, line 12: @yegor256 are you sure this is the right way to normalize such expression?

yegor256: @maxonfjvipon yes, that's how it should be - we must prevent endless recursion and this expression means error in normalization. We keep recursion possible, but _only_ through morphing (lambda functions).

maxonfjvipon (approved):

maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+534](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1699](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #970, opened 2026-07-03
fix(#969): canonize after meet compression so --compress still dedups

### LINKED ISSUE #969: `--canonize` defeats `--compress`: identical subexpressions with different F-numbers are never merged into `\phinoMeet`
When `--compress` and `--canonize` are used together for LaTeX sequence output (e.g. `dataize --output=latex --sequence --compress --canonize`), subexpressions that are structurally identical across reduction steps are not factored out into `\phinoMeet`/`\phinoAgain` — they are printed in full again and again. The cause is the ordering and independence of the two passes. `canonize` (`src/CLI/Runners.hs:64`) runs before the meet compression (`printRewrittens` at `src/CLI/Runners.hs:66`, which reaches `meetInExpressions` in `src/LaTeX.hs:98`), and `canonize` renames each step's λ-bindings to `F1`, `F2`, … restarting the counter at `1` for every expression (`src/Canonizer.hs:49`, numbering by traversal position at `src/Canonizer.hs:17`).

Because the counter restarts per step and numbers by position, the *same* logical lambda gets a *different* `Fn` in different steps (for instance a `number` object whose `times`/`plus` lambdas come out as `F3`/`F4` in one step and `F4`/`F5` in another). The meet pass compares candidates by structural equality (`matchExpression` / `==` inside `meetInExpressions`), so `Function "F3"` and `Function "F4"` are treated as distinct and the enclosing subexpressions never match. `--canonize` therefore silently sabotages the deduplication that `--compress` is meant to perform; dropping `--canonize` makes the same subexpression collapse to a single `\phinoMeet` as expected.

The smallest fix is to canonize *after* the meet compression instead of before it, so the meet pass sees the stable original lambda names: feed un-canonized expressions to `meetInExpressions` and apply `canonize` to its output. This also requires adding `ExPhiMeet`/`ExPhiAgain` cases to `canonizeExpression` (`src/Canonizer.hs:37`) so the body factored out into `\phinoMeet{…}{ … }` still gets its lambdas renamed.

yegor256:
Closes #969.

## Problem

When `--compress` and `--canonize` are combined for LaTeX sequence output
(e.g. `dataize --output=latex --sequence --compress --canonize`),
structurally identical sub-expressions across reduction steps were **not**
factored out into `\phinoMeet`/`\phinoAgain` — they were printed in full over
and over.

The cause was ordering: `canonize` ran *before* the meet compression. It
renames each step's λ-bindings to `F1`, `F2`, … numbering by traversal
position and **restarting the counter for every expression**. Because the
counter restarts per step and numbers by position, the *same* logical lambda
got a *different* `Fn` in different steps. The meet pass compares candidates by
structural equality, so `Function "F3"` and `Function "F4"` looked distinct and
the enclosing sub-expressions never matched — `--canonize` silently sabotaged
the deduplication `--compress` is meant to perform.

## Fix

Canonization now runs **after** the meet compression, so the meet pass sees the
stable, original lambda names:

- The `_canonize` flag is threaded through `PrintContext`/`LatexContext` and
  applied at render time rather than pre-applied to the chain in the runners
  (`runRewrite`/`runDataize` no longer canonize up front).
- For the LaTeX sequence path, `rewrittensToLatex` feeds un-canonized
  expressions to `meetInExpressions` (via `compressedRewrittens` /
  `compressedExpressions`) and canonizes only its output
  (`canonizedRewrittens` / `canonizedExpressions`).
- For the other output formats (no meet pass), `printRewrittens` canonizes just
  before rendering, preserving the previous behavior.
- `canonizeExpression` gains `ExPhiMeet`/`ExPhiAgain` cases so lambdas inside a
  factored-out `\phinoMeet{…}` body still get renamed.

## Tests

Added a `dataize --output=latex --compress --canonize --sequence` CLI test over
a whole-expression sequence that asserts a `\phinoMeet` is produced — the
scenario that previously collapsed only when `--canonize` was dropped. The
existing focused compress+canonize test continues to pass.

https://claude.ai/code/session_01JJxZDChgfupZEqUMbtSzHm

---
_Generated by [Claude Code](https://claude.ai/code/session_01JJxZDChgfupZEqUMbtSzHm)_

### DIFF
--- a/src/CLI/Helpers.hs
+++ b/src/CLI/Helpers.hs
@@ -9,6 +9,7 @@ module CLI.Helpers where
 import AST
 import CLI.Types
 import CLI.Validators (invalidCLIArguments)
+import Canonizer (canonize)
 import Control.Exception
 import Control.Monad ((>=>))
 import Data.Functor ((<&>))
@@ -24,7 +25,7 @@ import Logger
 import Parser (parseExpressionThrows)
 import qualified Printer as P
 import qualified Random as R
-import Rewriter (Rewrittens')
+import Rewriter (Rewritten, Rewrittens')
 import System.IO (getContents')
 import Text.Printf (printf)
 import XMIR (expressionToXMIR, parseXMIRThrows, printXMIR, xmirToPhi)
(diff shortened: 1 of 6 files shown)


yegor256:
@rultor merge

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1707](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #979, opened 2026-07-07
fix(#978): keep compressed meet bodies on one line under --margin

### LINKED ISSUE #978: `--margin` line-breaks inside `\phinoMeet{}` bodies, so compressed meets are no longer atomic
When `--compress` and `--margin` are combined in LaTeX output, the margin pass wraps the *inside* of a `\phinoMeet{id}{expr}` group whenever `expr` is wider than the margin. `withMargin'` recurses into the meet body at `src/Margin.hs:34` (`EX_PHI_MEET prefix idx (withMargin' cfg expr)`), and `render` at `src/Render.hs:208` then emits that multi-line `expr` between the `{ ... }` braces.

A `\phinoMeet` is drawn as a single `\overbracket` with a `Π` label — an atomic, one-line visual unit — so breaking lines inside its body is wrong on its own. It also breaks typesetting downstream: in an `aligned`/`gathered` context each newline in the body becomes a row break (`\\`), and a row break inside a braced macro argument raises `! Missing } inserted`. With `--flat` the body is always one line, so the fault only surfaces once margin wrapping is enabled.

Reproduce (phino 0.0.103) with the celsius example from #976:

```
phino dataize --output=latex --sweet --compress --canonize --sequence \
  --meet-prefix=m --meet-popularity=1 --meet-length=3 --margin=80 \
  --locator=Q.@ --focus=Q.@ ./examples/celsius.phi
```

The `m:5` meet opens `\phinoMeet{m:5}{ [[` and its closing `]] }` lands several lines later.

The smallest fix is to keep the meet atomic: render its body on one line regardless of margin. At `src/Margin.hs:34`, use `toSingleLine expr` (already imported and used in that module) instead of `withMargin' cfg expr`, so `EX_PHI_MEET` never line-wraps its argument.

yegor256:
Fixes #978.

## Problem

When `--compress` and `--margin` are combined in LaTeX output, the margin pass wrapped the *inside* of a `\phinoMeet{id}{expr}` group whenever `expr` was wider than the margin. `withMargin'` recursed into the meet body at `src/Margin.hs:34`, and `render` at `src/Render.hs:208` then emitted that multi-line body between the `{ ... }` braces.

A `\phinoMeet` is drawn as a single `\overbracket` with a `Π` label — an atomic, one-line visual unit — so breaking lines inside its body is wrong. It also breaks typesetting: in an `aligned`/`gathered` context each newline becomes a row break (`\\`), and a row break inside a braced macro argument raises `! Missing } inserted`.

## Fix

In `src/Margin.hs`, the `EX_PHI_MEET` case now renders its body with `toSingleLine expr` (already imported and used in the module) instead of `withMargin' cfg expr`, so the meet argument never line-wraps. The margin config argument is unused in this case and is bound to `_`.

## Test

Added a regression test in `test/PrinterSpec.hs` that embeds a wide `ExPhiMeet` in a formation and prints it with a narrow margin (`MULTILINE`, margin `30`). The outer formation wraps, but the meet body must stay on a single line. Without the fix the body wraps and the assertion fails.

https://claude.ai/code/session_01S21BqDo81wicpb4exuFDSp

---
_Generated by [Claude Code](https://claude.ai/code/session_01S21BqDo81wicpb4exuFDSp)_

### DIFF
--- a/src/Margin.hs
+++ b/src/Margin.hs
@@ -31,7 +31,7 @@ instance WithMargin EXPRESSION where
   withMargin' _ str@EX_STRING{} = str
   withMargin' cfg EX_DISPATCH{..} = EX_DISPATCH (withMargin' cfg expr) space attr
   withMargin' cfg EX_PHI_AGAIN{..} = EX_PHI_AGAIN prefix idx (withMargin' cfg expr)
-  withMargin' cfg EX_PHI_MEET{..} = EX_PHI_MEET prefix idx (withMargin' cfg expr)
+  withMargin' _ EX_PHI_MEET{..} = EX_PHI_MEET prefix idx (toSingleLine expr)
   withMargin' cfg@(extra, margin) ex@EX_APPLICATION{tab = tab@(TAB indt), ..} =
     let single = toSingleLine ex
         main = withMargin' cfg expr
--- a/test/PrinterSpec.hs
+++ b/test/PrinterSpec.hs
@@ -80,6 +80,17 @@ spec = do
             parseExpression printed `shouldBe` Right expr
       )
 
+  describe "printExpression keeps a compressed meet atomic under a narrow margin" $
+    -- A \phinoMeet is a single \overbracket visual unit, so its body must stay
+    -- on one line even when the surrounding margin forces the outer formation to
+    -- wrap. A newline inside the braced argument would raise "! Missing }
+    -- inserted" in an aligned/gathered LaTeX context (see #978).
+    it "renders the meet body on a single line even when the margin wraps its parent" $ do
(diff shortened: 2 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1833](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #995, opened 2026-07-23
Add `--seed` flag for reproducible rewriting

### LINKED ISSUE #994: `phino rewrite` is non-reproducible: `random-string` fresh names drawn from process entropy (src/Random.hs:28)
`phino rewrite` returns different output on every run for identical input and rules. Rules mint fresh synthetic names through the `random-string` built-in (`src/Functions.hs:145`), whose `%d`/`%x` directives expand via `randomRIO` on the global generator (`src/Random.hs:26,28`); that generator is seeded from process entropy and `app/Main.hs:11` never fixes it. So a name template such as `"lambda$hone$%d"` or `"L%d"` gets a different number each process — names stay unique within a run (the `strings` dedup set) but are disjoint across runs.

For a rewriter this breaks reproducible builds: the same `.phi` optimized twice yields byte-distinct output, which defeats caching and makes a result impossible to bisect. It is also the precondition for the intermittent, hard-to-reproduce mis-fusion reported in objectionary/hone-maven-plugin#764 — a `LambdaConversionException` / `NoSuchMethodError` on a synthetic unbox-lambda bridge that surfaces in roughly one build in twenty-five under the amd64 image and never on a native host run. With a random name base that fault is a heisenbug; with a fixed base it becomes deterministic and debuggable.

The smallest fix is to seed the generator deterministically before rewriting — for example `setStdGen (mkStdGen 0)` in `app/Main.hs:11` ahead of `runCLI`, or a seed derived from a hash of the input document so names stay stable per input — and optionally expose a `--seed` flag. The within-run uniqueness in `regenerate` (`src/Random.hs:36-43`) is unaffected.

yegor256:
Closes #994

## Problem

`phino rewrite` (and `dataize`/`match`) returned different output on every run for identical input and rules. Rules mint fresh synthetic names through the `random-string` built-in (`src/Functions.hs`), whose `%d`/`%x` directives expand via `randomRIO` on the global generator (`src/Random.hs`). That generator is seeded from process entropy and was never fixed, so a name template such as `"lambda$hone$%d"` gets a different number each process. Names stayed unique within a run but were disjoint across runs, breaking reproducible builds and turning synthetic-name faults into heisenbugs.

## Change

- Added a `--seed SEED` option (default `0`) to the `rewrite`, `dataize`, and `match` commands.
- Each runner now calls `setStdGen (mkStdGen _seed)` before rewriting, so the global generator is fixed deterministically. Identical inputs now yield byte-identical output across separate process runs; a different seed can be chosen when a distinct name base is needed.

The within-run uniqueness in `regenerate` (`src/Random.hs`) is unaffected — the `strings` dedup set still guarantees no collisions inside a single run.

## Files touched

- `src/CLI/Parsers.hs` — new `optSeed` parser, wired into all three command parsers.
- `src/CLI/Types.hs` — new `_seed :: Int` field on `OptsRewrite`, `OptsDataize`, `OptsMatch`.
- `src/CLI/Runners.hs` — seed the generator before each run.
- `test/CLISpec.hs` — flag-wiring tests (help output, default, acceptance, invalid value).
- `README.md` — documented the flag.

## Tests

Added CLI tests verifying the flag is accepted, appears in `--help` with its default of `0`, and rejects non-integer values, for `rewrite`, `dataize`, and `match`.

---
_Generated by [Claude Code](https://claude.ai/code/session_01P6cuXpWJk3xWGgR2a1bSAA)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,14 @@ to normalize expressions:
 phino rewrite --normalize hello.phi
 ```
 
+Some rules mint fresh synthetic names via the `random-string` built-in. To
+keep the output reproducible across runs, `phino` seeds the random generator
+deterministically with `0` by default. Use `--seed` to pick a different seed:
+
+```bash
+phino rewrite --seed=42 --rule=my-rule.yml hello.phi
+```
+
 If no input file is provided, the 𝜑-expression is taken from `stdin`:
 
 ```bash
--- a/src/CLI/Parsers.hs
+++ b/src/CLI/Parsers.hs
@@ -192,6 +192,17 @@ optStepsDir = optional (strOption (long "steps-dir" <> metavar "FILE" <> help "D
 optShuffle :: Parser Bool
 optShuffle = switch (long "shuffle" <> help "Shuffle rules before applying")
 
+optSeed :: Parser Int
+optSeed =
+  option
+    auto
+    ( long "seed"
+        <> metavar "SEED"
+        <> help "Seed for the random generator that mints fresh synthetic names, making output reproducible across runs"
+        <> value 0
+        <> showDefault
+    )
+
 optSugar :: Parser SugarType
 optSugar = flag SALTY SWEET (long "sweet" <> help (printf "Print result and intermediate (see %s option(s)) 𝜑-expressions using syntax sugar" _intermediateOptions))
 
@@ -259,6 +270,7 @@ dataizeParser =
             <*> optCanonize
             <*> optDepthSensitive
             <*> optShuffle
+            <*> optSeed
             <*> switch (long "quiet" <> help "Don't print the result of dataization")
             <*> optCompress
             <*> optMaxDepth
@@ -290,6 +302,7 @@ rewriteParser =
             <*> optMust
             <*> optNormalize
             <*> optShuffle
+            <*> optSeed
             <*> optOmitListing
             <*> optOmitComments
             <*> optDepthSensitive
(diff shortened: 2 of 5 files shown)


yegor256:
The `rewrite/normalize` regression reported by `regression-check` (+50.6%) is a false positive from runner noise, not an effect of this change.

This PR touches only three files — `src/CLI/Parsers.hs`, `src/CLI/Types.hs`, and `src/CLI/Runners.hs` — all in the CLI layer. The benchmark suite (`benchmark/Main.hs`) calls `rewrite expr normalizationRules rewriteCtx` directly from the `Rewriter` module and imports `Rewriter`/`Parser`/`Printer`/`XMIR`/`Yaml`/`Functions` only. It never imports `CLI.Runners` or invokes `runRewrite`, so the added `setStdGen (mkStdGen _seed)` call is never executed in the benchmark and cannot affect its timing.

The rest of the board confirms it's noise: the whole head run drifted upward — `parse/xmir` +4.6%, `print/sweet/flat` +11.4%, `print/sweet/multiline` +5.0%, `print/salty/multiline` +3.8% — all in code paths with no RNG and untouched by this diff. A uniform upward shift across unrelated benchmarks is the signature of a slow CI runner, not a real +50% regression in one case.

I don't have permission to re-run the job via the API (`403 Resource not accessible by integration`). A re-run of `regression-check` should clear it.

---

maxonfjvipon (approved):

maxonfjvipon:

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+788](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2692](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1000, opened 2026-07-29
Point parse errors at the typo, not at the first binding

### LINKED ISSUE #873: Parser reports syntax errors inside a formation at the first binding, not the offending token
When a syntax error sits inside a formation's binding value, the parser reports it at the formation's *first* binding rather than at the offending token. For example:

```
$ echo '{[[ ex -> [[ x -> ? ]] ( ~1 -> 42 ).x ]]}' | phino rewrite
[ERROR]: Couldn't parse given phi program, cause: program:1:5:
  |
1 | {[[ ex -> [[ x -> ? ]] ( ~1 -> 42 ).x ]]}
  |     ^^
unexpected "ex"
expecting "]]", '⟧', or binding
```

The real problem is the argument `~1` (≈ column 26), which is not a valid attribute or alpha, yet the caret points at `ex` and the message claims `ex` is unexpected even though `ex` is a perfectly valid binding. Replacing `~1` with a valid label such as `x` parses fine, which is why the location is so confusing.

This is the same symptom as #119, and that report's own example still reproduces today even though #119 was closed by #382 and released in `0.0.43`:

```
$ echo '{⟦ a ↦ Φ.foo() ⟧}' | phino rewrite
... program:1:4: unexpected "a " expecting "]]", '⟧', or binding
```

So the earlier fix did not resolve the reported defect — in a large file the caret still lands on the first binding instead of the typo, making it very hard to locate.

The cause is the over-broad backtracking in `binding` at `src/Parser.hs:287`: `try (tauBinding attribute)` wraps the *entire* tau value (`tauValue`, `src/Parser.hs:246`), so a parse failure arbitrarily deep inside the value rewinds all the way to the binding start (`src/Parser.hs:284`) and the furthest-error offset is discarded. The fix is to narrow the `try` so it commits once the `attribute ↦` prefix is matched — parse `attribute` and the arrow once, then branch on `?`/`∅` (void) versus a value expression by lookahead — so a failure inside the value keeps its real offset instead of being rolled back to the binding.

yegor256:
Every alternative of `binding` in `src/Parser.hs` used to be wrapped in `try`, so a syntax error anywhere inside a binding rewound the parser to the start of that binding. Megaparsec then blamed the formation's *first* binding, however valid it was, and the caret never landed near the actual typo. Each alternative now commits as soon as the token that tells it apart from its siblings is consumed, and the same narrowing is applied to `tauValue` and `argument`.

This matters most in big files, where `unexpected "a "` pointing at line 1 tells you nothing about a typo 3000 lines down. Two examples: `⟦ a ↦ Φ.foo() ⟧` now reports column 13 (the stray `()`) instead of column 3, and `[[ D> x ]]` now says `expecting bytes` at column 7 instead of `unexpected "D>"` — the delta branch was the only one that swallowed its own inner error, while the lambda branch had always reported correctly. `function` also got a label, so a bad lambda binding says `expecting function name`.

Six new cases in `ParserSpec` pin the reported positions and three more pin the expected-token sets; five of them fail on `master`. As a side effect the tau path no longer saves parser state for the whole value, so parsing a 535 KB formation is a hair faster rather than slower.

Closes #873

### DIFF
--- a/src/Parser.hs
+++ b/src/Parser.hs
@@ -82,15 +82,19 @@ label' = lexeme $ do
   return (T.pack (first : rest))
 
 function :: Parser String
-function = lexeme $ do
-  first <- oneOf ['A' .. 'Z']
-  rest <-
-    many
-      ( satisfy
-          (\ch -> isDigit ch || isAsciiLower ch || ch == '_' || ch == 'φ')
-          <?> "allowed character in function name"
-      )
-  return (first : rest)
+function =
+  lexeme
+    ( do
+        first <- oneOf ['A' .. 'Z']
+        rest <-
+          many
+            ( satisfy
+                (\ch -> isDigit ch || isAsciiLower ch || ch == '_' || ch == 'φ')
+                <?> "allowed character in function name"
+            )
+        return (first : rest)
+    )
+    <?> "function name"
 
 delta :: Parser String
 delta =
@@ -244,7 +248,7 @@ quotedStr = char '"' >> manyTill (choice [escapedChar, noneOf ['\\', '"']]) (cha
 tauValue :: Parser Expression
 tauValue =
   choice
-    [ try $ do
+    [ do
         _ <- arrow
         expression
     , do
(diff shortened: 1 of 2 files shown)


yegor256:
@maxonfjvipon good?

maxonfjvipon (approved):

maxonfjvipon:
@yegor256 merge manually

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (102 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2857](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+722](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1001, opened 2026-07-29
Need for `--hide-rho` option eliminating ρ bindings in printed expressions for cleaner output


yegor256:
ρ bindings clutter every printed expression. For the input
`[[ foo -> [[ x -> [[ ]], ^ -> $.y ]], y -> [[ ]] ]]`, `phino rewrite` prints
`⟦ foo ↦ ⟦ x ↦ ⟦ ρ ↦ ∅ ⟧, ρ ↦ ξ.y ⟧, y ↦ ⟦ ρ ↦ ∅ ⟧, ρ ↦ ∅ ⟧`, and `--sweet` only
drops the trailing void ρ (`withoutLastVoidRho`, `src/CST.hs:339`), so
`ρ ↦ ξ.y` survives. A new `--hide-rho` switch would remove every ρ binding from
the output, leaving `⟦ foo ↦ ⟦ x ↦ ⟦ ⟧ ⟧, y ↦ ⟦ ⟧ ⟧`, which is much easier to
read when following a long `--sequence` chain.

The change is cosmetic and belongs to the printing path alone: ρ must stay in the
AST while rewriting and matching, and `withVoidRho` (`src/Misc.hs:78`) keeps that
invariant. Dispatches such as `ξ.ρ` stay untouched too, since removing them would
break the expression rather than shorten it.

The fix: add `optHideRho` next to `optSugar` (`src/CLI/Parsers.hs:207`) with a
`_hideRho` field in `PrintContext` (`src/CLI/Types.hs:18`), and drop every
`BiVoid AtRho` and `BiTau AtRho _` binding recursively in `printExpression` and
`printInFormat` (`src/CLI/Helpers.hs:78` and `:85`) before rendering. The SALTY
path needs care: `bdWithVoidRho` (`src/Sugar.hs:24`) re-inserts `ρ ↦ ∅` into
every formation, so the flag must suppress that as well, otherwise the stripping
is silently undone.

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+2829](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1008, opened 2026-07-29
fix(sugar): strip ρ application arguments under --hide-rho (#1007)

### LINKED ISSUE #1007: `--hide-rho` flag fails to suppress application arguments in print outputs
The `--hide-rho` switch added in #1001 strips ρ from formation bindings only:
`goArgument` inside `withoutRho` deliberately leaves application arguments alone
(`src/Sugar.hs:80-95`). So an application such as `e(ρ ↦ ⟦ … ⟧)` — exactly what
the DOT rule emits on every step, see `resources/normalize/dot.yaml:15` — still
prints its `(ρ ↦ …)` argument with the flag on, and the noise the flag exists to
remove is still there.

With `--hide-rho` the ρ argument should disappear too, and when it was the only
argument the whole application should collapse to plain `e`, with no leftover
empty parens `e()`. An application that carries other arguments keeps them and
loses only ρ, so `e(α0 ↦ x, ρ ↦ y)` prints as `e(α0 ↦ x)`.

The fix: make `goArgument` and `goArgBindings` (`src/Sugar.hs:82` onwards) skip ρ
pairs through the existing `isRho` predicate (`src/Sugar.hs:106`), the way
`goBindings` already does for formations, and let the `EX_APPLICATION` case of
`goExpr` (`src/Sugar.hs:57`) return `goExpr expr` when the argument ends up
empty. Positional arguments (`AA_EXPRS`) need no change, since ρ cannot appear
there. One more case in the "printExpressionHidingRho strips every rho binding"
group (`test/PrinterSpec.hs:125`) would cover it.

yegor256:
Fixes #1007.

## Problem

The `--hide-rho` switch (added in #1001) stripped ρ from **formation bindings only**. `goArgument` inside `withoutRho` deliberately left application arguments alone, so an application such as `e(ρ ↦ ⟦ … ⟧)` — exactly what the DOT rule emits on every step (`resources/normalize/dot.yaml:15`) — still printed its `(ρ ↦ …)` argument with the flag on. The noise the flag exists to remove was still there.

## Fix

`src/Sugar.hs` — `goArgument`, `goArgBinding` and `goArgBindings` now skip ρ pairs through the existing `isRho` predicate, the same way `goBinding`/`goBindings` already do for formations:

- `goArgument` returns `Maybe APP_ARGUMENT`. When the strip leaves nothing (`AA_TAU` whose sole pair was ρ, or `AA_TAUS` reduced to `BI_EMPTY`), it returns `Nothing`.
- The `EX_APPLICATION` case of `goExpr` returns `goExpr expr` (the bare callee) when the argument is `Nothing`, so an application whose only argument was ρ collapses to plain `e` with **no leftover `e()`**.
- An application carrying other arguments keeps them and loses only ρ: `e(α0 ↦ x, ρ ↦ y)` now prints as `e(α0 ↦ x)`.
- Positional arguments (`AA_EXPRS`) are untouched, since ρ cannot appear there.

The module-level doc comment on `withoutRho` is updated to reflect that application arguments are now stripped too.

## Tests

Added three cases to the `printExpressionHidingRho strips every rho binding for --hide-rho` group in `test/PrinterSpec.hs`:

- an application whose only argument is ρ collapses to its callee (SWEET)
- the same collapse on the SALTY path, with no leftover parens
- an application keeps its other arguments and drops only ρ

---
_Generated by [Claude Code](https://claude.ai/code/session_01BkXA4SQyLzU9HwfpmJxRUB)_

### DIFF
--- a/src/Sugar.hs
+++ b/src/Sugar.hs
@@ -43,9 +43,10 @@ data SugarType = SWEET | SALTY
 -- Drop every ρ binding (ρ ↦ ∅, ρ ↦ e and ρ(…) ↦ e) from a rendered CST, the
 -- effect of the '--hide-rho' switch. It runs after 'withSugarType', so it also
 -- removes the ρ ↦ ∅ that 'bdWithVoidRho' re-inserts into every formation on the
--- SALTY path. Only formation bindings are stripped: dispatches such as ξ.ρ and
--- application arguments are left untouched, and a formation left empty by the
--- strip collapses to the compact '⟦⟧' layout.
+-- SALTY path. Both formation bindings and application arguments are stripped;
+-- dispatches such as ξ.ρ are left untouched. A formation left empty by the
+-- strip collapses to the compact '⟦⟧' layout, and an application left with no
+-- argument collapses to its bare callee (no leftover 'e()').
 withoutRho :: EXPRESSION -> EXPRESSION
 withoutRho = goExpr
   where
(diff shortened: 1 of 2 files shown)


yegor256:
@maxonfjvipon good?

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2897](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1010, opened 2026-07-29
fix(rule): treat ⊥ as xi-free so copy accepts a ⊥ argument (#1009)

### LINKED ISSUE #1009: copy rule refuses a ⊥ argument, so ⟦x ↦ ∅⟧(x ↦ ⊥) is a stuck normal form (src/Rule.hs:196)
The `copy` rule (`resources/normalize/copy.yaml`) binds the applied argument to the absolute meta-variable `𝑘`, and `_absolute` at `src/Rule.hs:188` gates every `𝑘` on its local `xiFree` helper, whose catch-all at `src/Rule.hs:196` returns `False` for `ExTermination`. So a void slot receiving `⊥` matches no normalization rule at all: `echo '⟦ r ↦ ∅ ⟧( r ↦ ⟦ x ↦ ∅ ⟧( x ↦ ⊥ ) )' | phino rewrite --normalize` prints the input back (with `ρ ↦ ∅` added) instead of `⟦ r ↦ ⟦ x ↦ ⊥ ⟧ ⟧`. The minimal case is `⟦ x ↦ ∅ ⟧( x ↦ ⊥ )`, which a DEBUG trace shows `copy` declining.

This is wrong because `⊥` holds no `ξ` to capture, which is the only thing absoluteness guards against, and `isNF ExTermination = True` at `src/Rule.hs:55` already treats `⊥` as a normal form. The `𝒜` enumeration adopted in #785 — `Φ`, formation, dispatch with absolute subject, application with absolute subject and argument — just never mentions `⊥`, and no clause can reach it. Two things follow: normalization terminates with an unreduced redex in place, and since `⊥` counts as non-absolute, morphing takes the `mad` branch added by #959 (`resources/morphing.yaml:131`) and collapses the whole object to `⊥`, so dataization discards the object instead of filling its slot.

Add `xiFree ExTermination = True` beside `xiFree ExRoot = True` at `src/Rule.hs:192`; with that single line the example above reduces to `⟦ r ↦ ⟦ x ↦ ⊥, ρ ↦ ∅ ⟧, ρ ↦ ∅ ⟧`, as expected. If `⊥` must stay outside `𝒜` by definition, then `copy` needs a companion axiom covering a `⊥` argument instead.

Reproduced with 0.0.107 and at `eeabf9c`.

yegor256:
Fixes #1009.

## Problem

The `copy` rule (`resources/normalize/copy.yaml`) binds the applied argument to the absolute meta-variable `𝑘`. The `_absolute` check in `src/Rule.hs` gates `𝑘` on its local `xiFree` helper, whose catch-all returned `False` for `ExTermination`. As a result a void slot receiving `⊥` matched **no** normalization rule and became a stuck normal form:

```
echo '⟦ r ↦ ∅ ⟧( r ↦ ⟦ x ↦ ∅ ⟧( x ↦ ⊥ ) )' | phino rewrite --normalize
```

printed the input back (with `ρ ↦ ∅` added) instead of reducing. Because `⊥` also counted as non-absolute, morphing then took the `mad` branch (`resources/morphing.yaml`) and collapsed the whole object to `⊥`, so dataization discarded the object instead of filling its slot.

## Fix

`⊥` holds no `ξ` to capture, which is the only thing absoluteness guards against, and `isNF ExTermination = True` already treats `⊥` as a normal form. Added `xiFree ExTermination = True` beside `xiFree ExRoot = True` so `⊥` is recognized as absolute. With that line the example above reduces to `⟦ r ↦ ⟦ x ↦ ⊥, ρ ↦ ∅ ⟧, ρ ↦ ∅ ⟧`, as expected.

## Changes

- `src/Rule.hs` — `xiFree ExTermination = True` in `_absolute` (plus updated comment).
- `test-resources/rewriter-packs/normalize/e-cpy-bot.yaml` — normalization pack for the issue's example.
- `test-resources/condition-packs/absolute-termination.yaml` — condition pack asserting `⊥` is absolute.

https://claude.ai/code/session_01EWiX1SCxXwhEJPxFFmEVwB

---
_Generated by [Claude Code](https://claude.ai/code/session_01EWiX1SCxXwhEJPxFFmEVwB)_

### DIFF
--- a/src/Rule.hs
+++ b/src/Rule.hs
@@ -176,11 +176,13 @@ _nf (ExMeta meta) (Subst mp) ctx = case M.lookup meta mp of
 _nf expr subst ctx = pure [subst | isNF expr ctx]
 
 -- An expression is xi-free when it contains no ξ outside of a formation: it is
--- Φ, a formation, a dispatch with a xi-free subject, or an application with a
--- xi-free subject and argument. Together with a normal-form check this is what
--- makes an expression absolute (𝒦 ⊆ 𝒩); the '𝑘' meta-variable applies this
--- xi-free check first (cheap, structural, rules out the ξ-recursion the
--- normal-form check could loop on) and the normal-form check second.
+-- Φ, ⊥, a formation, a dispatch with a xi-free subject, or an application with
+-- a xi-free subject and argument. ⊥ holds no ξ to capture, so it is xi-free
+-- (and 'isNF ⊥ = True' already), which lets the copy rule accept a ⊥ argument.
+-- Together with a normal-form check this is what makes an expression absolute
+-- (𝒦 ⊆ 𝒩); the '𝑘' meta-variable applies this xi-free check first (cheap,
+-- structural, rules out the ξ-recursion the normal-form check could loop on)
+-- and the normal-form check second.
 _absolute :: Expression -> Subst -> RuleContext -> IO [Subst]
 _absolute (ExMeta meta) (Subst mp) ctx = case M.lookup meta mp of
   Just (MvExpression expr) -> _absolute expr (Subst mp) ctx
@@ -190,6 +192,7 @@ _absolute expr subst _ = pure [subst | xiFree expr]
     xiFree :: Expression -> Bool
     xiFree (ExFormation _) = True
     xiFree ExRoot = True
+    xiFree ExTermination = True
     xiFree (ExApplication e (ArTau _ te)) = xiFree e && xiFree te
     xiFree (ExApplication e (ArAlpha _ te)) = xiFree e && xiFree te
     xiFree (ExDispatch e _) = xiFree e
--- a/test-resources/condition-packs/absolute-termination.yaml
+++ b/test-resources/condition-packs/absolute-termination.yaml
@@ -1,0 +1,7 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+expression: '[[ x -> T ]]'
+pattern: '[[ x -> !e ]]'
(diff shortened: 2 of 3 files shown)


yegor256:
The `regression-check` failure is benchmark noise, not a real regression from this change.

The only code change here is one O(1) constructor case added to the pure structural `xiFree` helper in `_absolute`:

```haskell
xiFree ExTermination = True
```

It sits before the existing catch-all and adds no cost to any existing path, so it cannot slow `rewrite/normalize` by 39%. The reported numbers are inside the measurement noise band:

- `rewrite/normalize` head run: avg 87362 μs, **std dev 18268 μs (~21%)**, min 65270, max 132046 — nearly a 2× spread within the single sampled run.
- The unrelated `print/sweet/flat` (+3.3%) and `print/sweet/multiline` (+3.1%) drifted up in the same run, indicating the head runner was simply slower this time.

The gate is a single-sample A/B with a +20% threshold, and `rewrite/normalize` is the smallest/fastest case (~63–87 ms), so it's the most susceptible to shared-runner jitter.

I don't have permission to re-run jobs on this repo (`rerun-failed-jobs` returns 403), so a re-run of the `regression-check` job would confirm this. Happy to do so if given access.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (28) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2901](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1014, opened 2026-08-05
Report PR benchmark deltas instead of gating on them

### LINKED ISSUE #1005: Pull Request Benchmark Yields Unreliable, Non-informative Outcome
Sometime ago we've added benchmark which is triggered on each pull request and shows if new changes bring performance boost or degradation. 
After a while it becomes clear that this benchmark does not show a real picture, it's rather noise which makes build red for no reason.
We either should figure out how to make it more adequate, or drop it

maxonfjvipon:
The benchmark gate measured one sample per side and compared it against a flat
+20% threshold. That threshold is below the noise floor of a shared runner. On
master alone, `rewrite/normalize` reported anywhere from 50k to 106k μs and
`parse/phi` from 126k to 179k μs across recent runs, so PRs failed on jitter:
#1010 was flagged at +39% and #1000 at +22.5%, and a renovate PR bumping an
action version got a red build too.

So the check keeps measuring, but it stops gating. Both bench binaries are
built up front and then run interleaved, alternating which side goes first so
that load drifting during the job cannot systematically favour one side:

```bash
for round in $(seq 1 "$BENCH_ROUNDS"); do
  if [ $((round % 2)) -eq 0 ]; then
    /tmp/bench-head >> /tmp/head.out
    /tmp/bench-base >> /tmp/base.out
  else
    /tmp/bench-base >> /tmp/base.out
    /tmp/bench-head >> /tmp/head.out
  fi
done
```

`bench-compare.sh` takes the median of the five rounds per case instead of a
lone sample, and prints the spread between the fastest and slowest round as a
noise column beside every delta. A case is called out only when it moves past
both the threshold and its own noise. Replaying the #1010 numbers through it,
the +39% "REGRESSION" becomes +0.7% against ±92.2% noise; a synthetic 2x
slowdown with tight rounds still reads +100.2% against ±2.1% and gets called
out.

Verified against real bench output too: three interleaved A/A rounds of the
same binary, where every delta is by definition noise, produce no callouts.

The job name is still `regression-check` even though nothing regressed can fail
it any more. Renaming it touches `post-comment.yml` and the check name people
see on PRs, so I left it alone; say the word and it becomes `benchmark-report`.

The second commit is unrelated and small: `renderFunc`, `nfMetaNames` and
`kMetaNames` each had a single call site, so they move into the `where` scope of
the equation that uses them.

Closes #1005

### DIFF
--- a/.github/scripts/bench-compare.sh
+++ b/.github/scripts/bench-compare.sh
@@ -5,17 +5,21 @@ set -euo pipefail
 
 usage() {
   cat <<'EOF'
-Compare two benchmark runs and gate on regressions.
+Compare two benchmark runs and report the deltas.
 
 Usage: bench-compare.sh BASE.txt HEAD.txt THRESHOLD
 
-BASE/HEAD: stdout captures of `cabal bench` (the bench prints `=== name ===`
-blocks with an `avg: X.YYY us` line per case).
+BASE/HEAD: stdout captures of the bench binary (it prints `=== name ===`
+blocks with an `avg: X.YYY us` line per case). Either file may hold several
+rounds appended one after another; every round counts as one sample, and the
+median of the samples is what gets reported.
 
-THRESHOLD: max allowed positive delta as a fraction, e.g. 0.20 for +20%.
+THRESHOLD: how far a case has to move to be called out, as a fraction, e.g.
+0.20 for 20%. A case is called out only when it also moves further than the
+noise measured across its own rounds.
 
-Prints a markdown comparison table to stdout, exits 1 if any case present in
-both runs regressed by more than THRESHOLD.
+Prints a markdown comparison table to stdout. A delta never fails the run:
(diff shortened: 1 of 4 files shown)


maxonfjvipon:
@yegor256 this makes the PR benchmark informational rather than a gate, since the +20% threshold sat below the runner noise floor; happy to explain any part of the median/noise logic or to rename the job if you prefer.

maxonfjvipon:
@rultor merge

0crat:
@maxonfjvipon Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (161 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+706](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1025, opened 2026-08-07
fix(cli): seed rule shuffle order with --seed (#1016)

### LINKED ISSUE #1016: --shuffle --seed=N is non-reproducible for the rule order (setStdGen runs after getRules)
`phino rewrite --shuffle --seed=N` is non-reproducible for the **rule order**: the rules are shuffled in `getRules` before `setStdGen` runs, so a fixed seed does not fix the order in which rules are tried.

## Problem

`src/CLI/Runners.hs:48` calls `getRules _normalize _shuffle _rules`, and `getRules` shuffles the rules via `Random.newStdGen` (see `src/Random.hs` `shuffle`). Then `setStdGen (mkStdGen _seed)` runs only at `src/CLI/Runners.hs:52`. Because the shuffle happens on the *unseeded* global generator, the resulting order of rules differs on every run:

```
echo '⟦ x ↦ ∅ ⟧' | phino rewrite --normalize --shuffle --seed=42
```

produces different intermediate output across separate runs for the same seed, so `--seed` does not make the output byte-identical as documented.

The `--seed` flag (added in #995, closing #994) fixed the determinism of synthetic **names** minted by `random-string`/`random-tau`. It did not fix the determinism of the **shuffle order**: `getRules` still consumes the generator before `setStdGen` has set it.

## Expected

`--shuffle` + a fixed `--seed` should yield the same rule ordering (and thus the same rewriting) on every run. Passing `--shuffle` without `--seed` may stay non-deterministic.

## Suggested fix

- Move `setStdGen (mkStdGen _seed)` to the start of the runner, before `getRules`; or
- Thread the seed explicitly into `getRules`/`shuffle` (a `shuffleWith :: StdGen -> [a] -> [a]`), avoiding dependence on the process-global generator.

The source `src/Random.hs` uses `newIOGenM =<< newStdGen` + `uniformRM` in `shuffle`, so it is driven by the global generator set via `setStdGen`. This is a bug report; the fix would land in a follow-up PR.

VasilevNStas:
Fixes #1016.

## Problem

`--shuffle --seed=N` did not reproduce the rule order. In `src/CLI/Runners.hs`, `getRules ... shuffle` ran (line 48) BEFORE `setStdGen (mkStdGen _seed)` (line 52). `Random.shuffle` (src/Random.hs:62) builds its generator with `newStdGen`, which reads/advances the GLOBAL generator, so the shuffle order depended on whatever was left in the global state — not on `--seed`. Three runs with the same `--seed=5` produced three different rule orders.

## Solution

Move `setStdGen (mkStdGen _seed)` before `getRules` in `runRewrite`, so the global generator is seeded before the shuffle happens.

## Test

Added a regression test in `test/CLISpec.hs`: runs `rewrite --normalize --shuffle --seed=42` twice and asserts identical output.

## Checklist

- [x] `make hlint` — no hints
- [x] `make fourmolu` — clean

### DIFF
--- a/src/CLI/Runners.hs
+++ b/src/CLI/Runners.hs
@@ -45,11 +45,11 @@ runRewrite OptsRewrite{..} = do
   included <- validatedDispatches "show" _show
   [loc] <- validatedDispatches "locator" [_locator]
   [foc] <- validatedDispatches "focus" [_focus]
+  setStdGen (mkStdGen _seed)
   rules <- getRules _normalize _shuffle _rules
   validateBreakpoint _breakpoint rules
   input <- readInput _inputFile
   expr <- parseInput input _inputFormat
-  setStdGen (mkStdGen _seed)
   seedTaus expr
   logDebug (printf "Amount of rewriting cycles across all the rules: %d, per rule: %d" _maxCycles _maxDepth)
   let listing = case (rules, _inputFormat, _outputFormat) of
--- a/test-resources/cli/swap-a.yaml
+++ b/test-resources/cli/swap-a.yaml
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2026 Objectionary.com
+# SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
 # SPDX-License-Identifier: MIT
 ---
 name: swap-a
--- a/test-resources/cli/swap-a.yaml
+++ b/test-resources/cli/swap-a.yaml
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+name: swap-a
+pattern: ⟦ x -> !e ⟧
+result: Q.a(!e)
--- a/test-resources/cli/swap-b.yaml
+++ b/test-resources/cli/swap-b.yaml
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+name: swap-b
+pattern: ⟦ x -> !e ⟧
+result: Q.b(!e)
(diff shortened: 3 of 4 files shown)


VasilevNStas:
@maxonfjvipon @yegor256 
please review

yegor256:
The fix in `src/CLI/Runners.hs` is right, but the regression test does not reproduce the bug.

I applied only the test hunk on top of unfixed `master` and ran it:

```
CLI
  rewriting
    reproduces the same shuffle order for the same --seed [✔]

Finished in 0.0036 seconds
1 example, 0 failures
```

It passes without the fix. The reason: for `[[ x -> Q.y ]]` under `--normalize`, the final expression is the same whatever order the rules are tried in, and without `--sequence` only that final expression is printed. Five separate runs of the unfixed binary agree:

```
$ for i in 1 2 3 4 5; do echo "[[ x -> Q.y ]]" | phino rewrite --normalize --shuffle --seed=42 --sweet | md5; done
ea082c32b498c7feacd0bd2a6d799c3d
ea082c32b498c7feacd0bd2a6d799c3d
ea082c32b498c7feacd0bd2a6d799c3d
ea082c32b498c7feacd0bd2a6d799c3d
ea082c32b498c7feacd0bd2a6d799c3d
```

So the test guards nothing and would not catch a regression of #1016.

Please make it observe the rule order. Two options:

- assert on `--sequence` output (the intermediate chain does depend on which rule fires first), for an input where the normalization path is genuinely order-sensitive; or
- test `Random.shuffle` directly: `setStdGen (mkStdGen 42)` then shuffle a list, twice, and assert the two orders are equal — that pins the seeding contract without depending on rewriting being order-sensitive.

Either way, confirm the new test fails on `master` before the one-line move in `Runners.hs`.

---

VasilevNStas:
@yegor256 
plz review this

yegor256:
Thanks for the contribution!

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (30) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+502](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1026, opened 2026-08-07
fix(regex): terminate replaceAll on zero-length matches (#1020)

### LINKED ISSUE #1020: Regexp.replaceAll loops forever on an empty match
`Regexp.replaceAll` never terminates when a pattern can match the empty string, so a rule like `--sed s//x/g` (or any zero-length match) hangs rewriting.

## Problem

`src/Regexp.hs:77` `replaceAll` walks the input:

```haskell
go bs acc = do
  result <- R.execute regex bs
  case result of
    Right (Just arr) -> do
      let (off, len) = arr ! 0                       -- len can be 0
          (before, rest1) = B.splitAt off bs
          (_, rest2) = B.splitAt len rest1           -- len==0 -> rest2 == rest1
      ...
      go rest2 (B.concat [acc, before, replacement])
```

When `len == 0` (an empty pattern, or an alternation that matches empty at the cursor), `rest2` equals `rest1`, so the recursion does not advance and `go` loops forever. The same pattern exists in the intermediate `replaceAll` usage in the `sed` where-function with `s//pat//`.

## Expected

A zero-length match must advance the cursor by at least one byte (or reject/ignore empty matches) so `replaceAll` always terminates. A regression test with an empty (`s//x/g`) or empty-matching pattern should complete (or error), not hang.

## Suggested fix

When `len == 0`, fall back to copying one byte and recursing on the tail (or skip a single match). The fix is small and low-risk.

VasilevNStas:
Fixes #1020.

## Problem

`Regexp.replaceAll` looped forever when the regex produced a zero-length match (`len == 0`), e.g. as `sed` argument `"s/^/X/g"`. In `src/Regexp.hs` the old body recursed as:

```haskell
go rest2 (B.concat [acc, before, replacement])
```

With `len == 0`, `B.splitAt len rest1` yields `rest2 == rest1 == bs` (the match consumed nothing), so `go` was called on the *same* input again while `acc` grew — an unbounded loop (observed as the CLI hanging / killed by timeout, exit=124).

## Solution

When a match has zero length, consume exactly one byte alongside the replacement and continue from there (a standard treatment for empty matches):

```haskell
if len == 0
  then
    let next = B.take 1 rest2
     in if B.null next
          then return $ B.concat [acc, before, replacement]
          else go (B.drop 1 rest2) (B.concat [acc, before, replacement, next])
  else go rest2 (B.concat [acc, before, replacement])
```

This preserves progress: every iteration consumes at least one byte, so the recursion always terminates. Non-empty matches behave exactly as before.

## Note on semantics

With an anchored `^` the result is now `s/^/X/g` on `"hello"` → `"XhXeXlXlXoX"` (a replacement at every position, since `replaceAll` re-runs the regex on the remaining tail). The loop is gone and the output is deterministic; if the single-anchor behavior is desired instead, a dedicated change (keeping the original offset) would be a separate follow-up.

## Test

Added two `RegexpSpec` cases: `^` on `"hello"` and an empty pattern — both assert the fixed (deterministic, terminating) output. `make test` passes (1124 examples).

## Checklist

- [x] `make hlint` — no hints
- [x] `make fourmolu` — clean

### DIFF
--- a/src/Regexp.hs
+++ b/src/Regexp.hs
@@ -89,4 +89,10 @@ replaceAll regex rep input = go input B.empty
               (_, rest2) = B.splitAt len rest1
           groups <- extractGroups regex bs
           let replacement = substituteGroups rep groups
-          go rest2 (B.concat [acc, before, replacement])
+          if len == 0
+            then
+              let next = B.take 1 rest2
+               in if B.null next
+                    then return $ B.concat [acc, before, replacement]
+                    else go (B.drop 1 rest2) (B.concat [acc, before, replacement, next])
+            else go rest2 (B.concat [acc, before, replacement])
--- a/test/RegexpSpec.hs
+++ b/test/RegexpSpec.hs
@@ -248,3 +248,13 @@ spec = do
       regex <- R.compile (B.pack "\\bword\\b")
       result <- R.replaceAll regex (B.pack "WORD") (B.pack "word in a word")
       result `shouldBe` B.pack "WORD in a WORD"
+
+    it "terminates on an empty-match pattern (anchored ^)" $ do
+      regex <- R.compile (B.pack "^")
+      result <- R.replaceAll regex (B.pack "X") (B.pack "hello")
+      result `shouldBe` B.pack "XhXeXlXlXoX"
+
+    it "terminates on an empty regex pattern" $ do
+      regex <- R.compile B.empty
+      result <- R.replaceAll regex (B.pack "X") (B.pack "hello")
+      result `shouldBe` B.pack "XhXeXlXlXoX"


VasilevNStas:
@maxonfjvipon @yegor256 
please review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (18) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+442](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1027, opened 2026-08-08
fix(functions): correct join error message to say join() (#1017)

### LINKED ISSUE #1017: join function reports errors as 'go instead of join
The `join` where-function reports errors under the wrong name: the message reads `Function 'go can work with bindings only`, where `go` is an internal (where-bound) worker, not the public function name. Users calling `join` see a misleading error.

## Problem

In `src/Functions.hs` the `_join` implementation delegates to a local `buildBindings`, whose catch-all throws:

```
throwIO (userError "Function 'go can work with bindings only")
```

Two defects:
- the name `go` (an internal worker) is leaked into a user-facing error, while the function the user actually called, `join`, is lost;
- an extra apostrophe makes the name read as `'go`.

So a rule that calls `join()` with an argument that is not a meta binding gets a confusing message instead of `Function join() can work with bindings only`.

## Expected

The message should name the public function: `Function join() can work with bindings only`. This is a trivial, low-risk fix and a good candidate for a first-time contribution.

VasilevNStas:
Fixes #1017.

## Problem

In `src/Functions.hs` the `join` where-function reported a wrong error message when given a non-binding argument:

```haskell
buildBindings _ = throwIO (userError "Function 'go can work with bindings only")
```

The message said `'go` (an internal worker name) instead of `join`, and the `()` suffix that every other `Function <name>() ...` message in the file uses was missing.

## Solution

```haskell
buildBindings _ = throwIO (userError "Function join() can work with bindings only")
```

Now consistent with the rest of `Functions.hs` (e.g. `Function sed() requires...`, `Function number() requires...`).

## Test

Added `test-resources/cli/join-broken.yaml` (a rule passing a non-binding arg to `join`) and a `CLISpec` case asserting the message contains `Function join() can work with bindings only`.

## Checklist

- [x] `make hlint` — no hints
- [x] `make fourmolu` — clean

### DIFF
--- a/src/Functions.hs
+++ b/src/Functions.hs
@@ -208,7 +208,7 @@ _join args subst = do
       bds <- buildBindingThrows bd subst
       next <- buildBindings args'
       pure (bds ++ next)
-    buildBindings _ = throwIO (userError "Function 'go can work with bindings only")
+    buildBindings _ = throwIO (userError "Function join() can work with bindings only")
     go :: [Binding] -> Set.Set Attribute -> IO [Binding]
     go [] _ = pure []
     go (bd : bds) attrs =
--- a/test-resources/cli/join-broken.yaml
+++ b/test-resources/cli/join-broken.yaml
@@ -1,0 +1,14 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+# This rule passes a non-binding argument to the 'join' where-function.
+# It is used to test that the error message names the 'join' function
+# correctly (it used to say "'go" instead of "join()").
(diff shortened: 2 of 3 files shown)


VasilevNStas:
@maxonfjvipon 
plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+454](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1028, opened 2026-08-08
fix(must): make Show/Read round-trip for the empty range (#1022)

### LINKED ISSUE #1022: Must Show/Read round-trip yields a string, read rejects
The `Show`/`Read` round-trip of `Must` is broken for a range with both bounds empty: `show (MtRange Nothing Nothing)` yields `".."`, but `readsPrec ".."` produces no parse, so the value cannot be read back.

## Problem

`src/Must.hs:19`:

```haskell
show (MtRange Nothing Nothing) = ".."
```

and at `src/Must.hs:26` `readsPrec` routes `".."` into `parseRange`, whose branch for both-empty ends returns `[]`:

```haskell
(Nothing, Nothing, True, True) -> []   -- Invalid range: empty range '..'
```

So `read (show m)` is not `Just m` for `MtRange Nothing Nothing`. While `--must ..` may never be a useful CLI input, the Show/Read instance is inconsistent: one side produces a string the other side rejects.

## Expected

Either:
- make `readsPrec ".."` produce `MtRange Nothing Nothing`, matching `show`; or
- drop/guard the `MtRange Nothing Nothing` case from `show` so the representation is rejectable in one direction.

Minor. The main ask is consistency of the Show/Read instance (explicit lower/upper bounds already round-trip correctly).

VasilevNStas:
Fixes #1022.

## Problem

`Show`/`Read` round-trip for `Must` was broken for the empty range:

```haskell
show (MtRange Nothing Nothing) = ".."   -- src/Must.hs:19
```

but `readsPrec ".."` routed into `parseRange` whose both-empty branch returned `[]`:

```haskell
(Nothing, Nothing, True, True) -> []   -- Invalid range: empty range '..'
```

So `read (show m)` was not `Just m` for `MtRange Nothing Nothing`, even though every other `Must` value round-trips.

## Solution

Make `readsPrec ".."` produce `MtRange Nothing Nothing`, matching `show`:

```haskell
(Nothing, Nothing, True, True) -> [(MtRange Nothing Nothing, "")] -- Empty range '..' round-trips
```

`MtRange Nothing Nothing` is a valid value (an unbounded range, `inRange` returns `True` for any count), so parsing it back is safe.

## Test

Updated `test/MustSpec.hs`: the case previously asserting `readMaybe ".." == Nothing` now asserts the round-trip `Just (MtRange Nothing Nothing)`. `make test` passes (1124 examples).

## Checklist

- [x] `make hlint` — no hints
- [x] `make fourmolu` — clean

### DIFF
--- a/src/Must.hs
+++ b/src/Must.hs
@@ -34,7 +34,7 @@ instance Read Must where
               hiPart = if null hiStr then Nothing else readMaybe hiStr
            in case (loPart, hiPart, null loStr, null hiStr) of
                 (Nothing, Nothing, False, False) -> [] -- Invalid range: non-numeric values
-                (Nothing, Nothing, True, True) -> [] -- Invalid range: empty range '..'
+                (Nothing, Nothing, True, True) -> [(MtRange Nothing Nothing, "")] -- Empty range '..' round-trips
                 (Nothing, Just hi, True, False) ->
                   [(MtRange Nothing (Just hi), "") | hi >= 0]
                 (Just lo, Nothing, False, True) ->
--- a/test/MustSpec.hs
+++ b/test/MustSpec.hs
@@ -121,9 +121,9 @@ spec = do
           it desc $ (readMaybe input :: Maybe Must) `shouldBe` expected
       )
 
-  describe "Read instance rejects empty range" $
-    it "fails on dots only" $
-      (readMaybe ".." :: Maybe Must) `shouldBe` Nothing
+  describe "Read instance parses empty range" $
+    it "round-trips dots only" $
+      (readMaybe ".." :: Maybe Must) `shouldBe` Just (MtRange Nothing Nothing)
 
   describe "Read instance rejects invalid range with negative minimum" $
     it "fails on negative min" $


VasilevNStas:
@yegor256 @maxonfjvipon 
please review this PR

maxonfjvipon:
@VasilevNStas you took the issue which was not labeled as bug so even if PR is merged, you won't get a reward
To confirm this is a real bug - provide a test case which would fail on production but does not after your fix

VasilevNStas:
@maxonfjvipon here's the failing test you asked for.

The new test case in this PR is `test/MustSpec.hs` → `Read instance parses empty range / round-trips dots only`:

```haskell
(readMaybe .. :: Maybe Must) `shouldBe` Just (MtRange Nothing Nothing)
```

**Without the fix** (i.e. on `master`, where `readsPrec ..` routes to the `(Nothing, Nothing, True, True) -> []` branch), this test FAILS:

```
test/MustSpec.hs:126:38:
1) Must, Read instance parses empty range, round-trips dots only
       expected: Just ..
        but got: Nothing
```

**With the fix** it passes:

```
round-trips dots only [✔]
```

So it's a genuine round-trip bug: `show (MtRange Nothing Nothing) == ..` but `read ..` returned `Nothing` before the change. The fix makes `read . show == id` hold for this value, consistent with every other `Must` constructor.

Note: our account can't add labels to issues (no triage rights on this repo), so `#1022` is unlabeled. If you consider it a real bug, could a maintainer add the `bug` label to `#1022`?

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+466](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1029, opened 2026-08-08
fix(dataize): number steps-dir files globally so they are not overwritten (#1023)

### LINKED ISSUE #1023: dataize --steps-dir overwrites steps, only normalization is stored
With `phino dataize --steps-dir DIR`, every normalization rewrite writes its steps to the same numbered files and overwrites the previous step's output, so the stored chain is not a faithful sequence of the dataization.

## Problem

In `src/Dataize.hs` `normalized` (around line 331):

```haskell
(rewrittens, _) <- rewrite whole normalizationRules (rewriteContext ctx)
```

`rewriteContext` (line 343) forwards `_saveStep` from the dataization context, so each `rewrite` invocation writes intermediate steps via `SaveStepFunc`. `saveStep` (`src/Deps.hs:50`) names files as `printf "%05d.%s" step ext`, counting from 1 for each rewrite. Two consequences:

- only the steps of the innermost/last normalization survive (they overwrite the same `00001…` names), and
- steps of the outer morphing/dataization (`𝕄`/`𝔻`) are not stored at all (they live only in the returned `NonEmpty Rewritten` chain in memory).

For a long `--sequence`, the `--steps-dir` output is therefore not a faithful record of the dataization.

## Expected

Confirm the intended behavior. If `--steps-dir` should capture every step of the whole `dataize` run, the numbering must be global (a shared counter), not restarting per `rewrite`. If only normalization steps are meant to be dumped, that should be documented (and the clutter of rewritten identical names avoided).

VasilevNStas:
Fixes #1023.

## Problem

`dataize --steps-dir DIR` did not produce a faithful sequence of steps. `saveStep` (`src/Deps.hs:46-53`) wrote files named `printf "%05d.%s" step ext`, where `step` was computed inside a single `rewrite` call (`((iteration - 1) * _maxDepth) + _count`, `src/Rewriter.hs:227`). Dataization runs normalization many times (`src/Dataize.hs` `normalized`), and each call restarted the counter from 1, so steps from later normalizations **overwrote** the files written by earlier ones (`00001.phi` was repeatedly clobbered). The outer morphing/dataization steps were never written at all.

## Solution

Make the step counter global for the whole run. `saveStepFunc` (`src/CLI/Helpers.hs`) now returns `IO SaveStepFunc` and closes over an `IORef` counter; every saved step takes the next number. Because the counter now supplies the number, the caller-supplied step parameter became dead weight, so `SaveStepFunc` no longer takes an `Int` (`src/Deps.hs`): it is now `Expression -> IO ()`, `saveStep` takes the counter-supplied `Int` itself, and `Rewriter` just calls `_saveStep updated` without computing `((iteration - 1) * _maxDepth) + _count`. Both `runRewrite` and `runDataize` obtain the counter once and thread the resulting `SaveStepFunc` into their context records (the `context` where-helpers were removed as they only forwarded the steps-dir).

With `dataize --steps-dir` on `[[ bytes(data) -> [[ @ -> $.data ]], number(as-bytes) -> [[ @ -> $.as-bytes, plus(x) -> [[ L> L_number_plus ]] ]], @ -> 5.plus(6) ]]` we now get 38 distinct files `00001.phi`..`00038.phi` instead of a handful of repeatedly overwritten names.

### Behaviour change for `rewrite --steps-dir`

The PR description previously claimed the output is unchanged, which understated the change. The old numbering was sparse: iteration 2 started at `_maxDepth + 1`, so the gaps encoded the cycle boundaries. Dense numbering is arguably better, but it **is** a behaviour change whenever a rewrite runs more than one cycle — step names shift and the gap markers disappear. This is intentional (dense is what makes the global counter meaningful), but calling it out here.

### Scope note

This fixes the overwriting of normalization steps. The outer 𝕄/𝔻 steps that #1023 also mentions are still not written — `_saveStep` is only called from `Rewriter.hs`. That is left for a separate follow-up, so #1023 should not be closed as fully addressed.

## Test

Updated `test/CLISpec.hs` case "saves dataize steps to dir with --steps-dir". Instead of pinning an exact file count (which depends on the current rule set), it asserts what #1023 is about: the step file names are **distinct and contiguous from `00001`** (`steps == map (\n -> printf "%05d.phi" n) [1..length steps]`), and the count **exceeds what a single normalization produces** (`length steps > 18`), proving several normalizations contributed rather than one overwriting pass. `make test` passes (1122 examples).

## Checklist

- [x] `make hlint` — no hints
- [x] `make fourmolu` — clean

### DIFF
--- a/src/CLI/Helpers.hs
+++ b/src/CLI/Helpers.hs
@@ -49,9 +49,9 @@ saveStepFunc stepsDir ctx@PrintCtx{..} = do
         | otherwise = show _outputFormat
       render = printInFormat ctx
       save :: SaveStepFunc
-      save expr _ = do
+      save expr = do
         step <- atomicModifyIORef' counter (\value -> (value + 1, value + 1))
-        saveStep stepsDir ioToExt render expr step
+        saveStep stepsDir ioToExt render step expr
   pure save
 
 -- Read input from file or stdin
--- a/src/CLI/Helpers.hs
+++ b/src/CLI/Helpers.hs
@@ -13,6 +13,7 @@ import Canonizer (canonize)
 import Control.Exception
 import Control.Monad ((>=>))
 import Data.Functor ((<&>))
+import Data.IORef
 import Data.List (intercalate)
 import Data.Maybe
 import Deps (SaveStepFunc, saveStep)
@@ -39,13 +40,19 @@ justMeetLength :: Maybe Int -> Int
 justMeetLength = fromMaybe defaultMeetLength
 
 -- Prepare saveStepFunc
-saveStepFunc :: Maybe FilePath -> PrintContext -> SaveStepFunc
-saveStepFunc stepsDir ctx@PrintCtx{..} = saveStep stepsDir ioToExt (printInFormat ctx)
-  where
-    ioToExt :: String
-    ioToExt
-      | _outputFormat == LATEX = "tex"
-      | otherwise = show _outputFormat
+saveStepFunc :: Maybe FilePath -> PrintContext -> IO SaveStepFunc
+saveStepFunc stepsDir ctx@PrintCtx{..} = do
+  counter <- newIORef (0 :: Int)
+  let ioToExt :: String
+      ioToExt
+        | _outputFormat == LATEX = "tex"
+        | otherwise = show _outputFormat
+      render = printInFormat ctx
+      save :: SaveStepFunc
+      save expr _ = do
+        step <- atomicModifyIORef' counter (\value -> (value + 1, value + 1))
+        saveStep stepsDir ioToExt render expr step
+  pure save
 
 -- Read input from file or stdin
 readInput :: Maybe FilePath -> IO String
--- a/src/CLI/Runners.hs
+++ b/src/CLI/Runners.hs
@@ -60,7 +60,8 @@ runRewrite OptsRewrite{..} = do
       printCtx = toPrintCtx xmirCtx foc
       exclude = (`F.exclude` excluded)
       include = (`F.include` included)
-  (rewrittens, exceeded) <- rewrite expr rules (context loc printCtx)
+  save <- saveStepFunc _stepsDir printCtx
+  (rewrittens, exceeded) <- rewrite expr rules (RewriteContext loc _maxDepth _maxCycles _depthSensitive buildTerm _must _breakpoint save)
   let rewrittens' = exclude $ include (if _sequence then NE.toList rewrittens else [NE.last rewrittens])
   logDebug (printf "Printing rewritten 𝜑-expression as %s" (show _outputFormat))
   exprs <- printRewrittens printCtx (rewrittens', exceeded)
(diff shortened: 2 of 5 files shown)


VasilevNStas:
@maxonfjvipon @yegor256 
plz review

yegor256:
The global counter is the right call for #1023, but it leaves dead weight behind.

**1. The `Int` in `SaveStepFunc` is now unused everywhere.**

`type SaveStepFunc = Expression -> Int -> IO ()` (`src/Deps.hs:42`). After this PR the only real implementation ignores it (`save expr _ = ...`), and `dontSaveStep` ignores it too. Meanwhile `src/Rewriter.hs:205` still computes the argument:

```haskell
_saveStep updated (((iteration - 1) * _maxDepth) + _count)
```

That arithmetic now exists only to be discarded, and the type still promises a caller-supplied step number that no longer means anything. Drop the parameter: `type SaveStepFunc = Expression -> IO ()`, `saveStep :: Maybe FilePath -> String -> (Expression -> IO String) -> Int -> SaveStepFunc` keeping the counter-supplied step, and `_saveStep updated` at the call site.

**2. The PR description understates the change to `rewrite --steps-dir`.**

It says the output is unchanged, but the old numbering was sparse across iterations: iteration 2 started at `_maxDepth + 1`, so gaps encoded the cycle boundaries. Dense numbering is the better behaviour, but it is a behaviour change whenever a rewrite runs more than one cycle. Say so in the description.

**3. `length steps `shouldBe` 38` is brittle.**

The count is a property of the current normalization rule set, not of the fix. Any rule added to `resources/*.yaml` breaks this test for a reason unrelated to step numbering. Assert what #1023 is actually about instead: that the file names are distinct and contiguous from `00001`, and that the count exceeds what a single normalization produces.

Also worth noting for the record: this fixes the overwriting, but the outer 𝕄/𝔻 steps that #1023 mentions are still not written — `_saveStep` is only called from `Rewriter.hs`. Fine to leave for a separate issue, just say it in the description so #1023 is not closed as fully addressed.

---

VasilevNStas:
@yegor256 thanks for the review — all three points are addressed in the updated commits:

**1. Dead `Int` in `SaveStepFunc`** — dropped. Now `type SaveStepFunc = Expression -> IO ()` (src/Deps.hs), `saveStep :: Maybe FilePath -> String -> (Expression -> IO String) -> Int -> SaveStepFunc` keeps the counter-supplied step, `dontSaveStep = saveStep Nothing "" (\_ -> pure "") 0`, and `Rewriter.hs` just calls `_saveStep updated` — the `((iteration - 1) * _maxDepth) + _count` arithmetic is gone.

**2. Behaviour change for `rewrite --steps-dir`** — the description now says it explicitly: old numbering was sparse (iteration 2 started at `_maxDepth + 1`, gaps encoded cycle boundaries); dense numbering is intentional but is a behaviour change whenever a rewrite runs more than one cycle.

**3. Brittle `length == 38`** — replaced. The test now asserts what #1023 is about: file names are distinct and contiguous from `00001` (`steps == map (\n -> printf "%05d.phi" n) [1..length steps]`), and the count exceeds what a single normalization produces (`length steps > 18`), so a later rule-set change won't break it.

**Scope note** (also added to the description): outer 𝕄/𝔻 steps are still not written — `_saveStep` is only called from `Rewriter.hs`. Left for a separate follow-up, so #1023 should not be closed as fully addressed.

`make test`/hlint/fourmolu pass.

yegor256:
Thanks for the contribution!

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+518](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1030, opened 2026-08-08
feat(random): pad %d to a fixed 4-digit width in random-string (#1019)

### LINKED ISSUE #1019: random-string %d expands without leading-zero padding
`Random.generate` expands the `%d` directive as `show <$> randomRIO (0::Int, 9999)`, producing a number without leading-zero padding (e.g. `42` instead of `0042`).

## Problem

In `src/Random.hs:28`:

```haskell
'd' -> show <$> randomRIO (0 :: Int, 9999)
```

So a pattern like `"L%d"` or `"target%d"` yields names with a 1-4 digit suffix (`L3`, `L42`, `L567`, `L8901`). If the intended semantics is a fixed-width 4-digit number (as `%04d` would give), then lengths of synthetic names are unstable across values, which can matter for byte-length-sensitive downstream tools or for consistent-looking output.

## Expected

Clarify the intended format. If a fixed-width 4-digit number is desired (matching `%04d`), pad the result with leading zeros so every `%d` expands to exactly 4 digits. Otherwise, document that `%d` is unpadded. The fix is trivial and low-risk.

VasilevNStas:
Fixes #1019.

## Problem

`Random.generate` expanded `%d` as `show <$> randomRIO (0 :: Int, 9999)`, producing a **variable-length** number: `"3"`, `"42"`, `"567"`, `"8901"`. Meanwhile `%x` expands to a **fixed** 8-char hex string. So synthetic names like `L%d` had unstable lengths across values, and `%d`/`%x` behaved inconsistently with each other.

## Proposed fix

Make `%d` a fixed **4-digit** number, matching `%04d` (and mirroring `%x`'s fixed width):

```haskell
'd' -> printf "%04d" <$> randomRIO (0 :: Int, 9999)
```

`42` now becomes `0042`. The numeric range is unchanged (`0..9999`), only the rendering is padded to a constant width. If a different format is preferred (e.g. unpadded, or a different width), this PR can be rejected — it is a proposal to make the output deterministic in length.

## Test

Updated `test/RandomSpec.hs`: the "%d pattern length" case now asserts `length result == 4` instead of the previous `1..4` window. All other cases (`%d` digits, range `0..9999`, prefix/suffix, uniqueness) still hold. `make test` passes (1122 examples).

## Checklist

- [x] `make hlint` — no hints
- [x] `make fourmolu` — clean

### DIFF
--- a/src/Random.hs
+++ b/src/Random.hs
@@ -13,6 +13,7 @@ import qualified Data.Vector.Mutable as M
 import GHC.IO (unsafePerformIO)
 import System.Random (newStdGen, randomRIO)
 import System.Random.Stateful (newIOGenM, uniformRM)
+import Text.Printf (printf)
 
 strings :: IORef (Set String)
 {-# NOINLINE strings #-}
@@ -25,7 +26,7 @@ generate ('%' : ch : rest) = do
     'x' -> replicateM 8 $ do
       v <- randomRIO (0, 15)
       pure (intToDigit v)
-    'd' -> show <$> randomRIO (0 :: Int, 9999)
+    'd' -> printf "%04d" <$> randomRIO (0 :: Int, 9999)
     _ -> pure ['%', ch]
   next <- generate rest
   pure (rep ++ next)
--- a/test/RandomSpec.hs
+++ b/test/RandomSpec.hs
@@ -44,10 +44,10 @@ spec = do
       result `shouldSatisfy` all isDigit
 
   describe "randomString with %d pattern length" $
-    it "generates 1-4 digit number" $ do
+    it "generates a fixed 4-digit number" $ do
       result <- randomString "%d"
       let len = length result
-      len `shouldSatisfy` (\l -> l >= 1 && l <= 4)
+      len `shouldBe` 4
 
   describe "randomString with %x pattern" $
     it "generates hex digits" $ do


VasilevNStas:
About the benchmark comparison reported here:

The `rewrite/normalize` row shows `-27.7% faster` with ±14.9% noise. This is an artifact of the shared CI runner's noise, not a real speedup: this PR only changes `Random.generate` (`%04d` padding for `random-string`), it does not touch rewriting. Note that `rewrite/normalize` is the fastest bench (~57-78 μs), so its rounds are the noisiest (other PRs here show ±20-30% for the same row). The report itself states `a delta never fails the build`, and no case is marked `slower` beyond its own noise.

VasilevNStas:
@yegor256 @volodya-lombrozo 
please review this PR

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (7) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+478](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1041, opened 2026-08-31
Keep bound ρ and λ function name in XMIR output

### LINKED ISSUE #1036: XMIR output drops bound ρ bindings and λ function names (src/XMIR.hs:155,157)
When an expression is rendered to XMIR, two parts of it are silently discarded: a bound `ρ` binding is dropped entirely (`src/XMIR.hs:155`, `formationBinding (BiTau AtRho _) _ = pure Nothing`), and the function name of a `λ` binding is erased (`src/XMIR.hs:157` renders `BiLambda _` as a bare `<o name="λ"/>`). For example, this (phino 0.0.109):

```
printf '⟦ k ↦ ⟦ x ↦ ∅, λ ⤍ Lorg_eolang_number_plus, ρ ↦ ⟦ Δ ⤍ 40-45-00-00-00-00-00-00 ⟧ ⟧ ⟧' | phino rewrite --output xmir
```

produces `<o name="k"><o base="∅" name="x"/><o name="λ"/></o>` — both the `ρ ↦ ⟦ Δ ⤍ … ⟧` binding and the name `Lorg_eolang_number_plus` survive only inside the escaped `<listing>`, not in the object tree.

This makes XMIR lossy as an output format: a `rewrite --input xmir --output xmir` pipeline cannot round-trip its own results, and the reader compensates by reconstructing the λ name positionally as `L_<fqn>` (`src/XMIR.hs:394`), which is wrong for any atom whose function name is not derivable from its location in the tree.

Suggested fix: in `formationBinding`, emit a bound `ρ` as a regular named child element (keeping the `pure Nothing` only for the void `ρ` case on line 158), and carry the λ function name in an attribute on the `λ` marker element (the way EO's XMIR carries `atom="…"`), with the reader at `src/XMIR.hs:394` preferring that attribute over positional reconstruction.

yegor256:
Closes #1036.

XMIR was throwing away two things on the way out. A bound `ρ` was dropped
outright, and the `λ` function name was erased down to a bare `<o name="λ"/>`.
The reader then guessed the name back from where the element sat in the tree,
as `L_<fqn>`, which is right only when the name happens to follow the position.
So `phino rewrite --input xmir --output xmir` could not reproduce its own
output, and any atom named something like `Lorg_eolang_number_plus` came back
as `L_k`.

The writer now emits a bound `ρ` the same way it emits any other named
attribute, and puts the `λ` name in the text of the marker element. Text rather
than an attribute, because the XMIR.xsd that phino itself points at leaves no
room: the `o` element has no `anyAttribute`, so its attribute list is closed,
and `atom` is typed as an atom return-type signature whose pattern rejects
anything starting with an uppercase letter. The same `o` element is declared
`mixed="true"`, so text is schema-valid. On the way back in, `name="ρ"` maps to
`AtRho` again, and the `λ` text wins over the positional guess when it is
there — when it is not, the old reconstruction still runs, so XMIR coming from
EO reads exactly as before.

One thing the reviewer should know: the example in #1036 still does not survive
a full round-trip, and this change is not why. A formation binding with `Δ`
loses its bytes on the way in — `⟦ k ↦ ⟦ Δ ⤍ 01-02 ⟧ ⟧` reads back as
`⟦ k ↦ ⟦⟧ ⟧` on master today, with or without this patch, because
`xmirToFormationBinding` never looks at the text content. Separate defect,
separate fix; I left it alone and it is worth its own issue.

`make test` gives 1132 examples and no failures, `make hlint` finds no hints,
`make fourmolu` is clean.

### DIFF
--- a/src/XMIR.hs
+++ b/src/XMIR.hs
@@ -143,22 +143,25 @@ expression (ExApplication expr arg) ctx = do
 expression expr _ = throwIO (UnsupportedExpression expr)
 
 formationBinding :: Binding -> XmirContext -> IO (Maybe Node)
-formationBinding (BiTau (AtLabel label) (ExFormation bds)) ctx = do
-  inners <- nestedBindings bds ctx
-  pure (Just (object [("name", T.unpack label)] inners))
-formationBinding (BiTau (AtLabel label) expr) ctx = do
-  (base, children) <- expression expr ctx
-  pure (Just (object [("name", T.unpack label), ("base", base)] children))
+formationBinding (BiTau (AtLabel label) expr) ctx = Just <$> namedBinding (T.unpack label) expr ctx
+formationBinding (BiTau AtRho expr) ctx = Just <$> namedBinding (show AtRho) expr ctx
 formationBinding (BiTau AtPhi expr) ctx = do
   (base, children) <- expression expr ctx
   pure (Just (object [("name", show AtPhi), ("base", base)] children))
-formationBinding (BiTau AtRho _) _ = pure Nothing
 formationBinding (BiDelta bytes) _ = pure (Just (NodeContent (T.pack (printBytes bytes))))
-formationBinding (BiLambda _) _ = pure (Just (object [("name", show AtLambda)] []))
+formationBinding (BiLambda (Function name)) _ = pure (Just (object [("name", show AtLambda)] [NodeContent name]))
 formationBinding (BiVoid AtRho) _ = pure Nothing
 formationBinding (BiVoid AtPhi) _ = pure (Just (object [("name", show AtPhi), ("base", "∅")] []))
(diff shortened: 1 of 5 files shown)


yegor256:
@maxonfjvipon could you take a look when you get a chance? Happy to explain any part of it, especially the choice to put the λ name in text instead of an attribute.

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1646](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1042, opened 2026-08-31
Apply --rule together with --normalize

### LINKED ISSUE #1037: rewrite --normalize silently ignores --rule files (src/CLI/Helpers.hs:116)
`phino rewrite --normalize --rule my.yaml` silently ignores the custom rule file. In `src/CLI/Helpers.hs:116`, `getRules normalize shuffle rules` returns only the built-in `normalizationRules` whenever `normalize` is true and never reads the `rules` argument, so `runRewrite` (`src/CLI/Runners.hs:48`) proceeds as if `--rule` had not been given. For example, a rule rewriting `⟦ λ ⤍ Marker ⟧` to `⟦ Δ ⤍ FF- ⟧` fires on `⟦ k ↦ ⟦ m ↦ ⟦ Δ ⤍ 01- ⟧ ⟧.m, j ↦ ⟦ λ ⤍ Marker ⟧ ⟧` when passed via `--rule` alone, but with `--normalize --rule marker.yaml` the output shows the dispatch normalized while `j` still holds `λ ⤍ Marker` (phino 0.0.109).

This is wrong twice over: the user's explicit instruction is dropped without a word, and the behavior is inconsistent with `explain`, which rejects the same flag combination outright (`src/CLI/Runners.hs:204`, "Only one of --rule, --normalize, --morph, --dataize or --contextualize can be specified"). A caller who wants normalization plus a few domain rules is today forced to ship private copies of all 14 built-in normalize YAMLs, which then drift on every phino upgrade.

Suggested fix: in `getRules`, when both `normalize` is true and `rules` is non-empty, concatenate — load the user YAMLs and append them to `normalizationRules` — so one rewriting run applies both sets; failing that, add the same `invalidCLIArguments` guard `rewrite` currently lacks, so the combination fails loudly instead of silently.

yegor256:
Closes #1037.

`getRules` used to return the built-in normalization set whenever
`--normalize` was on, without ever looking at the `--rule` file list. So
`phino rewrite --normalize --rule my.yaml` quietly behaved as if the second
flag had not been typed. Now the two sets are concatenated, with the user
rules running after the built-in ones, and `--shuffle` covers the whole list.

The reason this matters: anyone who wanted normalization plus a couple of
their own domain rules had to copy all fourteen built-in normalize YAMLs into
their own tree and keep them in sync by hand on every phino upgrade. There was
no way to say "normalize, and also do this".

Two things for the reviewer. First, `--rule` is validated now even under
`--normalize`, so a typo in a path or a dataization-only function in a rule
stops the run instead of slipping past — that is a behaviour change, though I
think it is the right one. Second, `explain` still turns down the same flag
combination in its own `validateOpts` guard, and I left that alone since the
issue is about `rewrite`; whether `explain --normalize --rule x.yaml` should
print both sets is worth deciding on its own.

`make test` gives 1133 examples and no failures, `make hlint` finds no hints,
`make fourmolu` and markdownlint are clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,13 @@ to normalize expressions:
 phino rewrite --normalize hello.phi
 ```
 
+Both flags may be combined, so that your own rules are applied
+alongside the built-in ones, in a single rewriting session:
+
+```bash
+phino rewrite --normalize --rule=my-rule.yaml hello.phi
+```
+
 Some rules mint fresh synthetic names via the `random-string` built-in. To
 keep the output reproducible across runs, `phino` seeds the random generator
 deterministically with `0` by default. Use `--seed` to pick a different seed:
--- a/src/CLI/Helpers.hs
+++ b/src/CLI/Helpers.hs
@@ -112,29 +112,32 @@ printCtxToLatexCtx :: PrintContext -> LatexContext
 printCtxToLatexCtx PrintCtx{..} =
   LatexContext _sugar _line _margin _nonumber _compress _canonize _meetPopularity _meetLength _focus _expression _label _meetPrefix _headers
 
--- Get rules for rewriting depending on provided flags
+-- Get rules for rewriting depending on provided flags. Both flags may be given
+-- together, in which case the user rules follow the built-in ones
 getRules :: Bool -> Bool -> [FilePath] -> IO [Y.Rule]
 getRules normalize shuffle rules = do
-  ordered <-
-    if normalize
-      then do
-        let rules' = normalizationRules
-        logDebug (printf "The --normalize option is provided, %d built-it normalization rules are used" (length rules'))
-        pure rules'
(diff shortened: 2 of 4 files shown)


yegor256:
@maxonfjvipon mind taking a look when you have a moment? Happy to explain anything, particularly the call to run the user rules after the built-in ones rather than before.

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1662](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1043, opened 2026-08-31
Implement the missing EO primitive atoms in dataize

### LINKED ISSUE #1038: dataize implements only 3 of the 12 EO primitive atoms (src/Dataize.hs:392)
The dataization atom table implements only 3 of the 12 primitive atoms that EO data operations reduce to. In `src/Dataize.hs:371-392` the `atom` function handles `L_number_plus`, `L_number_times` and `L_number_eq`, and the catch-all at line 392 throws `userError "Atom '%s' does not exist"`. Dataizing an expression as ordinary as a division, e.g. `⟦ φ ↦ ⟦ x ↦ ∅, λ ⤍ Lorg_eolang_number_div, ρ ↦ ⟦ Δ ⤍ 40-45-00-00-00-00-00-00 ⟧ ⟧(x ↦ ⟦ Δ ⤍ 40-00-00-00-00-00-00-00 ⟧) ⟧`, dies with `[ERROR]: user error (Atom 'Lorg_eolang_number_div' does not exist)` (phino 0.0.109).

EO's runtime defines exactly 12 primitive λ-atoms underneath all data operations: `number.plus/times/div/gt` and `bytes.and/concat/eq/not/or/right/size/slice` (declared in `eo-runtime/src/main/eo/number.eo` and `bytes.eo`, with reference semantics in `eo-runtime/src/main/java/org/eolang/EOnumber$EO*.java` and `EObytes$EO*.java`). With 9 of them missing, `phino dataize` cannot evaluate real EO expressions — any division, comparison, or bytes manipulation stops the run — which blocks using phino as a compile-time evaluator inside the EO compiler.

Suggested fix: extend the `atom` table with the missing 9 cases, mirroring the Java atoms' semantics (IEEE double division, `gt` yielding the bool selection, bitwise ops and shift on byte arrays, `concat`, `size`, and `slice` with its bounds behavior). This continues the path of #188, which added `L_number_times` the same way.

yegor256:
The atom table in `src/Dataize.hs` knew only `plus`, `times` and `eq`, so any
division, comparison or bytes manipulation stopped a run with "Atom does not
exist". It now covers all 12 primitives EO's data operations reduce to:
`number.div`/`gt` and `bytes.and`/`concat`/`eq`/`not`/`or`/`right`/`size`/`slice`
join the three that were already there. The pure byte-array arithmetic went into
`src/Bytes.hs`, next to the mapping from `Bytes` onto the octets underneath, and
`AST.dataBytes` is the new constructor for the `Φ.bytes(data ↦ ⟦ Δ ⤍ … ⟧)`
object the bytes atoms yield — the `DataObject` pattern reuses it now too.

The semantics follow the Java runtime. Division is plain IEEE double division.
`gt` and `bytes.eq` yield `Φ.true` or `Φ.false`, the same thing
`Data.ToPhi(boolean)` hands back, so the answer can still be dispatched with
`.if`. `and` and `or` refuse operands of different lengths, as
`BytesRaw.ofSameLength` does. `right` reproduces `BytesRaw.shift`, including the
left shift a negative distance asks for. `slice` turns down bounds below zero or
outside the int range, and an out-of-bounds window goes to the `cant-slice`
fallback carrying EO's own complaint text instead of terminating. The 23 new
cases in `test/DataizeSpec.hs` take their vectors straight from `bytes.eo` and
`number.eo`.

Two things worth deciding separately. phino spells these `L_number_div`,
`L_bytes_and` and so on, matching the existing `L_number_plus`, so the
`Lorg_eolang_number_div` in the issue's snippet still needs the phino spelling —
someone should say whether the runtime spelling ought to be accepted as an alias.
And that snippet also needs `bytes` and `number` declared in its universe:
without them phino spins forever on the unresolved `Φ.number` dispatch. That is
not new, the existing `L_number_plus` hangs the same way in the same bare
universe, but it deserves its own issue.

Closes #1038

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -228,15 +228,12 @@ pattern DataObject :: T.Text -> Bytes -> Expression
 pattern DataObject label bts <- (matchDataObject -> Just (label, bts))
   where
     DataObject label bts =
-      ExApplication
-        (BaseObject label)
-        ( ArTau
-            (AtLabel "as-bytes")
-            ( ExApplication
-                (BaseObject "bytes")
-                ( ArTau
-                    (AtLabel "data")
-                    (ExFormation [BiDelta bts, BiVoid AtRho])
-                )
-            )
-        )
+      ExApplication (BaseObject label) (ArTau (AtLabel "as-bytes") (dataBytes bts))
+
+-- The bytes object Φ.bytes(data ↦ ⟦ Δ ⤍ …, ρ ↦ ∅ ⟧) — what a 'bytes' atom
+-- yields and what a 'DataObject' carries under its 'as-bytes' argument
+dataBytes :: Bytes -> Expression
+dataBytes bts =
+  ExApplication
+    (BaseObject "bytes")
+    (ArTau (AtLabel "data") (ExFormation [BiDelta bts, BiVoid AtRho]))
--- a/src/Bytes.hs
+++ b/src/Bytes.hs
@@ -4,7 +4,9 @@
 -- SPDX-License-Identifier: MIT
 
 -- This module is a codec between 'Bytes' and the values they encode:
--- IEEE-754 doubles, UTF-8 strings and raw hex.
+-- IEEE-754 doubles, UTF-8 strings and raw hex. It also owns the byte-array
+-- operations that EO's 'bytes' atoms are built on, since only this module knows
+-- how a 'Bytes' maps onto the octets underneath it.
 module Bytes
   ( numToBts
   , strToBts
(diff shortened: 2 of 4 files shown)


yegor256:
@maxonfjvipon could you take a look when you get a chance? Happy to explain any of the atom semantics or split the naming question out into its own issue if you'd rather decide that separately.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (351 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1670](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1018, opened 2026-08-07
XMIR output hard-codes revision and ms attributes


VasilevNStas:
`expressionToXMIR` hard-codes fake XML attributes `revision="1234567"` and `ms="0"` on the generated `<object>` element instead of real values.

## Problem

`src/XMIR.hs` (`expressionToXMIR`, around `XMIR.hs:190,193`) produces the XML root like:

```xml
<object ms="0" revision="1234567" ...>
```

`revision` is a hard-coded placeholder `"1234567"` and `ms` is a constant `"0"`. The surrounding code correctly fills `dob`/`time` from the current clock and `version` from the package version, but these two attributes carry fake data. Downstream tools that rely on `revision` (or a meaningful `ms`) will read a wrong value.

## Expected

Either fill `revision` with a real value (e.g. the build version or a hash) and compute `ms` properly, or drop the attributes entirely if they are not meaningful for the 2D representation. Clarifying which is intended requires knowing the target XMIR schema expectations.

VasilevNStas:
Hi! To move this forward, we need a maintainer decision on the XMIR expectations (this is a spec question, not something we can resolve from the code alone):

1. Should `revision` be filled with a real value, and if so, what source — the build version (`version` already comes from the package), a source hash, or something else?
2. Should `ms` carry a real millisecond timestamp (like `dob`/`time` do), or can it stay `0`?
3. Or is the preferred fix simply to drop `revision`/`ms` from the `<object>` element entirely?

We're happy to implement whichever variant you choose — just let us know the intended behavior.

yegor256:
Two of your three questions are answered by the schema the output itself points at. `src/XMIR.hs:197` emits:

```
xsi:noNamespaceSchemaLocation="https://raw.githubusercontent.com/objectionary/eo/refs/heads/gh-pages/XMIR.xsd"
```

In that XSD, `complexType name="object"` declares:

```xml
<xs:attribute name="revision">
  <xs:appinfo>The Git SHA of the release of the parser that created the document</xs:appinfo>
  ... may be omitted in production systems.
  <xs:restriction base="xs:string"><xs:pattern value="[0-9a-f]{7}"/></xs:restriction>
</xs:attribute>
<xs:attribute name="ms" type="whole">
  <xs:appinfo>The amount of milliseconds that were spent on creating this document</xs:appinfo>
</xs:attribute>
```

So:

[...]

PR #1044, opened 2026-08-31
Declare the required author attribute in XMIR output

### LINKED ISSUE #1039: XMIR Output Invalid: Mandatory 'author' Attribute Omitted
Every document produced by `phino rewrite --output xmir` is invalid against the schema it declares. The root `<object>` element is built in `expressionToXMIR` at `src/XMIR.hs:190`, and its attribute list at `src/XMIR.hs:191-197` carries `dob`, `ms`, `revision`, `time`, `version`, `xmlns:xsi` and `xsi:noNamespaceSchemaLocation` — but not `author`.

That schema is the one phino itself points at, on line 197: `https://raw.githubusercontent.com/objectionary/eo/refs/heads/gh-pages/XMIR.xsd`. Its `complexType name="object"` declares six attributes, and `author` is the only one carrying `use="required"`:

```xml
<xs:attribute name="author" use="required">
  <xs:appinfo>The author of the XMIR file</xs:appinfo>
  <xs:restriction base="xs:string"><xs:pattern value="[a-z]+(-[a-z]+)*"/></xs:restriction>
</xs:attribute>
```

A document that advertises a schema and then violates its one mandatory attribute cannot be consumed by any tool that validates before parsing, which defeats the purpose of emitting `xsi:noNamespaceSchemaLocation` at all.

The fix is to add `("author", "phino")` to the attribute list at `src/XMIR.hs:191`. The value matches the required pattern `[a-z]+(-[a-z]+)*`.

This overlaps with #1018 in file and element but is a distinct defect: #1018 is about two optional attributes carrying fabricated values, while this one is a required attribute that is absent.

yegor256:
The root `<object>` element now carries `author="phino"`. That is the one
attribute the `object` complexType in `XMIR.xsd` marks `use="required"`, and the
value fits its `[a-z]+(-[a-z]+)*` pattern.

phino points at that schema itself, through the `xsi:noNamespaceSchemaLocation`
it writes on the same element, so until now every document out of
`phino rewrite --output xmir` was turned away by anything that validates before
parsing — which rather defeats the point of emitting the schema location. I
checked it by hand with `xmllint` against a local copy of the schema: the old
output says `Element 'object': The attribute 'author' is required but missing`
and fails, the new output validates.

The new printing pack `test-resources/xmir-printing-packs/document-author.yaml`
pins it with the xpath `/object[@author="phino"]`. The suite cannot fetch the
schema itself, since the tests assume no network, so that xpath is the guard.

Worth noting that #1018 is still open on this same element, about `ms="0"` and
`revision="1234567"` being placeholders rather than real values. I left those
alone — different defect.

Closes #1039

### DIFF
--- a/src/XMIR.hs
+++ b/src/XMIR.hs
@@ -191,7 +191,8 @@ expressionToXMIR expr@(ExFormation [BiTau (AtLabel _) arg, BiVoid AtRho]) ctx@Xm
             (Prologue [] Nothing [])
             ( element
                 "object"
-                [ ("dob", formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%S" now)
+                [ ("author", "phino")
+                , ("dob", formatTime defaultTimeLocale "%Y-%m-%dT%H:%M:%S" now)
                 , ("ms", "0")
                 , ("revision", "1234567")
                 , ("time", time now)
--- a/test-resources/xmir-printing-packs/document-author.yaml
+++ b/test-resources/xmir-printing-packs/document-author.yaml
@@ -1,0 +1,10 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+# yamllint disable rule:line-length
+---
+phi: |
+  [[
+    x -> 5
+  ]]
+xpaths:
+  - '/object[@author="phino"]'


yegor256:
@maxonfjvipon mind taking a look when you have a moment? It's a one-line change, but say the word if you'd rather the author value came from somewhere other than a hardcoded "phino".

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (13) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1674](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1047, opened 2026-08-31
fix(xmir): emit real revision and ms attributes, matching the schema (#1018)

### LINKED ISSUE #1018: XMIR output hard-codes revision and ms attributes

VasilevNStas:
Fixes #1018.

## Problem

`expressionToXMIR` emitted fabricated XML attributes on the `<object>` element: `revision="1234567"` (a fake Git SHA) and `ms="0"` (a constant). Per the XMIR schema (the `xsi:noNamespaceSchemaLocation` the output itself points at), `revision` is documented as *"The Git SHA of the release of the parser that created the document"* and `ms` as *"The amount of milliseconds that were spent on creating this document"*. `"1234567"` only happens to satisfy the `[0-9a-f]{7}` pattern; it asserts a commit that does not exist.

## Solution

- **`revision`** — the 7-character Git SHA of the build that produced the document, spliced at compile time via `Development.GitRev` (`gitrev`) Template Haskell (`take 7 $(gitHash)`, src/XMIR.hs). When built outside a git checkout `gitHash` yields `"UNKNOWN"`, and per the schema (the attribute is optional) it is then **omitted** rather than replaced with a placeholder.
- **`ms`** — the wall-clock milliseconds actually spent producing the document: a `getCurrentTime` at the start of `expressionToXMIR'`, subtracted and rounded at the end.

The `author` attribute required by the schema (`author="phino"`) was added separately in upstream #1044; this PR keeps it and only replaces the two placeholder values.

## Test

Added `test/CLISpec.hs` case "emits a real revision and ms in XMIR": asserts `revision` is 7 hex chars and not `1234567`, and `ms` is numeric. `make test` passes (1159 examples).

## Checklist

- [x] `make test` passes (1159 examples)
- [x] `make hlint` — no hints
- [x] `make fourmolu` — clean

### DIFF
--- a/phino.cabal
+++ b/phino.cabal
@@ -91,6 +91,7 @@ library
     directory >=1.3.7 && <1.4,
     file-embed >=0.0.15 && <0.0.17,
     filepath >=1.4.200 && <1.6,
+    gitrev >=1.3.1 && <1.4,
     megaparsec >=9.5 && <9.9,
     optparse-applicative >=0.18 && <0.20,
     random >=1.2 && <1.4,
--- a/src/XMIR.hs
+++ b/src/XMIR.hs
@@ -2,6 +2,7 @@
 {-# LANGUAGE NumericUnderscores #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 -- SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
 -- SPDX-License-Identifier: MIT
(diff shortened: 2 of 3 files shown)


VasilevNStas:
@yegor256 
please review this PR

yegor256:
This now has merge conflicts with master (#1029, which touched the same SaveStepFunc/IORef plumbing, just merged). Needs a rebase before it can go in.

---

VasilevNStas:
@yegor256 
I`m fixed it

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2142](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1050, opened 2026-09-01
Keep Δ data when reading named bindings from XMIR

### LINKED ISSUE #1045: XMIR reader drops Δ data in named bindings written by its own printer (src/XMIR.hs:396)
Follow-up to #1036, seen on the released 0.0.110: the XMIR round-trip still loses Δ data held in named bindings. The writer prints `a ↦ ⟦ Δ ⤍ 01-02 ⟧` as `<o name="a">01-02</o>` (`formationBinding` emits the bytes as `NodeContent`, `src/XMIR.hs:151`), but the reader's no-`@base` branch in `xmirToFormationBinding` (`src/XMIR.hs:391-403`) collects only child elements (`cur C.$/ C.element "o"`, line 396) and never looks at the element's text content, so the binding comes back as an empty formation.

For example, `printf '⟦ k ↦ ⟦ a ↦ ⟦ Δ ⤍ 01-02 ⟧, ρ ↦ ∅ ⟧ ⟧' | phino rewrite --output xmir` prints `<o name="k"><o name="a">01-02</o></o>`, and feeding that back through `phino rewrite --input xmir --output phi` yields `⟦ k ↦ ⟦ a ↦ ⟦ ρ ↦ ∅ ⟧, ρ ↦ ∅ ⟧, ρ ↦ ∅ ⟧` — the `01-02` is gone. The same happens for a bound `ρ` carrying data (`ρ ↦ ⟦ Δ ⤍ … ⟧` → `<o name="ρ">hex</o>` → `ρ ↦ ⟦ ρ ↦ ∅ ⟧`), which was the very shape #1036 was about: the λ-name half round-trips correctly now, but the new `xmir-printing-packs/lambda-and-rho.yaml` pack covers printing only, not re-parsing, so the read side stayed unnoticed.

Suggested fix: in the no-`@base` branch of `xmirToFormationBinding`, when the element carries non-whitespace text content, add a `BiDelta (bytesToBts text)` binding to the formation — mirroring the application-argument path that already parses text content via `bytesToBts` (`src/XMIR.hs:485`) — and pin it with a print-then-parse round-trip pack.

yegor256:
The reader now picks up the text content of an `<o>` element that has no
`@base` and turns it into a Δ binding, so `<o name="a">01-02</o>` comes back as
`a ↦ ⟦ Δ ⤍ 01-02, ρ ↦ ∅ ⟧` instead of an empty formation.

Our own printer writes Δ that way, so the round-trip was losing bytes: the
example from the issue, `⟦ k ↦ ⟦ a ↦ ⟦ Δ ⤍ 01-02 ⟧, ρ ↦ ∅ ⟧ ⟧`, printed to XMIR
and read back came out without the `01-02`. Same for a bound `ρ` carrying data.
The `lambda-and-rho.yaml` pack only checked printing, so nothing caught it.

There were two places building a formation out of an element — one for named
bindings, one for a formation reached through a dispatch such as
`⟦ Δ ⤍ 01-02 ⟧.plus` — and both dropped the text. They now share a single
`xmirToFormation`.

Covered by two new parsing packs and by print-then-parse assertions next to the
existing round-trip test.

Closes #1045

### DIFF
--- a/src/XMIR.hs
+++ b/src/XMIR.hs
@@ -393,13 +393,12 @@ xmirToFormationBinding cur fqn
   | not (hasAttr "name" cur) = throwIO (InvalidXMIRFormat "Formation children must have @name attribute" cur)
   | not (hasAttr "base" cur) = do
       name <- getAttr "name" cur
-      bds <- mapM (`xmirToFormationBinding` (name : fqn)) (cur C.$/ C.element (toName "o")) >>= uniqueBindings'
       case name of
         "λ" -> BiLambda . Function <$> lambdaFunction
         ('α' : _) -> throwIO (InvalidXMIRFormat "Formation child @name can't start with α" cur)
-        "φ" -> pure (BiTau AtPhi (ExFormation (withVoidRho bds)))
-        "ρ" -> pure (BiTau AtRho (ExFormation (withVoidRho bds)))
-        _ -> pure (BiTau (AtLabel (T.pack name)) (ExFormation (withVoidRho bds)))
+        "φ" -> BiTau AtPhi <$> formation name
+        "ρ" -> BiTau AtRho <$> formation name
+        _ -> BiTau (AtLabel (T.pack name)) <$> formation name
   | otherwise = do
       name <- getAttr "name" cur
       base <- getAttr "base" cur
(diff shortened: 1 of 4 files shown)


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any part of it.

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1700](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1055, opened 2026-09-01
No machine-readable protocol of atom evaluations in dataize


yegor256:
When `dataize` (or `rewrite`) applies an atom through `_evaluate`
  (`src/Dataize.hs:531`), the function name, its argument formation, and
  the term it returns are all in hand at that moment, yet no output mode
  exposes them as individual records.
The closest thing today is `--sequence --headers`, which prints every
  intermediate expression of the whole document with a rule-name header,
  so a caller must diff consecutive multi-hundred-token dumps just to
  learn that `L_number_plus` fired and produced
  `Φ.number(as-bytes ↦ Φ.bytes(...))`.

This blocks tools that drive phino as a subprocess.
The EO compiler's build-time lowering wants to learn, per evaluation,
  which atom fired and what term came back (for instance
  `L_number_gt → Φ.true`), so that it stops keeping its own copy of the
  atom name table and each atom's result forma on the Java side;
  scraping `--sequence` output is not workable as an API.

A small addition would cover it: an `--evaluations FILE` option next to
  `--steps-dir` (`src/CLI/Parsers.hs:203`) that appends one line per
  successful `_evaluate` application — the λ function name, the argument
  bindings, and the resulting expression, rendered in the syntax chosen
  by `--output`.
Defaults stay unchanged when the flag is absent.

yegor256:
@maxonfjvipon could you please take a look at this one when you have a moment? Happy to clarify if anything in the request is unclear.

yegor256:
@rultor release, tag is `0.0.113`

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+1772](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1062, opened 2026-09-02
feat(dataize): #1060 partial evaluation with --partial

### LINKED ISSUE #1060: `dataize` Unexpectedly Aborts Instead of Handling Unknown Atoms Properly
When `dataize` meets an atom it cannot execute, it aborts the whole run:
  the catch-all in `src/Dataize.hs:516` throws
  `userError "Atom '%s' does not exist"`.
On 0.0.113 an expression holding a placeholder formation
  `⟦ λ ⤍ Sym_arg_0 ⟧` behaves like this: `dataize` first evaluates the
  all-literal subterms for real — `--evaluations` (#1055) correctly records,
  say, `L_number_times` over `2` and `3` yielding `Φ.number(… 40-18-…)` —
  and then dies on the placeholder, discarding the partially reduced
  expression and reporting nothing about where it stuck.
`rewrite --normalize` is the mirror image: it exits 0 and parks the stuck
  saturated atoms silently in the normal form, but executes no atoms at
  all, so nothing is evaluated or recorded.

This blocks a caller that morphs an expression whose data inputs were
  replaced by such placeholders on purpose: the EO compiler's build-time
  lowering (objectionary/eo#8137) wants to evaluate everything decided by
  literals, read the `--evaluations` protocol, and keep the residue with
  the stuck applications intact for the next iteration.
Today no phino mode produces that residue: one command throws it away,
  the other never evaluates.

The smallest change seems to be a `dataize` flag, `--partial`
  (dataization becomes partial evaluation: what the known inputs decide
  is computed, the rest survives as the residual program),
  under which the catch-all at `src/Dataize.hs:516` — and the same
  condition for a known atom whose input dataization reaches a
  placeholder — leaves the application in place as a normal-form subterm
  instead of throwing, evaluation continues elsewhere, and the run ends
  successfully printing the residual expression through the existing
  result output.
Optionally each parked site could also land in the `--evaluations` file
  as a record with no result term, mirroring the #1055 shape, so the
  caller learns the stuck sites without parsing the residue.

yegor256:
Closes #1060

## What

`dataize` used to abort on an atom it cannot execute: the catch-all in `atom` threw `Atom '…' does not exist`, discarding the partially reduced expression and reporting nothing about where it got stuck. A caller that replaces data inputs with placeholder formations such as `⟦ λ ⤍ Sym_arg_0 ⟧` on purpose (the EO compiler's build-time lowering) had no phino mode that evaluates what literals decide and keeps the rest.

This adds `dataize --partial`: dataization becomes partial evaluation. What the known inputs decide is computed; the rest survives as the residual program. An atom that cannot fire (its λ function is unknown, or one of its inputs reaches such an atom) stays in place as a normal-form subterm, the run exits 0 and prints the residual program through the usual result output (`--sweet`, `--hide-rho`, `--focus`, `--sequence` all apply). Without the flag the run fails exactly as before, with the same message.

```
$ phino dataize --partial --sweet --hide-rho partial.phi   # φ ↦ 2.times(3).plus(⟦ λ ⤍ Sym_arg_0 ⟧)
⟦ x ↦ ⟦ λ ⤍ Sym_arg_0 ⟧, λ ⤍ L_number_plus ⟧
```

Each stuck site also lands in the `--evaluations` file (#1055) as a record with the first two fields only, no result term, so the caller learns the stuck sites without parsing the residual program. The inner unknown atom is reported before the known atom whose input reached it, matching the nested order of ordinary firings:

```
L_number_times	⟦ x ↦ 3 ⟧	6
Sym_arg_0	⟦⟧
L_number_plus	⟦ x ↦ ⟦ λ ⤍ Sym_arg_0 ⟧ ⟧
```

## How

* `atom`'s catch-all throws a typed `Stuck` signal instead of a `userError`. Each frame of the 𝕄/𝔻 spine (`morph`, `dataize'`) wraps its body in `parking`, which turns the first `Stuck` it sees into `StuckAt` carrying the frame's derivation chain; outer frames let `StuckAt` pass, since their chains are prefixes. Side-computations that run on a chain of their own (`_dataize` inside atoms, `_morph` premises) strip the chain back off with `unparked`, so the spine frame around them attaches the right one. The head of the resulting chain is the residual program.
* `dataize` now returns an `Outcome` (`Dataized Bytes | Residual Expression`) with the chain; without `_partial`, `StuckAt` is rethrown and shows the old message.
* `Evaluation._result` becomes `Maybe Expression`; `_evaluate` reports a stuck firing with `Nothing` under `_partial`, then rethrows.
* CLI: `--partial` flag on `dataize`, `printFocused` helper shared with `printRewrittens`.

## Notes

Evaluation stays demand-driven, as the calculus prescribes: an argument nothing asked for before the run got stuck (e.g. `2.times(3)` in `⟦ λ ⤍ Sym ⟧.plus(2.times(3))`) is left in normal form in the residual program for the next iteration rather than evaluated speculatively. A wrong operand still yields ⊥ and stops the run as before; only an atom that cannot fire is left in place.

Tests: 7 new unit cases in `DataizeSpec`, 7 new CLI cases in `CLISpec`; full suite is 2169 examples, 0 failures with `-Werror`. README documents the flag.

https://claude.ai/code/session_01Br5aAMQXJkacBohfGsensK

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,13 +125,13 @@ works with, since one record must fit into one line.
 An atom that cannot fire fails the run: its λ function is unknown to phino,
 or one of its inputs reaches such an atom. This is what happens when a
 data input is replaced on purpose by a placeholder formation, such as
-`⟦ λ ⤍ Sym_arg_0 ⟧`. With `--park-stuck`, the run ends successfully
-instead: the stuck application stays in place as a normal-form subterm,
-everything the calculus demanded before it stays evaluated, and the
-residual expression is printed in place of the bytes:
+`⟦ λ ⤍ Sym_arg_0 ⟧`. With `--partial`, dataization becomes partial
+evaluation instead: what the known inputs decide is computed, the rest
+survives as the residual program, which is printed in place of the bytes,
+and the run ends successfully:
 
 ```bash
-$ cat stuck.phi
+$ cat partial.phi
 ⟦
   bytes(data) ↦ ⟦ φ ↦ data ⟧,
   number(as-bytes) ↦ ⟦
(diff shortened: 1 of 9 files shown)


yegor256:
CI is green and this implements #1060 with solid test coverage, but it touches the core dataization engine's control flow, so it warrants a careful read rather than an automatic merge. Assigning to @yegor256.

---

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (394 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1971](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1068, opened 2026-09-04
fix(sugar): #1067 unescape a string literal before desugaring it

### LINKED ISSUE #1067: Desugaring a string escapes its bytes twice (src/Sugar.hs:183)
The sweet printer escapes the bytes of a `Φ.string` term and the desugarer never unescapes them, so a text loses a backslash on every round trip. `src/CST.hs:343` turns a `DataString` into `EX_STRING (btsToStr bts)`, and `btsToStr` (`src/Bytes.hs:187`) escapes `"`, `\`, `\n`, `\t` and every unprintable character. `src/Sugar.hs:183` then desugars that same `EX_STRING` back with `strToBts str`, and `strToBts` (`src/Bytes.hs:152`) takes the characters as they stand. Feeding `⟦ φ ↦ Φ.string(as-bytes ↦ Φ.bytes(data ↦ ⟦ Δ ⤍ 65-0A-65 ⟧)) ⟧` to `phino merge` prints `Δ ⤍ 65-5C-6E-65`: the newline became a backslash and an `n`. A quote grows to `5C-22`, a backslash doubles to `5C-5C`, and the text keeps growing with every further pass.

This is wrong because the printer and the parser disagree. `quotedStr` (`src/Parser.hs:196`) does unescape what it reads, so the sugar `"e\ne"` written by hand and the desugared form of the very same value end up as different bytes, and a program that merely passes through phino no longer holds the text it started with. The number path next to it is symmetric — `numToBts` inverts `btsToNum` at `src/Sugar.hs:176` — which is why only strings drift.

The fix is one inverse function: add an `unescapeStr` to `src/Bytes.hs` that undoes what `escapeStr` does, and call `strToBts (unescapeStr str)` at `src/Sugar.hs:183`, so the desugarer reads an escaped literal the same way the parser already does.

yegor256:
Closes #1067.

The sweet printer escapes the bytes of a `Φ.string` (`CST.hs:343` renders
a `DataString` as `EX_STRING (btsToStr bts)`, and `btsToStr` escapes `"`,
`\`, `\n`, `\t` and every unprintable character), but the desugarer at
`Sugar.hs:183` fed that escaped literal straight back to `strToBts`,
which takes the characters as they stand. A text therefore grew on every
round trip: a newline became `5C-6E`, a quote became `5C-22` and a
backslash doubled.

## Changes

- `Bytes.unescapeStr` — the inverse of the escaping `btsToStr` applies.
  It undoes `\"`, `\\`, `\n`, `\t` and `\xNN`; a backslash that starts no
  escape `btsToStr` can produce is kept as it stands, together with the
  character behind it.
- `Sugar.toSalty` now calls `strToBts (unescapeStr str)`, so the
  desugarer reads an escaped literal the same way `quotedStr`
  (`Parser.hs:196`) already does. The number path next to it was already
  symmetric and is untouched.

Before:

```
$ echo '⟦ φ ↦ Φ.string(as-bytes ↦ Φ.bytes(data ↦ ⟦ Δ ⤍ 65-0A-65, ρ ↦ ∅ ⟧)), ρ ↦ ∅ ⟧' | phino rewrite --flat
⟦ φ ↦ Φ.string( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ 65-5C-6E-65, ρ ↦ ∅ ⟧ ) ), ρ ↦ ∅ ⟧
```

After: the bytes come back as `65-0A-65`.

## Tests

- `BytesSpec` — 12 `unescapeStr` cases plus an 8-case
  `btsToStr`/`unescapeStr` round trip.
- `SugarSpec` — three `toSalty EX_STRING` cases covering a newline, a
  quote with a backslash, and a hex escape.
- `CLISpec` — the end-to-end case from the issue.

All five new tests fail on `master` (the CLI one with exactly the
`65-5C-6E-65` from the issue) and pass with this change. Full suite:
2193 examples, 0 failures, built with `-Werror`; `fourmolu --mode check`
and `hlint` are clean.

https://claude.ai/code/session_01JzXTPrXi7t69nQ2y9pBRAt

---
_Generated by [Claude Code](https://claude.ai/code/session_01JzXTPrXi7t69nQ2y9pBRAt)_

### DIFF
--- a/src/Bytes.hs
+++ b/src/Bytes.hs
@@ -12,6 +12,7 @@ module Bytes
   , strToBts
   , bytesToBts
   , btsToStr
+  , unescapeStr
   , btsToNum
   , btsToUnescapedStr
   , btsAnd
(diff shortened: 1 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2127](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1096, opened 2026-09-06
fix(filter): apply every fqn in include, not only the first one (#1088)

### LINKED ISSUE #1088: Filter.include silently keeps only the first locator and drops the rest (asymmetry with exclude)
`Filter.include` silently keeps only the first locator and drops the rest (asymmetry with `exclude`)

## Problem

`src/Filter.hs:57-60`:

```haskell
include :: [Rewritten] -> [Expression] -> [Rewritten]
include [] _ = []
include rs [] = rs
include ((expr, maybeRule) : rest) (fqn : _) = (include' expr fqn, maybeRule) : include rest [fqn]
```

`include` applies **only the first locator** (`fqn : _`) to the first expression, then recurses with `[fqn]` — every subsequent expression also gets only that same first locator, and all remaining locators are dropped.

Compare `exclude` (`Filter.hs:29-32`), which honestly applies **all** locators to every expression:

```haskell
exclude :: [Rewritten] -> [Expression] -> [Rewritten]
exclude ((expr, maybeRule) : rest) exprs = (exclude' expr exprs, maybeRule) : exclude rest exprs
```

## Repro

An API call like `Filter.include rewrittens [locator1, locator2]` keeps only the subtree under `locator1`; `locator2` is never applied. If `locator1` selects nothing but `locator2` would, the result is `⟦ ρ ↦ ∅ ⟧` (the `def` empty formation from `include'`, `Filter.hs:34-41`) — the user silently gets the wrong subtree.

## Why it is masked today

The CLI limits `--show` to one occurrence (`Runners.hs:77`: "The option --show can be used only once"), so the bug is not reachable through the CLI. But `Filter.include` is exported and is a legitimate library API, and the asymmetry with `exclude` is a silent trap for any caller passing several locators.

## Expected

Either make `include` apply all locators (like `exclude`), or document explicitly that `include` handles a single locator and reject a non-singleton list. The first is the consistent choice.

## Impact

Silent data loss for library users; the function's signature (`[Expression]`) implies it accepts several locators, but it uses only the first.

VasilevNStas:
Fixes #1088.

## Problem

`Filter.include` (`src/Filter.hs:57-60`) applied **only the first** locator and
silently dropped the rest:

```haskell
include ((expr, maybeRule) : rest) (fqn : _) = (include' expr fqn, maybeRule) : include rest [fqn]
```

`exclude` (`Filter.hs:29-32`) honestly applies all locators to every expression;
`include` did not, so a library call like `include rs [locator1, locator2]` kept
only the subtree under `locator1`. If `locator1` matched nothing but `locator2`
would, the result was the empty `⟦ ρ ↦ ∅ ⟧` — a silent wrong answer.

## Fix

- `src/Filter.hs` — `include'` now takes a list of fqns, applies each one, and
  merges the matching subtrees into a single formation (`withVoidRho`, with the
  per-subtree `ρ` removed so the result stays a valid formation).
- `include` passes the whole fqn list through, mirroring `exclude`.

The CLI is unaffected: `--show` is still limited to one occurrence, so the
single-fqn behavior is unchanged; only the library API stops dropping fqns.

## Changes

- `src/Filter.hs` — `include' :: Expression -> [Expression] -> Expression`
  applying every fqn and merging matches; `include` forwards the full list.
- `test/FilterSpec.hs` — new test: `include` with `Q.x` and `Q.y` keeps both
  subtrees (`[[ x -> ?, y -> ? ]]`); existing multi-element test renamed to
  reflect the new semantics.

## Tests

- `test/FilterSpec.hs` — both existing include fallbacks and the new
  merge-two-fqns case pass; full `CLI` suite (which drives `include` via
  `--show`) is green.

### DIFF
--- a/src/Filter.hs
+++ b/src/Filter.hs
@@ -34,7 +34,7 @@ exclude ((expr, maybeRule) : rest) exprs = (exclude' expr exprs, maybeRule) : ex
 
 include' :: Expression -> [Expression] -> Expression
 include' expr fqns = case mapMaybe pick fqns of
-  [] -> ExFormation [BiVoid AtRho]
+  [] -> def
   forms -> mergeForms forms
   where
     def :: Expression
--- a/src/Filter.hs
+++ b/src/Filter.hs
@@ -4,6 +4,7 @@
 module Filter (include, exclude) where
 
 import AST
+import Data.Maybe (mapMaybe)
 import Misc
 import Rewriter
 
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@yegor256 
plz review

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (37) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1718](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1098, opened 2026-09-06
fix(cli): reject --in-place together with a non-phi output format (#1089)

### LINKED ISSUE #1089: --in-place with --output=latex or --output=xmir is not validated and overwrites the source file with a different format
`--in-place` with `--output=latex` or `--output=xmir` is not validated and overwrites the source file with a different format

## Problem

`src/CLI/Runners.hs:70-84` (`runRewrite` validation) checks combinations of `--in-place`/`--target`/`--update`, but never checks that the **output format** is compatible with writing back into the source file. `validateLatexOptions` (`Validators.hs:39-48`) only *allows* the latex-only options; it does not forbid `--in-place` together with `LATEX` or `XMIR`.

So:

```
$ phino rewrite --in-place --output=latex my.phi
```

reads `my.phi`, rewrites it, renders the result as **LaTeX**, and writes the LaTeX back into `my.phi` — the phi source file is silently replaced by TeX markup. Likewise `--output=xmir` turns the source into XML.

For an XMIR input (`--input=xmir --in-place`), the passthrough additionally re-stamps `author`/`revision`/`dob`/`time` (see the metadata-loss report), rewriting the file's provenance on every run.

## Repro (verified)

```
$ echo '⟦ x ↦ 1, ρ ↦ ∅ ⟧' > my.phi
$ phino rewrite --in-place --output=latex my.phi
$ cat my.phi     # now contains LaTeX, the original phi is gone
```

No error is raised; the file is silently corrupted from the user's perspective.

## Call-site inventory

- `Runners.hs:103-116` (`output`) — writes `expr` back to the input file whenever `_inPlace` is set, with no format check.
- `Runners.hs:70-84` (`validateOpts`) — only `--in-place`/`--target`/`--update` interactions.
- `Validators.hs:39-48` (`validateLatexOptions`) — allows latex options, doesn't reject `--in-place` for latex.

## Expected

- Forbid `--in-place` when `--output` is not `PHI` (or, for an `--input=xmir` source, not the same format), with a clear message: "The option --in-place can only be used together with --output=phi".
- Alternatively, refuse to write back a format different from the input format.

## Impact

Silent source-file corruption via a plausible-looking option combination; data loss with no error.

VasilevNStas:
Fixes #1089.

## Problem

`runRewrite`'s validation (`src/CLI/Runners.hs:70-84`) checked `--in-place` only
against `--target`/`--update`/`--input`, never against the **output format**.
So:

```
$ phino rewrite --in-place --output=latex my.phi
```

read `my.phi`, rendered the result as LaTeX, and wrote it back into `my.phi` —
silently replacing the phi source with TeX markup (same for `--output=xmir`).
The source file was corrupted with no error.

## Fix

`src/CLI/Runners.hs` — validation now rejects `--in-place` unless the output
format is `PHI`:

```
The option --in-place can only be used together with --output=phi
```

`--in-place` is about rewriting a phi source file in place, so a non-phi output
is a user mistake; failing early (before any file I/O) avoids corrupting the
source.

## Changes

- `src/CLI/Runners.hs` — new validation rule in `validateOpts` (`runRewrite`).
- `test/CLISpec.hs` — new test: `--in-place --output=latex` fails with the
  message and leaves the file untouched.

## Tests

- `test/CLISpec.hs` — 105 CLI/rewriting examples, 0 failures (verified
  `--in-place --output=phi` still works and overwrites the file).

### DIFF
--- a/src/CLI/Runners.hs
+++ b/src/CLI/Runners.hs
@@ -71,6 +71,7 @@ runRewrite OptsRewrite{..} = do
     validateOpts = do
       when (_inPlace && isNothing _inputFile) (invalidCLIArguments "The option --in-place requires an input file")
       when (_inPlace && isJust _targetFile) (invalidCLIArguments "The options --in-place and --target cannot be used together")
+      when (_inPlace && _outputFormat /= PHI) (invalidCLIArguments "The option --in-place can only be used together with --output=phi")
       when (_update && _inPlace) (invalidCLIArguments "The options --update and --in-place cannot be used together")
       when (_update && isNothing _targetFile) (invalidCLIArguments "The option --update requires --target")
       when (_update && isNothing _inputFile) (invalidCLIArguments "The option --update requires an input file")
--- a/test/CLISpec.hs
+++ b/test/CLISpec.hs
@@ -226,6 +226,14 @@ spec = do
             ["rewrite", "--in-place", "--target=output.phi", path]
             ["--in-place and --target cannot be used together"]
 
+      it "fails when --in-place is used with a non-phi output format" $
+        withTempFile "inplaceXXXXXX.phi" $ \(path, h) -> do
+          hPutStr h "[[ ]]"
+          hClose h
+          testCLIFailed
+            ["rewrite", "--in-place", "--output=latex", path]
+            ["--in-place can only be used together with --output=phi"]
+
       forM_
         [ ("when --update is used without --target", "[[ ]]", ["rewrite", "--update"], ["--update requires --target"])
         ,


VasilevNStas:
@yegor256 
plz review

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (9) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1742](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1099, opened 2026-09-06
fix(yaml): require non-empty and/or conditions and reject fractional literals (#1090)

### LINKED ISSUE #1090: Small engine inconsistencies: rewrite does not seed initial expr into seen-set; Condition and()/or() arity; YAML Number literal silently rounds
Small engine inconsistencies: `rewrite` does not seed the initial expression into the seen-set; `Condition` parser `and()`/`or()` require ≥1 arg while `_and [] = [subst]`; YAML `Number` literal silently rounds fractions

Three small, low-severity inconsistencies found during a full-repo read.

## 1. `rewrite` does not insert the initial expression into `_unique`

`src/Rewriter.hs:236-239`:

```haskell
rewrite expr rules ctx@RewriteContext{..} = do
  (rewrittens, exceeded) <- _rewrite ((expr, Nothing) :| [], Map.empty, False) 0
  ...
```

The seen-set starts empty, so the **initial** expression is not recorded. A rewriting cycle that returns to the *original* term is therefore detected one step later than it could be (`a→b→a` is seen as `b→a` fresh, then `a→b` again). Not a correctness bug, but an extra step and an asymmetry with `seenInsert` usage inside `_rewrite`.

Fix: `seenInsert (hashExpression expr) expr Map.empty` before the first step.

## 2. `Condition` parser `and()`/`or()` require ≥1 argument, but `_and [] = [subst]`

`src/Condition.hs:86-94` parses `and(...)`/`or(...)` with `sepBy1`, so an empty list never parses from a `--when` string. But the YAML path (`Yaml.hs` `v .: "and"`/`.: "or"`) can produce `And []`/`Or []`, and `_and []` returns `[subst]` (true) in `Rule.hs:75-89` while `_or []` returns `[]` (false). So an empty `and`/`or` in a YAML rule evaluates to a value that the string parser cannot even express — the two parsers disagree about the language.

Fix: pick one semantics (e.g. `and`/`or` with at least one element enforced in `FromJSON` too, or allow empty in the string parser).

## 3. YAML `Number` literal silently rounds fractions

`src/Yaml.hs:80`:

```haskell
Number num -> pure (Literal (round num))
```

A fractional literal in a YAML condition (e.g. `gt: [2.5, 1]`) is silently rounded with `round` (banker's rounding: `round 2.5 = 2`). The string parser (`Condition.hs`, `L.decimal`) accepts only integers. So a YAML author writing `2.5` gets `2` without any warning, while the same condition via `--when` would be a parse error.

Fix: reject fractions in YAML (like the string parser does), or support them consistently in both.

## Expected

Each of these is a small consistency fix: seed the initial expression, align the `and`/`or` arity between the two condition parsers, and make `Number` literal handling consistent (no silent rounding).

VasilevNStas:
Fixes #1090 (the two implemented parts; see note below).

## Problem

Two small inconsistencies between the YAML condition parser and the string
`--when` parser:

1. **`and`/`or` arity.** The string parser (`Condition.hs`) parses `and(...)`/
   `or(...)` with `sepBy1`, so an empty list never parses. But the YAML path
   (`Yaml.hs` `v .: "and"`/`.: "or"`) accepted `and: []`/`or: []`, producing an
   `And []`/`Or []` that the string parser cannot express — the two parsers
   disagreed about the language.
2. **Fractional `Number` literals.** `Yaml.hs:80` did `Literal (round num)`,
   silently rounding a fraction with banker's rounding (e.g. `2.5` → `2`)
   without any warning, while the string parser (`Condition.hs` `L.decimal`)
   accepts only integers. A YAML author writing `2.5` got `2` silently.

## Fix

- `src/Yaml.hs` — `and`/`or` now require at least one element (matching the
  string parser); a fractional `Number` literal is rejected with a descriptive
  message instead of being rounded.

## Changes

- `src/Yaml.hs` — non-empty checks for `And`/`Or`; integer-only guard for
  `Number` literals.
- `test/YamlSpec.hs` — tests for `and: []`, `or: []`, `2.5` (rejected) and a
  whole `5` (accepted).

## Tests

- `test/YamlSpec.hs` — 35 examples, 0 failures; `ConditionSpec`, `RewriterSpec`
  and `CLI/rewriting` suites all green.

## Note

The third item in #1090 (seeding the initial expression into the rewriter's
seen-set) was considered but **rejected**: it changes rewrite semantics — a
legitimate return to the initial expression under `--sequence`/`--max-cycles`
would be flagged as a loop, breaking `prints many expressions with --sequence`
and the `--headers` tests. Left as-is.

### DIFF
--- a/src/Yaml.hs
+++ b/src/Yaml.hs
@@ -77,7 +77,9 @@ instance FromJSON Number where
         [ Length <$> o .: "length"
         , Domain <$> o .: "domain"
         ]
-    Number num -> pure (Literal (round num))
+    Number num
+      | toRational (round num :: Integer) == toRational num -> pure (Literal (round num))
+      | otherwise -> fail (printf "Expected an integer, got a fractional number %s" (show num))
     String txt -> case parseIndex (unpack txt) of
       Right mt -> pure (MetaIndex mt)
       Left err -> fail err
@@ -99,8 +101,16 @@ instance FromJSON Condition where
       ( \v -> do
           validateYamlObject v ["and", "or", "not", "nf", "absolute", "eq", "gt", "in", "matches", "part-of", "disjoint", "formation"]
           asum
-            [ And <$> v .: "and"
-            , Or <$> v .: "or"
+            [ do
+                conds <- v .: "and"
+                if null conds
+                  then fail "The 'and' condition requires at least one element"
+                  else pure (And conds)
+            , do
+                conds <- v .: "or"
+                if null conds
+                  then fail "The 'or' condition requires at least one element"
+                  else pure (Or conds)
             , Not <$> v .: "not"
             , NF <$> v .: "nf"
             , Absolute <$> v .: "absolute"
--- a/test/YamlSpec.hs
+++ b/test/YamlSpec.hs
@@ -156,3 +156,19 @@ spec = do
   describe "rejects a numerable expression that is neither an object, a number nor an index meta" $
     it "fails on a bare boolean" $
       (decodeYaml' "true" :: Either Yaml.ParseException Number) `shouldSatisfy` isLeft
+
+  describe "parses a literal number" $
+    it "accepts a whole number" $
+      (decodeYaml' "5" :: Either Yaml.ParseException Number) `shouldSatisfy` (not . isLeft)
+
+  describe "rejects a fractional literal number" $
+    it "fails on a fraction instead of silently rounding it" $
+      (decodeYaml' "2.5" :: Either Yaml.ParseException Number) `shouldSatisfy` isLeft
+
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 
plz review

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1754](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1100, opened 2026-09-06
fix(cli): stop leaking HasCallStack backtraces and duplicated optparse errors (#1077)

### LINKED ISSUE #1077: CLI error handler prints a GHC backtrace on every failure and duplicates optparse errors
CLI error handler prints a GHC backtrace on every failure and duplicates optparse errors

## Problem

`src/CLI.hs:32-37`:

```haskell
handler :: SomeException -> IO ()
handler e = case fromException e of
  Just ExitSuccess -> pure () -- prevent printing error on --version etc.
  _ -> do
    logError (displayException e)
    exitFailure
```

On GHC 9.10 `displayException` on a `SomeException` **includes a `HasCallStack` backtrace** (`collectBacktraces, called at libraries/ghc-internal/...`). So every failure — even benign, user-caused ones — prints a backtrace block to stderr. The handler also catches `ExitFailure` produced by optparse-applicative parse errors, so a bad option prints **twice**: once by optparse's own `ParserFailure` output, then again as `[ERROR]: ExitFailure 1` + backtrace.

## Repro (verified)

```
$ phino rewrite --badopt
(some optparse error text)
[ERROR]: ExitFailure 1
HasCallStack backtrace:
  collectBacktraces, called at libraries/ghc-internal/src/GHC/Internal/Exception.hs:169:13
  toExceptionWithBacktrace, called at ...
  ...
```

A non-existent input file likewise:

```
$ phino rewrite missing.phi
[ERROR]: File 'missing.phi' does not exist
HasCallStack backtrace:
  collectBacktraces, called at ...
```

The backtrace is appended to every single error path in the CLI, adding noise and making error output longer and less friendly. It does not help debugging because the failures are user input errors, not internal exceptions.

## Call-site inventory

- `CLI.hs:32-37` — the single `handler` for all `SomeException`s.
- `CLI.hs:22` — `handleParseResult (execParserPure ...)` throws `ParserFailure`/`ExitFailure` for bad options; the handler re-logs `ExitFailure 1`.
- `CLI.hs:51` — `checkPin` throws `VersionMismatch`, also routed through the same handler.

## Expected

- Catch typed exceptions where feasible and print a concise `show`/message without the internal backtrace.
- Do not re-print optparse parse errors as a second `[ERROR]: ExitFailure 1` (optparse already wrote its own message).
- Optionally, guard against `HasCallStack` noise globally (e.g. only print the backtrace under a debug flag).

VasilevNStas:
Fixes #1077.

## Problem

`src/CLI.hs:32-37` handled every failure via `displayException` on a
`SomeException`. On GHC 9.10 `displayException` includes a `HasCallStack`
backtrace, so every error (bad option, missing file, invalid arguments, ...)
printed a backtrace block. optparse-applicative parse errors were printed
twice: once by optparse, then again as `[ERROR]: ExitFailure 1` + backtrace.

Repro (before):
```
$ phino rewrite --badopt
Invalid option `--badopt'
...
[ERROR]: ExitFailure 1
HasCallStack backtrace:
  collectBacktraces, called at libraries/ghc-internal/...
```

## Fix

- `src/CLI.hs` — handle the `ParserResult` from `execParserPure` directly:
  - `Failure` with `ExitSuccess` (`--version`/`--help`) prints the message
    as-is;
  - `Failure` with a failing exit code prints one `[ERROR]: <first line>`
    line (optparse's message) and exits with that code — no backtrace, no
    duplicated `ExitFailure 1`;
  - `CompletionInvoked` falls back to `handleParseResult`.
- The top-level `handler` uses `show` (no backtrace) for everything else, and
  silently exits for `ExitFailure` (already logged).

## Changes

- `src/CLI.hs` — explicit `ParserResult` handling; `show` instead of
  `displayException`; `ExitFailure` short-circuit in the handler.
- `test/CLISpec.hs` — two tests asserting errors do not contain
  `HasCallStack backtrace` or `ExitFailure 1`, for both a validation error and
  an optparse error.

## Tests

- `test/CLISpec.hs` — 208 examples, 0 failures. Verified manually: bad option,
  missing file, invalid `--in-place`, `--version`/`--help` all clean.
- Note: the 13 `LocatorSpec` failures (GHC backtrace inside `displayException`
  of `Locator` exceptions) are a separate, module-level issue; this PR covers
  the CLI handling only.

### DIFF
--- a/src/CLI.hs
+++ b/src/CLI.hs
@@ -10,16 +10,32 @@ module CLI (runCLI) where
 import CLI.Parsers
 import CLI.Runners
 import CLI.Types
-import Control.Exception.Base (Exception (displayException), SomeException, fromException, handle, throwIO)
+import Control.Exception.Base (SomeException, fromException, handle, throwIO)
 import Data.Version (showVersion)
 import Logger
 import Options.Applicative
 import Paths_phino (version)
-import System.Exit (ExitCode (..), exitFailure)
+import System.Exit (ExitCode (..), exitFailure, exitWith)
+import System.IO (hPutStrLn, stderr)
 
 runCLI :: [String] -> IO ()
 runCLI args = handle handler $ do
-  CliArgs{_pin, _command} <- handleParseResult (execParserPure defaultPrefs parserInfo args)
+  let parsed = execParserPure defaultPrefs parserInfo args
+  CliArgs{_pin, _command} <- case parsed of
+    Success opts -> pure opts
+    Failure failure -> do
+      let (msg, code) = renderFailure failure "phino"
+      case code of
+        ExitSuccess -> do
+          putStrLn msg -- --version/--help output as-is
+          exitWith code
+        _ -> do
+          -- Keep the full optparse message (including the Usage/synopsis
+          -- block that follows a parse error), but without the GHC
+          -- HasCallStack backtrace; prefix just the first line with [ERROR]:.
+          hPutStrLn stderr (prefixFirstLine "[ERROR]: " msg)
+          exitWith code
+    CompletionInvoked _ -> handleParseResult parsed
   checkPin _pin
   setLogger _command
   case _command of
@@ -30,11 +46,15 @@ runCLI args = handle handler $ do
     CmdMerge opts -> runMerge opts
     CmdMatch opts -> runMatch opts
   where
+    prefixFirstLine :: String -> String -> String
+    prefixFirstLine _ "" = "Failure"
+    prefixFirstLine prefix msg = prefix ++ msg
     handler :: SomeException -> IO ()
     handler e = case fromException e of
       Just ExitSuccess -> pure () -- prevent printing error on --version etc.
+      Just (ExitFailure _) -> exitFailure -- already logged by the Failure branch above
       _ -> do
-        logError (displayException e)
+        logError (show e)
         exitFailure
     setLogger :: Command -> IO ()
     setLogger cmd =
--- a/test/CLISpec.hs
+++ b/test/CLISpec.hs
@@ -250,6 +250,18 @@ spec = do
             ["rewrite", "--in-place", "--output=latex", path]
             ["--in-place can only be used together with --output=phi"]
 
+      it "does not leak a HasCallStack backtrace into errors" $ do
+        (out, _) <- withStdout (try (runCLI ["rewrite", "--in-place"]) :: IO (Either ExitCode ()))
+        out `shouldNotContain` "HasCallStack backtrace"
+        out `shouldNotContain` "ExitFailure 1"
+        out `shouldContain` "[ERROR]:"
+
+      it "prints optparse errors once, without a backtrace" $ do
+        (out, _) <- withStdout (try (runCLI ["rewrite", "--badopt"]) :: IO (Either ExitCode ()))
+        out `shouldNotContain` "HasCallStack backtrace"
+        out `shouldNotContain` "ExitFailure 1"
+        out `shouldContain` "[ERROR]:"
+
       forM_
         [ ("when --update is used without --target", "[[ ]]", ["rewrite", "--update"], ["--update requires --target"])
         ,


VasilevNStas:
@yegor256 
plz review

yegor256:
Holding off on merging this one. The new Failure branch in CLI.hs only prints the first line of optparse's rendered message before exiting, which drops the usage/synopsis text that normally follows a parse error — that looks like it could be a real loss of information for users hitting a bad flag, not just a dedup of the backtrace. Also swapping `displayException e` for `show e` in the fallback handler is a behavior change worth a second look. Leaving this for manual review.

---

VasilevNStas:
Thanks for the review — both points addressed.

**1. First-line-only optparse message (usage dropped).** Fixed: the `Failure` branch now prints the **full** `renderFailure` message (including the `Usage:`/synopsis block) to stderr, with `[ERROR]:` prefixed only on the first line. So a bad flag shows the error plus usage, without the GHC backtrace. Verified: `phino rewrite --badopt` now prints

```
[ERROR]: Invalid option `--badopt'

Usage: phino [--pin VERSION] COMMAND
```

**2. `show e` vs `displayException e` in the fallback handler.** Kept `show e`, and it is deliberate: on GHC 9.10 `displayException` on a `SomeException` that was thrown via `throwIO` from a `HasCallStack` context appends the whole `HasCallStack backtrace:` block, which is exactly the noise this issue (#1077) is about. `show e` renders the same message for the typed exceptions (e.g. `CmdException`, `VersionMismatch`) but without the internal backtrace. I added a clarifying comment on the branch. If you would prefer `displayException` on the *unwrapped* exception type instead, I can do that too — just say the word.

Rebased onto current master (conflict in `test/CLISpec.hs` resolved; both the `--in-place` test from #1098 and the new backtrace tests are kept).

yegor256:
@maxonfjvipon please, check

maxonfjvipon (changes requested):
On the second of @yegor256's points, `show e` vs `displayException e`: checked,
and the reasoning holds. Every exception type in `src/` (`BuildException`,
`CmdException`, `ConditionException`, `DataizeException`, `FsException`,
`LocatorException`, `MergeException`, `ParserException`, `XMIRException`)
declares an explicit `instance Show` and only *derives* `Exception`, so
`displayException` already defaults to `show` for all of them. `show e` on the
`SomeException` therefore loses nothing but GHC's backtrace. No objection.

The first point is a different story: the fix is in, but nothing holds it in
place.

**1. The usage block is untested.** `test/CLISpec.hs` asserts only the absence
of `HasCallStack backtrace` and `ExitFailure 1` and the presence of
`[ERROR]:`. Not one assertion pins the `Usage:` line that the previous round
dropped — go back to first-line-only and CI stays green. That is the whole
regression; it needs `"Usage: phino"` in the expected output of the
`--badopt` test.

**2. Both new tests threw away the exit code.** They hand-roll
`withStdout (try (runCLI ...))` and bind `(out, _)`, so the code the new
`Failure` branch chooses is never checked — in a PR whose subject is exit
paths. `testCLIFailed` (`test/CLISpec.hs:121`) already asserts
`Left (ExitFailure 1)` together with the substrings; use it and keep the two
`shouldNotContain` lines on top of it.

**3. `prefixFirstLine` does not touch the first line.** It is `(prefix ++)`
with a name that promises otherwise, at one call site — inline it.
Its `prefixFirstLine _ "" = "Failure"` branch is speculative
(`renderFailure` does not answer an empty message) and, if it ever fired,
would print a bare `Failure` with no `[ERROR]:` prefix, contradicting the
branch it lives in. `CLAUDE.md`: inline single-use expressions, fail fast over
fail safe.

**4. `"[ERROR]: "` now lives in two places.** `Logger.logMessage`
(`src/Logger.hs:41`) builds that exact prefix for the same handle; the new
line re-implements it, so the two drift the day the logger's format changes.
`logError msg` writes the same thing — and brings `--log-lines` back, which
the raw `hPutStrLn` bypasses. If bypassing the limit is deliberate for usage
text, say so in the comment.

**5. `Just (ExitFailure _) -> exitFailure` discards the code the branch above
just picked.** `exitWith code` throws, the handler catches it and re-exits
with 1 regardless, so the `exitWith code` reads as if it preserves the code
while it cannot. optparse only ever answers `ExitFailure 1` today, so nothing
visible changes — which is exactly why it should be
`Just code@(ExitFailure _) -> exitWith code` before something else starts
exiting with 2.

One question rather than a request: `renderFailure failure "phino"` hardcodes
the program name where `handleParseResult` used `getProgName`. That reads like
an improvement (deterministic under the test suite, which would otherwise
print `spec`) — confirming it is intended, not incidental.

VasilevNStas:
Thanks @maxonfjvipon for the thorough review. All five points are addressed in `7974f4a8`:

**1. Usage block pinned.** The `--badopt` test now asserts `"Usage: phino"` in the output, so going back to first-line-only would fail CI.

**2. Exit code asserted.** Both tests now use `testCLIFailed` (which asserts `Left (ExitFailure 1)` together with the substrings); the two `shouldNotContain` lines stay on top of it.

**3. `prefixFirstLine` inlined.** The branch prints `"[ERROR]: " ++ msg` directly; the speculative `prefixFirstLine _ "" = "Failure"` case (which would have printed a bare `Failure` with no `[ERROR]:` prefix) is gone.

**4. `[ERROR]: ` prefix.** Kept the direct `hPutStrLn stderr` on purpose, with a comment now stating why: routing through `Logger.logError` would apply the `--log-lines` limit to the usage/synopsis text, and a truncated synopsis helps nobody. There is also a structural reason: `setLogger` runs only *after* parsing, so on a parse error the global logger IORef still holds whatever the previous run left behind (in the test suite that can be `--log-lines=1` or `NONE`), which would make the error output nondeterministic. Bypassing the logger here keeps this path deterministic.

**5. Exit code preserved.** `Just (ExitFailure _) -> exitFailure` → `Just code@(ExitFailure _) -> exitWith code`, so the code the `Failure` branch picked is no longer flattened to 1.

**Question (program name).** Yes, `"phino"` in `renderFailure failure "phino"` is intentional: under the test suite `getProgName` answers `spec`, so hardcoding makes the rendered usage deterministic. optparse uses the string only for rendering; nothing else reads the program name on this path.

`cabal build all --ghc-options=-Werror`, `hlint` and `fourmolu --mode check` pass; the CLI describe block (215 examples) is green.

VasilevNStas:
@maxonfjvipon 
plz review

maxonfjvipon (approved):
All five addressed in `7974f4a8`, and checked rather than taken on trust:

- `Usage: phino` is now in the expected output of the `--badopt` test, and
  both tests go through `testCLIFailed`, so the exit code is asserted again.
- `prefixFirstLine` is gone; the branch reads `hPutStrLn stderr ("[ERROR]: " ++ msg)`.
- `Just code@(ExitFailure _) -> exitWith code` no longer flattens the code.

On keeping the raw `hPutStrLn` over `logError` — your second reason is the
better one and settles it. `logger` is an `unsafePerformIO` IORef
(`src/Logger.hs:26-28`) and hspec runs the whole suite in one process, so on a
parse error — which happens before `setLogger` — the level and line limit are
whatever the previous test left in it. Routing this path through the logger
would make it order-dependent. The comment in the code gives the `--log-lines`
half of that; the global-state half is the part worth having there too, though
not worth another round.

Hardcoding `"phino"` in `renderFailure`: agreed, and it is what makes
`"Usage: phino"` assertable in the first place.

All 24 checks green on `7974f4a8`, no conflict with master.

@rultor merge

maxonfjvipon:
@rultor merge

maxonfjvipon:
@VasilevNStas try to update to fresh master and we'll try to merge again

VasilevNStas:
Rebased onto fresh master — the branch is now based on the latest `upstream/master` (which gained `--atoms` registry #1119, `morph` command #1115, #1118 and #1112). No conflicts; the CLI describe block (248 examples) is green, `hlint` and `fourmolu --mode check` pass.

Note: `cabal build all --ghc-options=-Werror` locally fails on a pre-existing redundant `Data.List (foldl')` import in `src/AST.hs:14` under GHC 9.10.3 — unrelated to this PR and not present on the CI toolchain (9.6.7/9.8.4). The rultor run just now built and passed the full suite on 9.6.7, so this is local-only.

Ready to merge again.

VasilevNStas:
@maxonfjvipon 
I`m fixed it

maxonfjvipon:
@rultor merge

VasilevNStas:
@maxonfjvipon 
I’ve fixed it. Please call @rultor so he can start the merge.

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2416](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+1437](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1101, opened 2026-09-06
fix(random): bound regenerate attempts so randomString cannot loop forever (#1081)

### LINKED ISSUE #1081: randomString can loop forever when the %d space (10000 values) is exhausted; uniqueness is process-global
`randomString` can loop forever when the `%d` space (10000 values) is exhausted; uniqueness is process-global, not per-run

## Problem

`src/Random.hs:18-20,37-49`:

```haskell
strings :: IORef (Set String)
{-# NOINLINE strings #-}
strings = unsafePerformIO (newIORef Set.empty)

regenerate :: String -> Set String -> IO String
regenerate pat set = do
  next <- generate pat
  if next `Set.member` set
    then regenerate pat set          -- ← no attempt counter
    else do
      modifyIORef' strings (Set.insert next)
      pure next

randomString :: String -> IO String
randomString pat
  | randomized pat = readIORef strings >>= regenerate pat
  | otherwise = generate pat
```

1. **Infinite loop.** `regenerate` recurses without any bound. A `%d` pattern draws from exactly 10,000 values (`randomRIO (0,9999)` padded to `%04d`, `Random.hs:29`). The `strings` set is **process-global and monotonic** — it is never cleared. After 10,001 `randomString "%d"` calls in one process, every value is already in the set, `Set.member` is always `True`, and `regenerate` never returns. Same risk (after astronomically more calls) for `%x`, which has 16⁸ values.
2. **Uniqueness is process-global, not per-run.** The set accumulates across separate `runCLI` invocations in the same process (tests, embedding, `--sequence` reruns). A fresh run starts with a polluted set, so `--seed` no longer determines the produced names — same seed, different history → different names. This is the same class of global-state leak as the `Tau.taus` IORef.

## Repro

```
> replicateM 10002 (randomString "%d")   -- hangs after the 10001st
```

In the CLI this is reachable if a user rule uses `random-string` with `%d` many times (e.g. a long normalization chain), or from tests that call `randomString` in a loop.

## Expected

- Add a bounded number of attempts (e.g. retry a few times, then widen the space or raise a clear error instead of spinning).
- Make the uniqueness set per-run (reset at the start of a run, like `seedTaus` does for `Tau`), so `--seed` fully determines output and the same rule behaves identically across runs.

## Impact

Potential hang/DoS and non-reproducible `--seed` output for rules that mint many `%d` random strings. Related to the general "global mutable state" concerns in `Tau.hs` and `Logger.hs`.

VasilevNStas:
Fixes #1081.

## Problem

`regenerate` (`src/Random.hs:37-44`) recursed without any bound:

```haskell
regenerate pat set = do
  next <- generate pat
  if next `Set.member` set
    then regenerate pat set        -- ← unbounded recursion
    else ...
```

The `strings` set is process-global and monotonic. A `%d` pattern has exactly
10,000 values, so after 10,001 calls in one process every value is already in
the set and `regenerate` never returns — a hang (and a real DoS risk for rules
that mint many `%d` strings).

## Fix

`src/Random.hs` — `regenerate` now gives up after `maxAttempts = 100000`
attempts and throws a descriptive `userError` instead of recursing forever.
The limit is 10x the `%d` space so that finding the last free value of a
nearly-full space still succeeds with overwhelming probability
(`(9999/10000)^100000 ≈ 4.5e-5`), while a fully-exhausted space fails fast.

## Changes

- `src/Random.hs` — `maxAttempts` constant; `regenerate`/`go` with a bounded
  attempt counter and a descriptive error on exhaustion.
- `test/RandomSpec.hs` — new test filling the whole `%d` space and asserting
  the next call raises an error (guarded by `timeout` so a regression would
  fail, not hang).

## Tests

- `test/RandomSpec.hs` — 28 examples, 0 failures; full suite green except the
  pre-existing 13 `LocatorSpec` failures (documented in #1077).


VasilevNStas:
@yegor256 
please review this pull request

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1770](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1102, opened 2026-09-06
fix(cli): reject --show/--hide overlap and deduplicate repeated --rule files (#1085)

### LINKED ISSUE #1085: Repo hygiene: committed src/Must.hs.bak; getRules does not deduplicate --rule; --hide + --show not validated
Repo hygiene: committed `src/Must.hs.bak`; `getRules` does not deduplicate repeated `--rule` files; `--hide` + `--show` combined is not validated

## Problem

Three small hygiene/robustness gaps noticed during a full-repo read:

### 1. `src/Must.hs.bak` is committed

`git ls-files` shows `src/Must.hs.bak` is tracked — an identical copy of `src/Must.hs`, despite `*.bak` being in `.gitignore`. It is dead code that only confuses readers and code-search.

### 2. `getRules` does not deduplicate `--rule` files

`src/CLI/Helpers.hs:157-180`: if the same file is passed via `--rule` twice (or a file overlaps with a built-in normalize rule set), the rule is applied twice:

```
$ phino rewrite --normalize --rule=foo.yaml --rule=foo.yaml input.phi
```

The user probably meant the rule once; the engine silently doubles its effect. `--breakpoint` matching (`Runners.hs:95-101`) additionally picks the first rule with a given name, so a duplicate silently changes which rule a breakpoint stops at.

### 3. `--hide` and `--show` together are not validated

`rewrite`/`dataize`/`merge` accept `_hide` and `_show` (`CLI/Types.hs:102-103`, `:154-155`) and apply them as `exclude $ include …` (`Runners.hs:61-65`, `:150-156`). There is no validation that the two option sets are disjoint or that a `--show` path is not also in `--hide`; the composition order (`exclude` over `include`) can silently wipe out exactly the subtree `--show` was meant to keep. `--show` is limited to one occurrence (`Runners.hs:77`), but `--hide` accepts many, so an accidental overlap produces a confusing empty result.

## Expected

- Remove `src/Must.hs.bak` from the repo.
- In `getRules`, deduplicate by file path (or warn when the same file is listed more than once).
- Validate that `--hide` and `--show` paths do not overlap (or document the composition order and warn on overlap).

## Impact

Low severity, but all three are cheap to fix and remove silent surprises: dead file in the tree, doubled rule effects, and an option combination that silently does the wrong thing.

VasilevNStas:
Fixes #1085 (the two code parts; the third item, `src/Must.hs.bak`, was a
local un-tracked file — not in git history, removed from the working tree).

## Problem

1. **`--hide` + `--show` overlap is not validated.** `runRewrite`/`runDataize`
   apply `exclude $ include …` (`src/CLI/Runners.hs:65,152`), so a `--show`
   locator that is also in `--hide` silently gets the very subtree wiped out
   that `--show` was meant to keep.
2. **`getRules` does not deduplicate `--rule` files.** Passing the same file
   twice (`--rule=x.yaml --rule=x.yaml`) loads the rule twice, doubling its
   effect; a `--breakpoint` by that rule's name then silently stops at the
   first copy.

## Fix

- `src/CLI/Validators.hs` — new `validateNoOverlap` rejecting a `--show`
  locator that is also listed in `--hide`.
- `src/CLI/Runners.hs` — call `validateNoOverlap` in both `runRewrite` and
  `runDataize` right after `validatedDispatches`.
- `src/CLI/Helpers.hs` — `getRules` deduplicates the `--rule` paths via `nub`
  before loading.

## Changes

- `src/CLI/Validators.hs` — `validateNoOverlap`.
- `src/CLI/Runners.hs` — wire the new validation into `runRewrite`/`runDataize`.
- `src/CLI/Helpers.hs` — `nub rules` in `getRules`.
- `test/CLIHelpersSpec.hs` — `getRules` dedups the same file listed twice.
- `test/CLISpec.hs` — `--show=Q.x --hide=Q.x` fails with the overlap message.

## Tests

- `test/CLIHelpersSpec.hs`, `test/CLISpec.hs` — new cases pass; full suite
  green except the pre-existing 13 `LocatorSpec` failures (documented in
  #1077).

### DIFF
--- a/src/CLI/Helpers.hs
+++ b/src/CLI/Helpers.hs
@@ -14,7 +14,7 @@ import Control.Exception
 import Control.Monad ((>=>))
 import Data.Functor ((<&>))
 import Data.IORef
-import Data.List (intercalate)
+import Data.List (intercalate, nub)
 import Data.Maybe
 import Deps (SaveEvalFunc, SaveStepFunc, dontSaveEval, saveEval, saveStep)
 import Encoding
@@ -176,7 +176,7 @@ getRules normalize shuffle rules = do
           pure []
       | otherwise = do
           logDebug (printf "Using rules from files: [%s]" (intercalate ", " rules))
-          yamls <- mapM ensuredFile rules
+          yamls <- mapM ensuredFile (nub rules)
           mapM (Y.yamlRule >=> validateRewriteRule) yamls
 
 -- Pass a user-supplied rewriting rule through unchanged, or fail fast if it
--- a/src/CLI/Runners.hs
+++ b/src/CLI/Runners.hs
@@ -45,6 +45,7 @@ runRewrite OptsRewrite{..} = do
   included <- validatedDispatches "show" _show
   [loc] <- validatedDispatches "locator" [_locator]
   [foc] <- validatedDispatches "focus" [_focus]
+  validateNoOverlap "show" included "hide" excluded
   setStdGen (mkStdGen _seed)
   rules <- getRules _normalize _shuffle _rules
   validateBreakpoint _breakpoint rules
@@ -142,6 +143,7 @@ runDataize OptsDataize{..} = do
   included <- validatedDispatches "show" _show
   [loc] <- validatedDispatches "locator" [_locator]
   [foc] <- validatedDispatches "focus" [_focus]
+  validateNoOverlap "show" included "hide" excluded
   input <- readInput _inputFile
   expr <- parseInput input _inputFormat
   setStdGen (mkStdGen _seed)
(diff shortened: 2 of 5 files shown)


VasilevNStas:
@yegor256 
plz review

yegor256:

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1782](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1074, opened 2026-09-05
L_number_eq yields a number (or self.y), not a boolean — contradicts EO


VasilevNStas:
`L_number_eq` yields a number (or `self.y`), not a boolean — contradicts EO

## Problem

`src/Dataize.hs:501-509`:

```haskell
atom "L_number_eq" self univ state ctx = do
  (x, lstate) <- _dataize (ExDispatch self (AtLabel "x")) univ state ctx
  (rho, rstate) <- _dataize (ExDispatch self AtRho) univ lstate ctx
  case (asNumber x, asNumber rho) of
    (Just first, Just self') ->
      if self' == first
        then pure (DataNumber (numToBts first), rstate)   -- number, not boolean
        else pure (ExDispatch self (AtLabel "y"), rstate) -- dispatch, not false
    _ -> pure (ExTermination, rstate)
```

On equality it returns the **left number** as a `DataNumber`; on inequality it **dispatches `self.y`**. Modern EO (`number.eo`, `boolean.eo`) returns a `boolean` from `eq`. Any program that branches on the result of `eq` will misbehave.

## Repro (verified)

```
$ phino dataize '⟦ φ ↦ 5.eq(5) ⟧'
40-14-00-00-00-00-00-00        -- the number 5, not a boolean object

$ phino dataize '⟦ φ ↦ 5.eq(6) ⟧'
self.y                         -- dispatch on 'y', not a false object
```

(These are the raw atom results; through a universe that provides `true`/`false` the `self.y` branch dataizes to whatever `y` is.)

## Context

- The same file already has a `boolean` helper (`Dataize.hs:477-479`) returning `BaseObject "true"/"false"`, used by `L_number_gt` (`Dataize.hs:516-521`) — so `eq` *should* have used it too.
- Our earlier issue #1024 (closed as intent) documented that `L_number_eq` on **non-numbers** yields ⊥; the current report is about the **numeric** branches returning the wrong type.
- Happy paths of `L_number_eq`, `L_number_plus`, `L_number_times` are **not covered** by direct atom tests (`DataizeSpec.hs:685-741` tests `div`, `gt`, all bytes atoms, but skips `eq`/`plus`/`times`; `eq` appears only in the stuck-atom test at `:754`).

## Expected

- Equality → `boolean True`, inequality → `boolean False` (`BaseObject "true"/"false"`), consistent with `L_number_gt`.
[...]

PR #1110, opened 2026-09-07
feat(cli): `atoms` command listing the λ functions phino implements

### LINKED ISSUE #1108: `atoms` command to list the λ functions phino implements
phino fails a run when a λ function is unknown to it — the README says so, and the
error is `[ERROR]: Atom 'L_foo' does not exist` — but there is no way to ask which
ones it does know. `phino --help` lists `rewrite`, `dataize`, `explain`, `merge` and
`match`; `explain` prints the rewriting rules and never mentions atoms; the README
uses `L_number_plus`, `L_number_times` and `L_number_minus` in its examples without
saying what the set is. So the only way to find out whether a name exists is to build
a universe around it and dataize, one name at a time.

That is how we found out in `objectionary/eo`. Its `eo-lowering` module names 18
distinct `L_*` functions across `universe.phi` and `ops.tsv`, and probing each one
against 0.0.115 gives three different answers: 12 exist and agree with EO
(`L_number_plus`, `L_number_times`, `L_number_div`, `L_number_gt`, and the eight
`L_bytes_*`), 5 do not exist at all (`L_bool_if`, `L_string_slice`, `L_tuple_length`,
`L_tuple_at`, `L_object_as_bytes`), and 1 exists with semantics that differ from EO's
— `L_number_eq`, which is #1074. The five that do not exist are relied on deliberately:
eo-lowering declares them so that `--partial` parks on them and renders them to Java
itself. So both "exists" and "does not exist" are legitimate states for a caller to
depend on, which is exactly why a caller needs to tell them apart mechanically rather
than by experiment — and needs to notice when a release moves a name from one state to
the other. `L_number_minus` makes the same point from the other side: it appears in
the README's own `merge` example, and it does not exist. The example never fires it,
so nothing there is wrong, but an `L_*` name in a `.phi` file reads exactly the same
whether phino implements it or not.

An `atoms` command would close this. Plain `phino atoms` printing one name per line is
enough for a human, and `phino atoms --json` with the details would let a build check
its own table against the binary and fail when the two drift: for each function, the
name, the labels it reads off the formation (`x`, `b`, `start`/`len`/`cant-slice`),
whether it reads `ρ`, the forma of what it answers, and a one-line statement of the
semantics. #482 asks for an `ebnf` command in the same spirit — the tool describing
itself rather than the caller guessing.

yegor256:
Closes #1108

## What

A new `atoms` command that makes phino describe its own set of λ functions,
so that a caller no longer has to build a universe around one name and
dataize it to find out whether that name exists.

```bash
$ phino atoms
L_bytes_and
L_bytes_concat
L_bytes_eq
L_bytes_not
L_bytes_or
L_bytes_right
L_bytes_size
L_bytes_slice
L_number_div
L_number_gt
L_number_plus
L_number_times
```

With `--json`, each function comes with the details the issue asks for — the
labels it reads off the formation it fires against, whether it reads `ρ`, the
forma of the object it answers, and a one-line statement of its semantics:

```bash
$ phino atoms --json
[
  {
    "name": "L_bytes_slice",
    "labels": ["start", "len", "cant-slice"],
    "rho": true,
    "forma": "Φ.bytes, or the forma of cant-slice",
    "semantics": "The len bytes of ρ starting at offset start; a window reaching past the end of ρ applies cant-slice to a complaint string instead; ⊥ unless start and len are non-negative whole numbers within the 32-bit range."
  },
...
]
```

Keys are written in a fixed order and the values are the catalogue's own, so
two runs of the same binary print byte-identical output and a build can diff
it against its own table. Both forms honour `--target`, like the other
commands.

## How

- `src/Atoms.hs` — the catalogue (name, labels, `ρ`, forma, semantics) plus
  the two renderers. Names are alphabetical, which is also the printed order.
  An atom that answers off more than one path names every forma it can
  answer, not just its happy path.
- `src/Dataize.hs` — the λ functions moved from the clauses of `atom` into an
  association list, `implementations`, with `atom` now a `lookup` over it and
  each firing a named `Firing` under `where`. This makes the implemented names
  enumerable (`implementedAtoms`) instead of hidden inside patterns; the
  behaviour of every atom is unchanged.
- `src/CLI/{Types,Parsers,Runners}.hs`, `src/CLI.hs` — the command, its
  `--json` switch and its runner.
- `README.md` — a new `## Atoms` section.

## Testing

A printed list is only worth what its accuracy is worth, so the catalogue is
held to the engine from both sides. The two ways of being wrong fail
differently: a catalogued name that is dead is a **false positive** — a caller
trusts it and its run dies on the λ; a name that fires while unlisted is a
**false negative** — a caller reads the list, leaves the name for `--partial`
to park on, and gets it computed instead.

- `test/AtomsSpec.hs`
  - `names exactly the λ functions the dataizer implements` — compares the
    catalogue with `Dataize.implementations`, settling both directions
    exhaustively for the dispatch table itself.
  - `no false positive: every catalogued λ function really fires` — runs the
    engine over all twelve. One probe suffices per name and needs no per-atom
    operands: a formation carrying nothing but the λ, dataized, makes an
    implemented function fire, read its unbound operands and answer ⊥, while an
    unimplemented one never fires and reports itself missing.
  - `no false negative: no λ function outside the catalogue fires` — the same
    probe over names the catalogue omits. An infinite name space cannot be
    enumerated, so it holds the ones that matter: the five EO's lowering
    declares precisely so that `--partial` parks on them (#1108),
    `L_number_eq` as a standing guard on #1074, the `L_number_minus` of the
    README's own merge example, and every near miss of a catalogued name,
    generated from the catalogue so the set grows with it.
  - plus the alphabetical order, the one-line semantics, one name per line, and
    that the JSON parses back into the very catalogue it was printed from.
- `test/DataizeSpec.hs` — a `catalogue` block for the case a name check cannot
  see: an atom that keeps its name and changes what it answers. It fires each
  such atom and asserts both the entry's `_forma` text and the object the
  firing actually answers, beside the `primitives` universe they fire in.
- `test/CLISpec.hs` — a `describe "atoms"` block for the help text, the plain
  and `--json` output and `--target`.

Each guard was checked to bite, not assumed to: cataloguing a name nothing
implements fails the false-positive test, and implementing `L_bool_if` fails
the false-negative one.

Local checks: `cabal test --ghc-options=-Werror`, `hlint src app test`,
`fourmolu --mode check src app test` (0.17.0.0, as in CI), `typos`,
`markdownlint`, and `make coverage` at 82% against the 80% threshold.

## Notes

Synced with `master`, which merged #1112: `L_number_eq` is gone from the
engine, since EO's `number.eq` is pure EO composition over `L_bytes_eq`.
That change landed on the clause-based `atom` this branch had already turned
into a table, so the two collided; the resolution keeps the table, drops the
atom from it and carries master's reasoning about the deliberately absent
name — and about `L_bool_if`/`L_string_slice` — into the table's docblock.
The catalogue lists the same twelve names the engine now fires.

https://claude.ai/code/session_01GzHFG1wiuuaS5oUMky6viE

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,6 +376,59 @@ $ phino explain --contextualize
 
 For more details, use `phino [COMMAND] --help` option.
 
+## Atoms
+
+A λ binding names a function that `phino` has to implement itself. When it
+doesn't, the run fails with `Atom 'L_foo' does not exist`, unless `--partial`
+parks on it — and a build may well rely on either answer. The `atoms` command
+prints the names this version implements, one per line, so that no one has to
+find them out by dataizing a universe around every single name:
+
+```bash
+$ phino atoms
+L_bytes_and
+L_bytes_concat
+L_bytes_eq
+L_bytes_not
+L_bytes_or
+L_bytes_right
+L_bytes_size
+L_bytes_slice
+L_number_div
+L_number_eq
+L_number_gt
+L_number_plus
+L_number_times
+```
+
+With `--json`, every function comes with its details: the labels it reads off
+the formation it fires against, whether it reads `ρ`, the forma of the object
+it answers, and a one-line statement of its semantics. A build may diff this
+against its own table of names and fail when the two drift apart:
+
+```bash
+$ phino atoms --json
+[
+  {
+    "name": "L_bytes_and",
+    "labels": ["b"],
+    "rho": true,
+    "forma": "Φ.bytes",
+    "semantics": "Bitwise AND of ρ and b; ⊥ when the two byte arrays differ in size."
+  },
+...
+  {
+    "name": "L_number_plus",
+    "labels": ["x"],
+    "rho": true,
+    "forma": "Φ.number",
+    "semantics": "The sum of ρ and x; ⊥ unless both operands are 8-byte numbers."
+  }
+]
+```
+
+Both forms are printed to the console, or to the file given by `--target`.
+
 ## Rule structure
 
 This is BNF-like yaml rule structure. Here types ended with
--- a/phino.cabal
+++ b/phino.cabal
@@ -35,6 +35,7 @@ library
   import: warnings
   exposed-modules:
     AST
+    Atoms
     Builder
     Bytes
     Canonizer
@@ -124,6 +125,7 @@ test-suite spec
   hs-source-dirs: test
   other-modules:
     ASTSpec
+    AtomsSpec
     BuilderSpec
     BytesSpec
     CanonizerSpec
--- a/src/Atoms.hs
+++ b/src/Atoms.hs
@@ -1,0 +1,168 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+-- SPDX-License-Identifier: MIT
+
+-- The catalogue of λ functions — atoms — phino implements. A λ binding naming
+-- a function absent from here cannot fire: dataization stops with
+-- "Atom 'L_foo' does not exist", unless '--partial' parks on it. Both answers
+-- are legitimate states for a caller to depend on, so the catalogue is printed
+-- by the 'atoms' command, letting a build check its own table of names against
+-- the binary instead of probing one name at a time. The names here are exactly
+-- the ones 'Dataize.implementedAtoms' fires, and 'AtomsSpec' keeps the two
+-- from drifting apart.
+module Atoms (Atom (..), atoms, printAtoms, printAtomsInJSON) where
+
+import Data.Aeson.Text (encodeToLazyText)
+import Data.Text (Text)
+import qualified Data.Text as T
+import qualified Data.Text.Lazy as TL
+
+-- One λ function of the catalogue
+data Atom = Atom
+  { _name :: Text
+  -- ^ The name of the function, as it stands in a λ binding
+  , _labels :: [Text]
+  -- ^ The labels the function reads off the formation it fires against
+  , _rho :: Bool
+  -- ^ Whether the function reads ρ of that formation
+  , _forma :: Text
+  -- ^ The forma of the object the function answers
+  , _semantics :: Text
+  -- ^ One-line statement of what the function computes
+  }
+  deriving (Eq, Show)
+
+-- Every λ function phino implements, in alphabetical order. An operand the
+-- atom cannot read leaves it with ⊥ (termination), which the 'semantics' line
+-- of each entry spells out.
+atoms :: [Atom]
+atoms =
+  [ Atom
+      { _name = "L_bytes_and"
+      , _labels = ["b"]
+      , _rho = True
+      , _forma = "Φ.bytes"
+      , _semantics = "Bitwise AND of ρ and b; ⊥ when the two byte arrays differ in size."
+      }
+  , Atom
+      { _name = "L_bytes_concat"
+      , _labels = ["b"]
+      , _rho = True
+      , _forma = "Φ.bytes"
+      , _semantics = "The bytes of ρ followed by the bytes of b."
+      }
+  , Atom
+      { _name = "L_bytes_eq"
+      , _labels = ["b"]
+      , _rho = True
+      , _forma = "Φ.bool"
+      , _semantics = "Φ.true when ρ and b are the same byte array octet by octet, Φ.false otherwise."
+      }
+  , Atom
+      { _name = "L_bytes_not"
+      , _labels = []
+      , _rho = True
+      , _forma = "Φ.bytes"
+      , _semantics = "Bitwise NOT of every byte of ρ."
+      }
+  , Atom
+      { _name = "L_bytes_or"
+      , _labels = ["b"]
+      , _rho = True
+      , _forma = "Φ.bytes"
+      , _semantics = "Bitwise OR of ρ and b; ⊥ when the two byte arrays differ in size."
+      }
+  , Atom
+      { _name = "L_bytes_right"
+      , _labels = ["x"]
+      , _rho = True
+      , _forma = "Φ.bytes"
+      , _semantics = "ρ shifted right by x bit positions, or left when x is negative, keeping its size; ⊥ unless x is a whole number within the 32-bit range."
+      }
+  , Atom
+      { _name = "L_bytes_size"
+      , _labels = []
+      , _rho = True
+      , _forma = "Φ.number"
+      , _semantics = "The number of bytes in ρ."
+      }
+  , Atom
+      { _name = "L_bytes_slice"
+      , _labels = ["start", "len", "cant-slice"]
+      , _rho = True
+      , _forma = "Φ.bytes"
+      , _semantics = "The len bytes of ρ starting at offset start; a window reaching past the end of ρ applies cant-slice to a complaint string instead; ⊥ unless start and len are non-negative whole numbers within the 32-bit range."
+      }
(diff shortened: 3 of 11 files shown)


yegor256:
@maxonfjvipon good?

maxonfjvipon (changes requested):
Two things in the catalogue have to be right before this lands, because the
whole point of `atoms --json` is that a build trusts it.

**1. `_forma = "Φ.bool"` names an object phino never produces.**
`boolean` (`src/Dataize.hs:479-481`) answers `BaseObject "true"` /
`BaseObject "false"` — `Φ.true` / `Φ.false`. There is no `bool` anywhere in
`src/`, `resources/` or the README, so `L_bytes_eq` and `L_number_gt` both
advertise a forma that does not exist. The record even contradicts itself: its
own `semantics` line already spells the answer as "Φ.true … Φ.false". The one
field that names a thing must name a real one.

While there: `L_bytes_slice` claims `Φ.bytes`, but the `cant-slice` path
answers an application of the caller's own `cant-slice`, not bytes. Either
`_forma` covers every path the atom can take, or its docblock says it names
the happy path only.

**2. The `L_number_eq` entry is about to become a lie, and no test will
notice.** It documents the `y` dispatch, which #1112 removes — that is the
maintainer's decision in #1074. `AtomsSpec` ties the two lists by
`map _name atoms == implementedAtoms` and nothing else, so whichever of the
two PRs lands second leaves the binary misdescribing itself with CI green.
Please rebase on top of #1112 and move the entry with it: `_labels = ["x"]`,
`_forma` the boolean one, and semantics to match. Worth saying in the
`Atoms.hs` docblock that names are the only automated tie and the rest is a
hand-checked promise — otherwise the next atom change silently drifts the
same way.

**3. `printAtoms` and `printAtomsInJSON` are values, not actions.** They
render; `printOut` in the runner does the printing. Per the CQRS naming rule
in `CLAUDE.md` a query takes a noun — `listing` and `json`, say. The `print`
prefix reads as `IO ()` at every call site.

The `Dataize.hs` restructuring into `implementations` is the right move and
reads well — making the names enumerable instead of buried in patterns is
worth the churn, and `Firing` earns its name.

yegor256:
Thanks — points 1 and 2 were right and are fixed in `b077acf` (which also merges current `master`, since #1111 and #1047 landed under this branch).

**1. The formas.** You are right that `Φ.bool` names nothing: `boolean` answers `BaseObject "true"`/`BaseObject "false"`, and `grep` finds no `bool` in `src/`, `resources/` or the README. `L_bytes_eq` and `L_number_gt` now name `"Φ.true or Φ.false"`, agreeing with their own semantics lines. `L_bytes_slice` now names `"Φ.bytes, or the forma of cant-slice"`, and the field's comment says it covers every path the atom can answer off, not the happy one.

**2. The drift.** Rather than only warn in the docblock, I made the forma column actually checked. `DataizeSpec` now has a `catalogue` block that, for each atom whose answer can change, asserts both the entry's `_forma` text and the object a real firing answers — pinned in the `primitives` universe, which is why the cases live there and not in `AtomsSpec`. Changing such an atom reddens the suite, and getting it green again means moving the entry with it.

I checked that it bites rather than assuming so: applying your #1112 change to `numberEq` locally fails exactly the two cases that describe today's behaviour —

```
1) Dataize.catalogue.L_number_eq answers ρ itself, a number, when the operands are equal
     expected: Dataized (BtMany ["40","14","00","00","00","00","00","00"])
      but got: Dataized (BtOne "FF")
2) Dataize.catalogue.L_number_eq answers the y of the formation when they are not
     expected: Dataized (BtMany ["40","1C","00","00","00","00","00","00"])
      but got: Dataized (BtOne "00")
```

So whichever of the two PRs lands second cannot land with the catalogue stale. I did not rebase onto #1112: it is an unmerged branch on a fork and currently conflicted against `master`, so rebasing would pull its commit into this PR and couple the two so neither can land alone. With the test above, that coupling buys nothing. Whoever lands second flips the `L_number_eq` entry to `_labels = ["x"]`, the boolean forma and matching semantics, and updates the pinned firings in the same commit — I'm happy to do that here the moment #1112 merges, or to fold it in now if you would rather this PR wait on it.

The `Atoms.hs` docblock now states plainly that the names are the only automatic tie and every other field is a promise kept by hand, with the pinned entries called out.

**3. The rename** I have left alone, because I could not confirm the rule and the repo's own convention points the other way — please tell me if you still want it and I will push it as a one-liner. `CLAUDE.md` on `master` has no CQRS or noun/verb rule; its `### Style` section is five bullets, on `where` scoping, explicit signatures, single-char variables, the `foo`/`foo'` pair and naming recursive workers `go`. And a pure `print<Thing> :: … -> String` is what this codebase does everywhere: `printExpression`, `printAttribute`, `printAlpha`, `printBinding`, `printBytes`, `printExtraArg`, `printMetaValue`, `printSubst`, `printSubsts` in `Printer.hs`, `printXMIR`, `printElement`, `printNode` in `XMIR.hs`, `printPhi` in `CLI/Helpers.hs` — none of them do IO, and `printOut` is the one that writes. `printAtoms :: String` follows that; `listing`/`json` would make this module the exception. Your underlying point stands either way, though — the runner is where the printing happens.

Also glad the `implementations`/`Firing` restructuring reads well; that was the part I was least sure was worth the churn.

Local checks on the new head: `cabal test --ghc-options=-Werror`, `hlint src app test`, `fourmolu --mode check src app test` (0.17.0.0), `typos`, `markdownlint`, and `make coverage` at 82% against the 80% threshold.

---

yegor256:
Synced with `master` in `cb11b9d`. #1112 landed in the meantime and settled your point 2 more decisively than either of us expected: `L_number_eq` is not boolean now, it is **gone** — EO's `number.eq` is pure EO composition over `L_bytes_eq`, so no atom implements it.

It landed on the clause-based `atom` that this branch had already turned into `implementations`, so the two edits collided in `src/Dataize.hs`. Resolved by keeping the table and dropping the atom from it, and carrying master's reasoning into the table's own docblock — why the name is deliberately absent, and why `L_bool_if`/`L_string_slice` must stay unimplemented so `--partial` parks on them. The catalogue entry and its `catalogue` firings went with it, and the README listing too, so `phino atoms` now prints the same twelve names the engine fires.

Worth noting which guard actually caught this, since it bears on your point 2: the name equality in `AtomsSpec` did — a dropped atom breaks `map _name atoms == implementedAtoms` immediately. The forma firings are for the case that equality cannot see, an atom keeping its name while changing what it answers. I have reworded that helper's docblock to say so, rather than citing #1074 as a change still to come.

Also deduped two lines that #1112's docblock edit repeated in the `primitives` comment — they were in the conflict region.

Point 3 (the `printAtoms` rename) is still open on your word; the evidence is in my previous comment.

New head: `cabal test --ghc-options=-Werror`, `hlint`, `fourmolu` 0.17.0.0, `typos` and `markdownlint` all clean, `master` fully merged, no conflicts.

---

yegor256:
@maxonfjvipon better now?

yegor256:
@maxonfjvipon we'll do this instead: https://github.com/objectionary/phino/issues/1113

0crat:
@maxonfjvipon Thanks for the review! You've earned +11 points for this: +12 as a basis; +6 for the 678 hits-of-code that you reviewed; -8 for absolutely no comments posted.  Your running score is [+1372](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1115, opened 2026-09-08
feat(cli): add `morph` command, exposing 𝕄 outside of 𝔻

### LINKED ISSUE #1114: No `morph` command: 𝕄 in src/Dataize.hs:184 is reachable only through dataize
Morphing 𝕄 is implemented (`morph`, `src/Dataize.hs:184`) and formalized
(`resources/morphing.yaml`), but it is reachable only from inside 𝔻:
`runDataize` (`src/CLI/Runners.hs:141`) is its only caller, through the
`norm` rule of `resources/dataization.yaml`. There is no way to ask phino
for 𝕄(n, Φ) on its own. `explain --morph` prints the rules; it does not
run them.

𝕄 answers a different question from 𝔻, and the question is worth asking.
𝔻 insists on bytes: it unboxes φ (`box`), fires a bare saturated
λ-formation (`fire`) and terminates on a formation carrying neither Δ, λ
nor φ (`none`). 𝕄 stops at the first formation it reaches — resolving Φ
against the universe, peeling dispatches and applications through 𝒩, and
firing along the way whichever atoms sit under a dispatch (`ml`) — then
hands that formation back untouched (`mf`). That is "evaluate as far as
the object model allows, without demanding data", which is what one wants
when the term has no Δ at the end of it, or when the interesting answer is
the formation rather than its bytes.

Concretely, with two chained atom calls:

```
$ cat two.phi
⟦
  bytes(data) ↦ ⟦ φ ↦ data ⟧,
  number(as-bytes) ↦ ⟦ φ ↦ as-bytes, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧,
  φ ↦ 5.plus( 6 ).plus( 7 )
⟧
$ phino dataize --sweet --hide-rho two.phi
40-32-00-00-00-00-00-00
$ phino morph --locator=Q.φ --sweet --hide-rho two.phi
⟦ x ↦ 7, λ ⤍ L_number_plus ⟧
```

The inner `5.plus( 6 )` fires under `ml`, because `.plus` is dispatched on
its result, and `11` lands in the ρ hidden by `--hide-rho`. The outer
application is saturated but bare, so `mf` returns it and 𝕄 is finished;
firing it is 𝔻's job and takes `dataize` on to `18`. That boundary is real
today but visible only by reading a chain — in
`phino dataize --sequence --headers --quiet --sweet --flat two.phi` the
step labelled `mf` is the last 𝕄 step and the next one is `evaluate`. A
`morph` command would print that step directly.

Two behaviours worth pinning down while implementing. The default locator
`Q` morphs the whole top formation, which `mf` returns unchanged, so
`--locator` is how one aims 𝕄 at a subterm, exactly as in `dataize`. And ⊥
is a legal answer for 𝕄 (`dead`, `xi`, `mg`, `mad`, `maad`), so `morph`
should print `⊥` where `dataize` fails the run, since 𝕄 is total and 𝔻 is
not.

The change: add `command "morph" (info morphParser (progDesc "Morph the
𝜑-expression"))` next to `dataize` in `src/CLI/Parsers.hs:398`, an
`OptsMorph` record mirroring `OptsDataize` (`src/CLI/Types.hs:76`), a
`runMorph` beside `runDataize`, and a top-level wrapper next to `dataize`
(`src/Dataize.hs:242`) that locates the subterm, calls `morph` and returns
the expression with its chain instead of `(Outcome, [Rewritten])`. The
whole `dataize` option surface applies unchanged — `--sequence`,
`--headers`, `--steps-dir`, `--evaluations`, `--partial`, `--max-steps`,
`--shuffle`/`--seed`, `--output`, `--focus` — and so would the
`--atoms=a.json` registry proposed in #1113, since `ml` fires atoms just
as `fire` does.

yegor256:
Closes #1114.

Morphing 𝕄 was implemented (`morph`, `src/Dataize.hs`) and formalized
(`resources/morphing.yaml`), but reachable only from inside 𝔻, through the
`norm` rule of `resources/dataization.yaml`. `explain --morph` printed the rules;
it did not run them. This adds a `morph` command that runs 𝕄 on its own.

## What changed

`src/Dataize.hs` splits 𝕄 the way 𝔻 is already split: the recursive worker the
rules drive is now `morph'`, and a new top-level `morph` wrapper locates the
subterm, threads the whole input expression as the universe and returns the
morphed expression together with its chain of steps (oldest first), instead of
`(Outcome, [Rewritten])`.

Two behaviours the issue asked to pin down:

* the default locator `Q` morphs the whole top formation, which `mf` returns
  unchanged, so `--locator` is how one aims 𝕄 at a subterm, exactly as in
  `dataize`;
* 𝕄 is total where 𝔻 is not, so `morph` prints `⊥` (`dead`, `xi`, `mg`, `mad`,
  `maad`) where `dataize` fails the run. Only the atoms `ml` fires can still get
  stuck, and `--partial` parks them exactly as under 𝔻, answering with the
  residual subterm.

`src/CLI/Types.hs` gains `CmdMorph` and an `OptsMorph` record mirroring
`OptsDataize`; `src/CLI/Parsers.hs` gains `morphParser` and
`command "morph" (info morphParser (progDesc "Morph the 𝜑-expression"))` next to
`dataize`; `src/CLI/Runners.hs` gains `runMorph` beside `runDataize`. The whole
`dataize` option surface applies unchanged — `--sequence`, `--headers`,
`--steps-dir`, `--evaluations`, `--partial`, `--max-steps`, `--shuffle`/`--seed`,
`--output`, `--focus`.

Two shared help strings were generalized, since they now cover three commands:
`--locator` no longer says "object to dataize", and `--partial` no longer ends
with "instead of bytes".

## The boundary it makes visible

```
$ cat two.phi
⟦
  bytes(data) ↦ ⟦ φ ↦ data ⟧,
  number(as-bytes) ↦ ⟦ φ ↦ as-bytes, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧,
  φ ↦ 5.plus( 6 ).plus( 7 )
⟧
$ phino dataize --sweet --hide-rho two.phi
40-32-00-00-00-00-00-00
$ phino morph --locator=Q.φ --sweet --hide-rho two.phi
⟦ x ↦ 7, λ ⤍ L_number_plus ⟧
```

The inner `5.plus( 6 )` fires under `ml`, because `.plus` is dispatched on its
result. The outer application is saturated but bare, so `mf` returns it and 𝕄 is
finished; firing it is 𝔻's job and takes `dataize` on to `18`.

## Tests and docs

* `test/CLISpec.hs`: a `describe "morph"` block — the example above, the
  ⊥-versus-failure contrast against `dataize`, `--locator`, `--sequence`
  with `--headers`, `--quiet`, `--evaluations`, `--steps-dir`, `--partial` with
  and without a stuck atom, `--max-steps`, and the option validations.
* `test/DataizeSpec.hs`: wrapper tests (formation returned untouched, ⊥ answer,
  chain shape, stuck atom with and without `_partial`), plus the rename of the
  existing `morph` cases to `morph'`.
* `README.md`: a `## Morph` section; `CLAUDE.md`: the command count and the
  wrapper/worker split.

https://claude.ai/code/session_01HCXhcLD16gx4e1t41dLA7i

---
_Generated by [Claude Code](https://claude.ai/code/session_01HCXhcLD16gx4e1t41dLA7i)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,9 +48,9 @@ via `benchmark/mvnw`.
 components: `library` (`src/`), executable `phino` (`app/`), test suite
 `spec` (`test/`), benchmark suite `bench` (`benchmark/`).
 
-### Five CLI commands
+### Six CLI commands
 
-`rewrite` | `dataize` | `explain` | `merge` | `match` — all wired in
+`rewrite` | `dataize` | `morph` | `explain` | `merge` | `match` — all wired in
 `src/CLI/Runners.hs`, parsed in `src/CLI/Parsers.hs`.
 
 ### Two-phase rendering pipeline
@@ -86,7 +86,11 @@ type alias.
 `Dataize.hs` implements the formal Morphing (M) and Dataization (D)
 functions with named rules: PRIM, NMZ, LAMBDA, PHI (morphing) and DELTA,
 BOX, NORM (dataization). All configuration is threaded through
-`DataizeContext` and `RewriteContext` records — no global state.
+`DataizeContext` and `RewriteContext` records — no global state. Each
+function has a top-level wrapper that locates the subterm and starts the
+chain (`morph`, `dataize`) and a recursive worker the rules drive
+(`morph'`, `dataize'`); the `morph` and `dataize` commands enter through
+the wrappers.
 
 ### Test pattern: YAML packs
 
--- a/README.md
+++ b/README.md
@@ -177,6 +177,47 @@ $ phino dataize --max-steps=50 problem.phi
 [ERROR]: Dataization did not finish before reaching the limit of steps: --max-steps=50
 ```
 
+## Morph
+
+Dataization insists on bytes. Morphing 𝕄 asks a different question: evaluate
+as far as the object model allows, without demanding data. It resolves Φ
+against the universe, peels dispatches and applications through
+normalization, fires whichever atoms sit under a dispatch, and stops at the
+first formation it reaches, handing that formation back untouched. The
+`morph` command runs 𝕄 on its own:
+
+```bash
+$ cat two.phi
+⟦
+  bytes(data) ↦ ⟦ φ ↦ data ⟧,
+  number(as-bytes) ↦ ⟦ φ ↦ as-bytes, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧,
+  φ ↦ 5.plus( 6 ).plus( 7 )
+⟧
+$ phino dataize --sweet --hide-rho two.phi
+40-32-00-00-00-00-00-00
+$ phino morph --locator=Q.φ --sweet --hide-rho two.phi
+⟦ x ↦ 7, λ ⤍ L_number_plus ⟧
+```
+
+The inner `5.plus( 6 )` fires, because `.plus` is dispatched on its result,
+and `11` lands in the `ρ` hidden by `--hide-rho`. The outer application is
+saturated but bare, so 𝕄 returns it and is finished; firing it is
+dataization's job and takes `dataize` on to `18`.
+
+The default locator `Q` morphs the whole top formation, which 𝕄 returns
+unchanged, so `--locator` is how one aims 𝕄 at a subterm, exactly as in
+`dataize`. Unlike 𝔻, 𝕄 is total: where no formation is reachable the answer
+is the terminator `⊥`, printed rather than reported as a failed run:
+
+```bash
+$ phino morph --locator=Q.x <<< '⟦ x ↦ ξ ⟧'
+⊥
+```
+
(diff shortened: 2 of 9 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (450 >= 100); -8 for way too many hits-of-code (450 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2354](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1118, opened 2026-09-08
Treat a bare meta-variable as an anonymous match

### LINKED ISSUE #218: `𝜏?` syntax would be helpful (for matching what's not used later)

yegor256:
A meta-variable written with no index is now anonymous. It matches whatever term stands in its place, each occurrence on its own, and nothing can refer back to it. So `⟦ 𝜏 ↦ 𝑒, 𝜏 ↦ 𝑒 ⟧` is a legal pattern asking for any two attributes, where the indexed spelling `⟦ 𝜏1 ↦ 𝑒1, 𝜏1 ↦ 𝑒1 ⟧` is rejected as a duplicated attribute and `⟦ 𝜏1 ↦ 𝑒1, 𝜏2 ↦ 𝑒2 ⟧` forces you to name four captures the result never mentions.

That naming chore is what #218 asked to get rid of. Rather than a new sigil like `𝜏?`, this follows the shorter option the issue itself proposed: bare means "whatever", indexed means "and I'll use it later". An anonymous meta still lands in the substitution — keyed by the kind and the offset it was parsed at — so the rewriter's pattern round-trip keeps working untouched, which is where the earlier attempt in #675 got stuck. Since nothing can name one, writing an anonymous meta in a `result`, `when`, `where`, `having`, a premise or `--when` is a mistake in the rule, and it is reported as the rule loads instead of quietly failing to match later.

This is a breaking change: an unindexed meta used to be a named variable with an empty suffix, and rules that reference one need an index now. The built-in rules and the test packs are migrated here, with bare metas kept only where nothing reads them back — `dc`, `dd`, `dl`, `miss` and `mad` now read the way the issue wanted. `README.md` documents the two spellings.

Closes #218


maxonfjvipon (approved):
Read the whole diff. The design holds up: anonymous metas are legal only inside
a `pattern`/`match`/`e-match`/`c-match`, `referenceless` rejects them in every
other field as the rule loads, and `Builder` refuses to resolve a slot that the
substitution at hand did not bind. All six new constructors are handled
everywhere they can reach — `CST`, `Printer`, `Misc`, `Bytes`, `Rewriter`,
`Rule`, `Builder` — and `metasWithPrefix` returning `[Expression]` keeps a bare
`𝑛` under the same normal-form check as `𝑛1`, which is the subtle part and is
right. Coverage is thorough: every rejection site has a `YamlSpec` case, the
independence of two same-kind slots is pinned in `MatcherSpec`, and `--when` is
covered end to end in `CLISpec`. `mad`/`maad`/`miss`/`amiss` read considerably
better for it.

Two notes, neither blocking — both look like follow-up issue material rather
than something to change here.

**1. A slot is unique per parsed string, but two parsed strings share one
substitution.** `Slot` identity is `(kind, parse offset)` from `getOffset` in
`metaVar`, so uniqueness holds only within one scalar. `matchExpressionBy` then
does `combineMany seed (matcher ptn expr)` (`src/Rule.hs:385`), and for
morphing and dataization the seed is `matchExpression' rule.ematch univ` while
`ptn` is `rule.match` (`src/Dataize.hs:197`, `src/Dataize.hs:318`) — two
separately parsed fields, both starting at offset 0, both exempt from
`referenceless`, and the README names all of `match`/`e-match`/`c-match` as
places a bare meta is allowed. A bare meta of the same kind at the same offset
in both fields would land on one `Anon (Slot kind off)` key and `combine` would
force the two unrelated captures equal, silently dropping the match. Nothing
shipped hits it — every `e-match`/`c-match` today is `𝑒0`, `𝑘0` or `Φ` — and
these rule sets are embedded, not user-supplied, so it is not reachable from
the outside. It is a trap for whoever writes the next inference rule, though.

**2. The paths that anonymous metas must never reach answer "no" instead of
failing.** `_isFormation` and `_matches` (`src/Rule.hs:210`, `src/Rule.hs:220`)
got no `ExAny` clause and `numToInt` (`src/Rule.hs:116`) no `AnyIndex`/`BiAny`
clause, so a slot arriving there falls through to `[]`/`Nothing` rather than
saying anything. That is correct today only because `referenceless` covers
every `when`/`having`/`where` field — an invariant held by a hand-kept list of
record fields rather than by a type. Same shape as (1): one field added without
a `referenceless` call and the failure is a rule that quietly stops matching.
Both would be closed by giving the slot space a per-field origin, or by
splitting the pattern-only terms from the referencing ones in the types.

Smaller, take or leave: `printMeta` drops the offset (`src/Printer.hs:109`) and
so does the LaTeX renderer, which means the README's own headline pattern
`⟦ 𝜏 ↦ 𝑒, 𝜏 ↦ 𝑒 ⟧` reports two rows labelled `t` and two labelled `e` under
`match`, and `explain` typesets it as `[[ \tau -> e, \tau -> e ]]` — the exact
shape the README says would be rejected as a duplicated attribute. The subst
printer documents the trade-off deliberately; the LaTeX side reads more like a
consequence than a choice.

Worth a line in the release notes: a bare `𝑒` used to be a named meta called
`e` that a `result` could read back, and now is not. The failure is loud and
the README explains it, so this is only about telling users with rule files of
their own.

Approving.

maxonfjvipon:
All 24 checks green on `9ee7bafb`, merges cleanly into `master`. Reviewed in full — the two design notes above are follow-up issue material, not changes for this branch.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1770 >= 100); -8 for way too many hits-of-code (1770 >= 400). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2362](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@maxonfjvipon Thanks for the review! You've earned +20 points for this: +12 as a basis; +16 for the 1770 hits-of-code that you reviewed; -8 for absolutely no comments posted.  Your running score is [+1433](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1119, opened 2026-09-08
feat(dataize): replace built-in atoms with an `--atoms` registry

### LINKED ISSUE #1113: Built-in atoms in src/Dataize.hs should be replaced by an external --atoms=a.json script registry
Today the λ functions phino can fire are baked into the binary. `atom`
at `src/Dataize.hs:498` pattern-matches on eleven fixed names
(`L_number_plus`, `L_bytes_eq`, `L_bytes_slice`, …) and falls through to
`throwIO (Stuck func)` at `src/Dataize.hs:563` for everything else. An EO
program that names any other λ cannot be dataized until somebody edits
Haskell and ships a new release.

That is the wrong place for this knowledge. Which atoms exist is a
property of the object model being dataized, not of the calculus tool.
Hard-coding them pins phino to one snapshot of EO's primitives, forces a
rebuild for every dialect or experiment, and gives phino no way to say
"here is my runtime" to a user who brought their own.

The proposal: delete every `atom` clause, leave the `Stuck` fallback as
the only behaviour for an unregistered name, and add `--atoms=FILE` to
`dataize` (`optAtoms` in `src/CLI/Parsers.hs`, a field on `OptsDataize`
at `src/CLI/Types.hs:76`, threaded into `DataizeContext`). The file is a
JSON registry keyed by λ name:

```json
{
  "L_bytes_eq": {
    "rt": "js",
    "script": "const fs = require('fs'); ..."
  }
}
```

When 𝔼 (`_evaluate`, `src/Dataize.hs:596`) reaches a λ found in the
registry, phino writes `script` to a temporary file, runs it as a POSIX
process under the interpreter named by `rt` — only `js` for now, meaning
`node` — with the λ name as the first command-line argument:

```
node /tmp/phino-atom-4f2a.js L_bytes_eq
```

The name matters: one script may be registered under several λ names and
branch on `process.argv[2]`, which is where node puts it. Then phino feeds
the script JSON on stdin:

```json
{
  "b": "⟦ x ↦ 6, λ ⤍ L_number_plus ⟧",
  "s": "⟦ bytes(data) ↦ ⟦ φ ↦ data ⟧, … ⟧"
}
```

`b` is the formation being evaluated, `s` is the universe Φ. The script
writes one JSON object to stdout:

```json
{ "n": "11" }
```

phino parses `n` back into an `Expression` and hands it to 𝔼 as the
atom's raw result, normalizing it exactly as it does today, so
`--evaluations`, `--partial` and `--max-steps` keep working unchanged. A
non-zero exit, unparsable output or a missing `n` fails the run with the
script's stderr in the message. An unknown `rt` fails at option-parsing
time, before any dataization starts.

A script gets at the parts of `b` by calling phino again, so no API has
to be exposed:

```js
const fs = require('fs');
const { execFileSync } = require('child_process');
const atom = process.argv[2];
if (atom !== 'L_bytes_eq') throw new Error(`unsupported atom ${atom}`);
const { b, s } = JSON.parse(fs.readFileSync(0, 'utf8'));
const dataize = (expr) => execFileSync(
  'phino',
  ['dataize', '--atoms=a.json', '--quiet=false', '/dev/stdin'],
  { input: s.replace('⟦', `⟦ tmp ↦ ${expr},`), encoding: 'utf8' }
).trim();
const equal = dataize(`${b}.ρ`) === dataize(`${b}.b`);
process.stdout.write(JSON.stringify({ n: equal ? 'Φ.true' : 'Φ.false' }));
```

That last snippet shows the one part that still needs a decided
convention: how a script asks phino to dataize a sub-expression *inside*
the given universe. `_dataize` at `src/Dataize.hs:448` already does this
internally by prepending a synthetic binding to the universe and
relocating; the external contract should expose the same trick rather
than leave every script to reinvent the `s.replace` hack above.

Fallout worth naming: `README.md:95-165` documents the built-in atoms and
has to be rewritten around `--atoms`; `test/DataizeSpec.hs` and
`test/CLISpec.hs` are the only places that reference `L_*` names and need
a fixture registry instead; and #1108, which asks for an `atoms` command
listing the λ functions phino implements, becomes moot, since the answer
would be "none, read your own JSON file".

yegor256:
Closes #1113.

`atom` in `src/Dataize.hs` pattern-matched eleven fixed λ names and fell
through to `Stuck` for everything else, so which atoms exist was pinned to
one snapshot of EO's primitives. It is now read from outside the binary.

## What changed

- **`src/Atoms.hs`** (new) — reads the JSON registry given with
  `--atoms=FILE`, keyed by λ name:

  ```json
  { "L_number_plus": { "rt": "node", "script": "const fs = require('fs'); ..." } }
  ```

  and fires an entry as a POSIX process under the executable its `rt` names
  (`node` is the only one for now), with the λ name as the first command-line
  argument so one script may stand for several names and branch on
  `process.argv[2]`. The script is fed `{"b": …, "s": …}` on stdin — the
  formation being evaluated and the universe Φ — and answers with `{"n": …}`
  on stdout, whose 𝜑-expression becomes the atom's raw result, normalized by 𝔼
  exactly as a built-in answer used to be, so `--evaluations`, `--partial` and
  `--max-steps` keep working unchanged.
- **Every `atom` clause is gone**, and with it `_dataize`, `asNumber`,
  `asInt`, `boolean` and `bitwise`. A name the registry does not carry leaves
  `Stuck` as the only behaviour.
- **`--inside=EXPRESSION`** on `dataize` and `morph` — the decided convention
  the issue asked for. `insideUniverse` binds the expression to a fresh
  synthetic attribute of the universe, normalizes it there and aims the run at
  it: the very trick `_dataize` played internally, now the way a script asks
  phino to reduce a part of the `b` it was handed, instead of splicing it into
  the text of `s` by hand. It cannot be combined with `--locator`.
- **Failure modes** — a non-zero exit (message carries the script's stderr),
  output that is not JSON, a missing `n`, an `n` that does not parse: each
  fails the run. An unknown `rt`, a malformed registry or a missing registry
  file fails when the file is read, before the input is even parsed.
- **`README.md`** documents the registry, the wire protocol, `--inside` (with
  a worked `L_number_plus` that reduces its own operands) and the reshaped
  `--partial` story.
- **Tests** — `test-resources/atoms/primitives.js` is a fixture registry
  (number `plus`/`times`/`div`/`gt`, `bytes` `eq`/`not`, with `number.eq`
  composed out of `L_bytes_eq` the way `eq.eo` composes it), wired in through
  `test/Fixtures.hs`; `test/AtomsSpec.hs` covers registry parsing and the
  firing protocol including every failure path; `DataizeSpec` and `CLISpec`
  drive that registry instead of the built-in names, and `CLISpec` gains
  `--atoms` and `--inside` blocks.

## Consequences worth naming

- **An atom no longer reduces its own operands.** `fire`/`ml` hand 𝔼 the
  formation with its arguments unreduced, and the built-ins used to dataize
  them internally. An external script has to ask phino instead, which is what
  `--inside` is for — the README's `L_number_plus` does exactly that, and it
  takes EO's `Factorial` to `6` through recursive `phino` invocations. It
  costs a process per operand, so a registry that wants speed will want a
  longer-lived protocol one day.
- **`--partial` now parks unregistered names only.** "A known atom whose
  input reaches an unknown atom" is no longer a phino-side case, since inputs
  are the script's business; the placeholder story (`⟦ λ ⤍ Sym_arg_0 ⟧`, an
  operation deliberately left out of the registry) is unchanged and still
  parks. There is currently no way for a script to report "I am stuck" — worth
  a follow-up if the EO lowering needs it.
- **The primitive-semantics cases are gone** from `DataizeSpec` (byte
  `and`/`or`/`concat`/`size`/`slice`/`right`, the int32 and non-8-byte
  rejections): they asserted behaviour phino no longer owns, and keeping them
  would mean maintaining an EO runtime inside phino's test resources. What
  replaces them tests the mechanism, plus the end-to-end cases the fixture can
  still answer for.
- **`Bytes.hs` is left as it is, on purpose.** `btsAnd`, `btsOr`, `btsNot`,
  `btsConcat`, `btsSlice` and `btsShift` existed only to implement the byte
  atoms this PR deletes, so nothing in `src/` calls them any more (their
  `BytesSpec` cases still do). Stripping them is a cleanup the issue does not
  ask for and it touches a module unrelated to the change, so it is left for
  you to call — say the word and I will take them out.
- **The atom-firing specs need `node`.** They are marked pending, not red,
  where it is missing; every GitHub runner has it.

## How it was checked

`cabal test --ghc-options=-Werror` on GHC 9.6.7: 2310 examples, 0 failures
(and the same suite pending-not-failing with `node` off the `PATH`).
`fourmolu 0.17.0.0 --mode check src app test`, `hlint src app test` and
`markdownlint-cli2` clean; `make coverage` reports 81% against the 80%
threshold. The binary was also driven by hand: the registry protocol,
`--inside`, `--partial`, `--evaluations`, a payload well past a pipe's buffer,
and EO's `Fahrenheit` and `Factorial` examples through recursive scripts.

https://claude.ai/code/session_01GmCGnQHJyTAw6whNeZpSuV

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,6 +75,17 @@ Matching (`Matcher.hs`) produces `[Subst]` — a list of
 `Map Text MetaValue` — and conditions filter that list. `Builder.hs` then
 applies a substitution to a result template.
 
+### Atoms live outside the binary
+
+`phino` implements no λ function. `Atoms.hs` reads a JSON registry of them
+(the `--atoms` option) and fires each one as a POSIX process under the
+interpreter its `rt` names, feeding it the formation and the universe as JSON
+on stdin and reading the 𝜑-expression it answers with back from stdout. A λ
+name the registry does not carry gets stuck, which is what `--partial` parks
+on. Because a script cannot reduce its own operands, it asks `phino` for them
+with `--inside`, which binds an expression to a synthetic attribute of the
+universe and aims the run at it (`insideUniverse` in `Dataize.hs`).
+
 ### Dependency inversion for circular imports
 
 `Deps.hs` exists solely to break the cycle
--- a/README.md
+++ b/README.md
@@ -109,15 +109,15 @@ registry given with `--atoms`, keyed by λ name:
 ```json
 {
   "L_number_plus": {
-    "rt": "js",
+    "rt": "node",
     "script": "const fs = require('fs'); ..."
   }
 }
 ```
 
-The `rt` field names the interpreter the `script` is written for. Only `js` is
-supported for now, meaning `node`; a registry naming any other runtime is
-refused when the file is read, before dataization starts.
+The `rt` field names the executable the `script` is run under. Only `node` is
+supported for now; a registry naming any other runtime is refused when the file
+is read, before dataization starts.
 
 When 𝔼 reaches a λ function the registry carries, `phino` writes its `script`
 to a temporary file and runs it as a POSIX process under that interpreter, with
--- a/README.md
+++ b/README.md
@@ -100,6 +100,114 @@ $ phino dataize hello.phi
 68-65-6C-6C-6F
 ```
 
+### Atoms
+
+Which λ functions exist is a property of the object model being dataized, not
+of the calculus, so `phino` implements none of them. They come from a JSON
+registry given with `--atoms`, keyed by λ name:
+
+```json
+{
+  "L_number_plus": {
+    "rt": "js",
+    "script": "const fs = require('fs'); ..."
+  }
+}
+```
+
+The `rt` field names the interpreter the `script` is written for. Only `js` is
+supported for now, meaning `node`; a registry naming any other runtime is
+refused when the file is read, before dataization starts.
+
+When 𝔼 reaches a λ function the registry carries, `phino` writes its `script`
+to a temporary file and runs it as a POSIX process under that interpreter, with
+the λ name as the first command-line argument:
+
+```text
+node /tmp/phino-atom-4f2a.js L_number_plus
+```
+
+The name matters: one script may be registered under several λ names and branch
+on it, which is where `node` puts it — `process.argv[2]`. The script is then
+fed one JSON object on `stdin`:
+
+```json
+{
+  "b": "⟦ x ↦ Φ.number( as-bytes ↦ … ), ρ ↦ ⟦ … ⟧ ⟧",
+  "s": "⟦ bytes ↦ ⟦ … ⟧, number ↦ ⟦ … ⟧, φ ↦ … ⟧"
+}
+```
+
+Here `b` is the formation being evaluated, with its λ binding removed so that
+the script may dispatch on it, and `s` is the universe Φ. Both are canonical
+𝜑-calculus on a single line — no syntax sugar, whatever `--sweet` says about
+the output of the run — so a script never has to know about `phino`'s sugar in
+order to find a datum: every byte array is spelled out as a Δ binding.
+
+The script writes one JSON object to `stdout`:
+
+```json
+{ "n": "11" }
+```
+
+The `n` field is the 𝜑-expression the atom answers with, in any syntax
+`phino`'s parser reads — syntax sugar included, so the `11` above and the
+`Φ.number( … )` it stands for are the same answer. `phino` parses it back
+and hands it to 𝔼 as the atom's raw result, normalizing it exactly as it
+normalizes anything else, so `--evaluations`, `--partial` and `--max-steps`
+keep working unchanged. A non-zero exit, output that is not JSON, a missing
+`n` or an `n` that does not parse fails the run, with the script's own
+`stderr` in the message.
+
+A λ name the registry does not carry has no λ function at all, so 𝔼 gets stuck
+on it. Without `--atoms` the registry is empty and every atom gets stuck.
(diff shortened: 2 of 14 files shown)


0crat:
@yegor256 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`claude/clever-cannon-asqwvv`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`1118`". Your running score is [+2437](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1948 >= 100); -8 for way too many hits-of-code (1948 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2366](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1122, opened 2026-09-08
Let the `--atoms` registry name an executable, not only a script

### LINKED ISSUE #1120: The --atoms registry accepts only rt=node with an inline script; an executable at a path cannot be registered
The `--atoms` registry can only carry a script, never a program. `Runtime`
in `src/Atoms.hs:67` has the single constructor `RtNode`, and
`instance FromJSON Atom` at `src/Atoms.hs:128` demands both `rt` and
`script`, so every entry has to spell its λ function out as inline
JavaScript text, which `fireAtom` (`src/Atoms.hs:171`) then writes to a
temporary `.js` file and hands to `node`.

That leaves out the obvious case. An atom already compiled — a Go or Rust
binary, a `chmod +x` shell script, a wrapper the user keeps in their own
tree — has no interpreter to name and no source to inline, and pasting a
whole program into a JSON string is not a way to distribute one. Since the
protocol between phino and an atom is already a POSIX process with the λ
name in `argv[1]` and JSON on stdin/stdout, nothing about the calculus side
needs an interpreter at all.

Add a second runtime, `"rt": "exec"`, that takes `path` where `node` takes
`script`:

```json
{
  "L_number_plus": { "rt": "exec", "path": "/opt/eo/atoms/number-plus" }
}
```

Firing it should `proc` the file at `path` directly, with the λ name as
the first argument and the same stdin/stdout JSON both ways — no
interpreter, no temporary file, no `extension`. Concretely: give `Atom`
two shapes so `path` and `script` cannot both be given (or both be
missing), let `FromJSON Atom` pick by `rt`, and in `fireAtom` skip the
`withTemp` that stages the script when the entry is an `exec` one. A
`path` that does not exist or is not executable should fail as `NoRuntime`
does today, when the registry is read rather than mid-dataization.

yegor256:
The `--atoms` registry now takes a second kind of entry. Next to
`{"rt": "node", "script": "..."}` an entry may say
`{"rt": "exec", "path": "/opt/eo/atoms/number-plus"}`, and phino spawns
that file directly, as the executable it is: no interpreter, no
temporary file, no staging. Everything else stays the same — the λ name
arrives as the first argument, the formation and the universe arrive as
JSON on stdin, and the answer comes back under `n` on stdout.

The reason is that an atom which is already a program had nowhere to go.
A Go or Rust binary, a `chmod +x` script, a wrapper someone keeps in
their own tree — none of those has an interpreter to name or source to
inline, and pasting a whole program into a JSON string is not a way to
ship one.

`Atom` grew two shapes instead of a record, so `path` and `script` can
no longer both be given or both be missing, and a `path` that names no
file (or a file nobody may run) is refused where the registry is read,
next to the unknown runtimes, rather than half-way through a run. The
new spec cases run a small shell script as the atom, so they are pending
on Windows, where no shebang runs.

Closes #1120

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,9 +115,9 @@ registry given with `--atoms`, keyed by λ name:
 }
 ```
 
-The `rt` field names the executable the `script` is run under. Only `node` is
-supported for now; a registry naming any other runtime is refused when the file
-is read, before dataization starts.
+The `rt` field names the interpreter the `script` is run under. Only `node` is
+supported for now; a registry naming any other interpreter is refused when the
+file is read, before dataization starts.
 
 When 𝔼 reaches a λ function the registry carries, `phino` writes its `script`
 to a temporary file and runs it as a POSIX process under that interpreter, with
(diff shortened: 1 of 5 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (280 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2390](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1123, opened 2026-09-08
Speak the paper's letters to every atom and make `serve` a flag

### LINKED ISSUE #1121: Excessive Processes Spawned for Each λ fire in The --atoms registry
Every λ fire costs a process.
`fireAtom` (`src/Atoms.hs:169`) writes the script to a temporary file, spawns the interpreter with the λ name in `argv`,
  feeds one JSON object on stdin, reads one on stdout, and waits for the exit (`src/Atoms.hs:191`–`200`).
The object carries the whole universe every time (`payload`, `src/Atoms.hs:245`),
  and because a script can only get at its operands by running `phino` again, one fire is typically three processes:
  the interpreter plus one `phino dataize` per operand, each of them re-parsing that universe.

Measured on a fragment with seven fires under 0.0.117: 21 processes, 1.35 s, with `node` alone costing 57 ms to start.
A JVM-hosted registry (the EO compiler wants to serve the atoms from the Maven process it already runs in)
  would pay the same price per fire whatever #1120's `exec` points at, so the runtime is not the problem; the process per fire is.

Add a runtime that stays resident for the whole run:

```json
{
  "L_number_plus": { "rt": "serve", "path": "/opt/eo/atoms" },
  "L_number_times": { "rt": "serve", "path": "/opt/eo/atoms" }
}
```

`phino` starts the program at `path` once, on the first fire, keeps its stdin and stdout open until the run ends, and kills it on exit.
Both directions carry one JSON object per line.
The first line hands the program the universe, and each fire after that is one request and one reply, paired by `id`:

```
  {"𝑒": "⟦ bytes ↦ … ⟧"}
  {"id": 1, "λ": "L_number_plus", "𝑏": "⟦ x ↦ …, ρ ↦ … ⟧"}
  {"id": 1, "𝑛": "Φ.number( … )"}
```

The λ name moves from `argv` into the request, since one process now serves every name the registry maps to it;
  several entries naming the same `path` share one process.
`phino` fires atoms sequentially, so one request is in flight at a time, and `id` is there for pairing and for pipelining later.
A line that is not JSON, a reply without `𝑛`, an `𝑛` that does not parse, or a pipe the program closes early
  fails the run the way a bad answer does today, with the program's stderr in the message.
A `script` may be accepted alongside `path` and written to a temporary file once per run, for a resident `node` program.

The keys are the letters of the evaluation rule in the calculus paper (objectionary/calculus-paper, `sections/operators.tex`, Definition "Evaluation"):
  𝔅 × 𝔈 × 𝔖 ⇀ 𝔑 × 𝔖, where ⟨𝑒₁, 𝑠₂⟩ = 𝑓(⟦ 𝐵₁, λ ⤍ 𝑓, 𝐵₂ ⟧, 𝑒, 𝑠₁) and 𝑒₁ normalizes to 𝑛.
There 𝑏 is the atom being evaluated, 𝑒 is the universe it is evaluated in, 𝑛 is the normal form it evaluates to, λ marks the function, and 𝑠 is the state.
The current payload already spells the atom `b` and the answer `n`, but it calls the universe `s`, which in the paper is the state.
The new protocol should use the paper's letters as they are, `𝑏`, `𝑒`, `𝑛` and `λ`, so that a reader of the paper reads the wire format without a legend;
  the one-shot `node` protocol may switch `s` to `𝑒` in the same change.

A `path` that does not exist or is not executable should be refused when the registry is read, as #1120 asks for `exec`.
With this in place the EO side needs no JavaScript at all: one fragment costs one `phino` process and a few lines over a pipe.

yegor256:
Every atom program is now talked to the same way, whether phino starts it for one fire or keeps it for the run: one JSON object per line, in the letters of the evaluation rule of the calculus paper, 𝔼(𝑏, 𝑒, 𝑠) = 𝑛. The universe goes under `𝑒` in a line of its own, then a request carries an `id`, the λ name under `λ` and the formation under `𝑏`, and the program answers with the same `id` under `𝑛`. A one-shot program has its stdin closed behind the request, so it may slurp its input or read it line by line, and the very same program works both ways. The λ name is no longer passed on the command line, since every request names it.

Whether a program is kept for the run is a boolean `serve` on any registry entry, off by default, rather than a runtime of its own, so a `node` script can be resident just as an `exec` file can. Every λ name registered on the same program is served by one process; the runners close the registry when the run is over, closing the program's stdin as its cue to quit and terminating it if it has not quit within a second. A reply that is not JSON, lacks `𝑛`, answers another `id`, a program that quits without answering, or a one-shot program that exits non-zero fails the run with the program's stderr in the message.

The keys of the registry are regular expressions over λ names, tried top to bottom in the order the file lists them, and the first one matching the whole name wins. So one entry such as `".*"` with `"serve": true` stands for every atom an object model brings, instead of the same program being spelled once per name, while a plain name still means that one atom. Since aeson forgets the order of keys, the registry is walked token by token; a key that is not a regular expression, a file that is not a JSON object, or text after the object are refused where the file is read.

The README documents the protocol, the flag and the keys, and the test fixture `primitives.js` reads lines with `readline`, so it stands as an example of a program that runs either way. The old `b`/`s`/`n` payload is gone, so every registered script has to move to the new lines.

Closes #1121

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ $ phino dataize hello.phi
 
 Which λ functions exist is a property of the object model being dataized, not
 of the calculus, so `phino` implements none of them. They come from a JSON
-registry given with `--atoms`, keyed by λ name:
+registry given with `--atoms`, keyed by regular expressions over λ names:
 
 ```json
 {
@@ -177,22 +177,31 @@ the run. So does a reply that is not JSON, carries no `𝑛`, answers another
 `id`, or an `𝑛` that does not parse, or a program that quits without
 answering — always with the program's own `stderr` in the message.
 
-A λ name the registry does not carry has no λ function at all, so 𝔼 gets stuck
-on it. Without `--atoms` the registry is empty and every atom gets stuck.
+Each key of the registry is a regular expression, and it must match the whole
+λ name, so a plain name such as `L_number_plus` means that one atom and nothing
+else, while `L_number_.*` stands for every atom of `number`. When 𝔼 reaches a
+λ function, the keys are tried top to bottom, in the order the file lists them,
+and the first one that matches is the entry fired, so a key placed above
+another hides whatever the two have in common. A key that is not a regular
+expression is refused where the registry is read.
+
+A λ name no key matches has no λ function at all, so 𝔼 gets stuck on it.
+Without `--atoms` the registry is empty and every atom gets stuck.
 
 One process per fire is where a program that is slow to start — a JVM, say —
 spends most of the run. An entry saying `serve` has `phino` start its program
 once, on the first fire, and keep it for the rest of the run, whether it is a
-`script` or a `path`:
+`script` or a `path`. Together with a key that matches many names, this is how
+one program stands for a whole object model without being spelled once per
+atom:
 
 ```json
 {
-  "L_number_plus": {
-    "rt": "exec",
-    "path": "/opt/eo/atoms/resident",
-    "serve": true
+  "L_bytes_eq": {
+    "rt": "node",
+    "script": "const readline = require('readline'); ..."
   },
-  "L_number_times": {
+  ".*": {
     "rt": "exec",
     "path": "/opt/eo/atoms/resident",
     "serve": true
@@ -200,8 +209,9 @@ once, on the first fire, and keep it for the rest of the run, whether it is a
 }
 ```
 
-Every λ name registered on the same program is served by the same process, so
-there is one of it, however many atoms it stands for. The lines are the same:
+Every λ name registered on the same program, under one key or under several,
+is served by the same process, so there is one of it, however many atoms it
+stands for. The lines are the same:
 the program reads request after request off its `stdin`, each with the next
 `id`, and answers each in turn. The universe is told again only when a fire
 comes with a different one; the program keeps the last one it was told. When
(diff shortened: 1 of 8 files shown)


yegor256:
Redesigned per the follow-up on the issue: the 𝑏/𝑒/λ/𝑛 lines now go to every program, one-shot or resident, and `serve` is a boolean on any entry rather than a runtime, so a `node` script can be kept for the run too. The argv λ name is gone with it.

yegor256:
Pushed one more commit: registry keys are now regular expressions matched against the whole λ name, tried top to bottom, so a single `".*"` entry with `"serve": true` can stand for every atom. Happy to split it out if you would rather review it apart.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (1307 >= 100); -8 for way too many hits-of-code (1307 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2326](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1125, opened 2026-09-09
Enter the bindings of the answer under morph --deep

### LINKED ISSUE #1124: `morph` stops at the first formation (`mf`) and never enters its bindings or fires their λ: add `--deep` (src/Dataize.hs:246)
`morph` (`src/Dataize.hs:246`, run by `runMorph` in `src/CLI/Runners.hs:226`) morphs only the expression at the locator and stops at the first formation the `mf` rule (`resources/morphing.yaml:3`) hands back. The bindings of that formation are never entered, and its own λ is never evaluated, because `ml` fires a λ only when a dispatch lands on the formation. With a universe holding `bytes`, `number` (`times ↦ ⟦ x ↦ ∅, λ ⤍ L_number_times ⟧` served by `--atoms`), a foreign `bar ↦ ⟦ x ↦ ∅, λ ⤍ L_bar ⟧`, and

```
demo ↦ ⟦ foo ↦ ⟦ n ↦ Φ.number( Φ.bytes( ⟦ λ ⤍ S1 ⟧ ) ), φ ↦ Φ.bar( ξ.n.times( 5 ).times( 7 ) ) ⟧ ⟧
```

`morph --partial --atoms=… --inside='Φ.demo.foo'` answers `foo` as written, nothing inside touched, and aimed at the argument, `--inside='Φ.number( Φ.bytes( ⟦ λ ⤍ S1 ⟧ ) ).times( 5 ).times( 7 )'`, it answers `⟦ x ↦ 7, λ ⤍ L_number_times ⟧`: the inner `times` fired on the way (`ml`, while `.times( 7 )` was being dispatched), the outer one is a formation, so `mf` returns it with its λ unevaluated.

objectionary/eo#8548 needs the whole program reduced as far as the registry allows while it stays a program: every application that can be evaluated replaced by the atom's answer, in lazy positions too, and every other one left as it was written. Neither command gives that today. `morph` reduces the head only, and `dataize` follows the one path dataization demands and collapses it into bytes or a residual, so the argument of `bar`, which nothing demands, is never reduced and the object structure is lost.

Proposal: a `--deep` switch for `morph`. After 𝕄 returns the located formation, walk its bindings recursively; for each binding run 𝕄 and, while the formation it reaches carries a λ the registry serves, evaluate it the way `ml` does and morph the answer again; under `--partial` an unserved λ parks. A binding on whose spine at least one atom fired is replaced by the last answer; a binding where none fired stays as written, so `Φ.number( … )` and the parked `Φ.bar( … )` keep their names and the result is still the same program, and only its own sub-expressions are walked. `--max-steps` bounds the walk. For the example, `morph --deep --partial --atoms=… --inside='Φ.demo.foo'` would answer

```
⟦ n ↦ Φ.number( Φ.bytes( ⟦ λ ⤍ S1 ⟧ ) ), φ ↦ Φ.bar( Φ.number( as-bytes ↦ Φ.bytes( data ↦ ⟦ λ ⤍ S3 ⟧ ) ) ) ⟧
```

with `S2` and `S3` minted by the served script.

yegor256:
Adds a `--deep` switch to `morph`. Without it, 𝕄 stops at the first
formation it reaches and hands back everything that formation holds
exactly as written, so a λ the `--atoms` registry serves never fires
unless something demands it. With it, the walk carries on into those
bindings: each one is morphed in the scope of its siblings, and while
the formation it lands on carries a served λ, that λ fires and its
answer is morphed again. The last answer stands in the place of what
computed it, while a binding where nothing fired stays as written, so
names survive and the result is still the same program.

The eo-lowering engine in objectionary/eo#8548 needs this to fold the
constant arithmetic a program holds without dataizing it, which would
end in bytes and lose the program.

One rule is worth a second look: only a saturated formation fires. A
formation still holding a void is a method waiting for its argument,
not an application waiting to be computed, so the method table of an
object model passes through untouched. Firing those turned every
`times(x) ↦ ⟦ λ ⤍ L_number_times ⟧` into ⊥ in the first cut. An
unserved λ is not fired at all, one that gets stuck deeper fails the
run and parks under `--partial`, `--max-steps` bounds the walk, and the
answer joins the chain as a step named `deep`.

Closes #1124

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,6 +404,60 @@ The whole `dataize` option surface applies unchanged — `--atoms`, `--inside`,
 `--sequence`, `--headers`, `--steps-dir`, `--evaluations`, `--partial`,
 `--max-steps`, `--shuffle`/`--seed`, `--output`, `--focus` and the rest.
 
+### Deep morphing
+
+𝕄 stops at the first formation it reaches and hands its bindings back as they
+were written, since firing a bare λ is dataization's job, and `dataize`
+follows the one path dataization demands and ends in bytes. What a program
+holds but nothing demands — the argument of an atom the registry does not
+serve, for one — is therefore reduced by neither. The `--deep` flag enters it:
+
+```bash
+$ cat gap.phi
+⟦
+  bytes(data) ↦ ⟦ φ ↦ ξ.data ⟧,
+  number(as-bytes) ↦ ⟦ φ ↦ ξ.as-bytes, times(x) ↦ ⟦ λ ⤍ L_number_times ⟧ ⟧,
+  bar(x) ↦ ⟦ λ ⤍ L_bar ⟧,
+  demo ↦ ⟦ foo ↦ ⟦ n ↦ 3, φ ↦ Φ.bar( ξ.n.times( 5 ).times( 7 ) ) ⟧ ⟧
+⟧
+$ phino morph --atoms=atoms.json --inside='Q.demo.foo' \
+    --sweet --hide-rho gap.phi
+⟦ n ↦ 3, φ ↦ Φ.bar( n.times( 5 ).times( 7 ) ) ⟧
+$ phino morph --deep --atoms=atoms.json --inside='Q.demo.foo' \
+    --sweet --hide-rho gap.phi
+⟦ n ↦ 3, φ ↦ Φ.bar( 105 ) ⟧
+```
+
+Every binding of the formation is entered, recursively. 𝕄 is asked about the
+term standing there and, where it lands on a saturated formation whose λ the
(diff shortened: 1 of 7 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (393 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2330](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1130, opened 2026-09-09
fix(normalize): exclude ρ void from domain count and guard alpha against ρ-first formations (#1127)

### LINKED ISSUE #1127: `domain` Function Incorrectly Counts `ρ` Void as Positional Slot, Preventing EO Atom Saturation
The `domain` function at `src/Rule.hs:122` counts every binding of a formation except `Δ` and `λ` (its `notAsset` helper at `src/Rule.hs:125-127`), so a `ρ ↦ ∅` void counts as a positional slot. The three rules that resolve an α index use it — `alpha`, `amiss` and `overa` under `resources/normalize/` — so every index shifts by one whenever `ρ` is declared before the real voids:

```
$ phino rewrite --normalize --hide-rho   # ⟦ f ↦ ⟦ ρ ↦ ∅, x ↦ ∅ ⟧( 42 ) ⟧
⟦ f ↦ ⟦ x ↦ ∅ ⟧ ⟧                        # 42 was bound to ρ, x stayed void
$ phino rewrite --normalize --hide-rho   # ⟦ f ↦ ⟦ x ↦ ∅, ρ ↦ ∅ ⟧( 42 ) ⟧
⟦ f ↦ ⟦ x ↦ 42 ⟧ ⟧
```

This is wrong because `ρ` is bound by dispatch, not by application position, and phino's own XMIR reader produces the first shape from real EO code: an atom there declares its receiver as `? > ^`, so `number.plus` reads back as `⟦ ρ ↦ ∅, x ↦ ∅, λ ⤍ L_number_plus ⟧`. A plain `1.plus 1` against such a universe never saturates, since the argument lands on `ρ` and `x` stays void, and `dataize --partial` then runs to `--max-steps` instead of parking the atom. Removing the `ρ` void from the same universe makes it park and print the residual.

The smallest fix is one more clause in `notAsset` at `src/Rule.hs:125-127`, excluding a `ρ` void (`BiVoid AtRho`) from the count the way `Δ` and `λ` are already excluded, so all three α rules agree that positions are counted over the other voids only.

Needed by objectionary/eo#8548, where the universe is the real parsed program rather than a hand-written one.

VasilevNStas:
Fixes #1127.

## Problem

The `domain` function at `src/Rule.hs:122` counted `ρ ↦ ∅` as a positional slot, so every α-index shifted by one whenever `ρ` was declared before the real positional voids:

```
$ phino rewrite --normalize --hide-rho   # ⟦ f ↦ ⟦ ρ ↦ ∅, x ↦ ∅ ⟧( 42 ) ⟧
⟦ f ↦ ⟦ x ↦ ∅ ⟧ ⟧                        # 42 was bound to ρ, x stayed void  ← WRONG
```

This is what XMIR produces for real EO atoms (`number.plus` reads back as `⟦ ρ ↦ ∅, x ↦ ∅, λ ⤍ L_number_plus ⟧`), so `dataize --partial` would run to `--max-steps` instead of parking the atom.

## Root cause

Two issues, not one:

1. **`notAsset`** did not exclude `BiVoid AtRho`, so `domain` counted the ρ void alongside the positional voids.
2. **The `alpha` rule** had no guard against ρ as the target when B1 is empty: the matcher's first split (`B1=[], τ1=ρ`) always satisfied `eq(0, domain([])=0)` regardless of the `notAsset` fix, because `domain([])` is trivially 0.

## Fix

1. `src/Rule.hs` — added `notAsset (BiVoid AtRho) = False` alongside the existing `BiDelta` and `BiLambda` exclusions.
2. `resources/normalize/alpha.yaml` — added a guard `(τ1 ≠ ρ) ∨ (|B1| > 0)`: ρ may still receive a positional argument when it is not the first binding (the existing `alpha-skips-lambda-asset-hits-rho` test, where α1 lands on ρ after λ and x), but never when B1 is empty.
3. New test pack `alpha-rho-void-before-x.yaml` covering the ρ-before-x shape from the report.

## Verification

- Both shapes from the issue now produce the same correct result:

```
$ phino rewrite --normalize --hide-rho --sweet <<< '⟦ f ↦ ⟦ ρ ↦ ∅, x ↦ ∅ ⟧( 42 ) ⟧'
⟦ f(ρ) ↦ ⟦ x ↦ 42 ⟧ ⟧

$ phino rewrite --normalize --hide-rho --sweet <<< '⟦ f ↦ ⟦ x ↦ ∅, ρ ↦ ∅ ⟧( 42 ) ⟧'
⟦ f ↦ ⟦ x ↦ 42 ⟧ ⟧
```

- All 110 Rewriter tests pass, including `alpha-skips-lambda-asset-hits-rho` (ρ as a legitimate positional slot after other bindings).
- `hlint` and `fourmolu --mode check` are clean.
- The 14 pre-existing failures (LocatorSpec + CLI.explain on GHC 9.10.3) are unrelated — they are the HasCallStack backtrace issue tracked in #1100.

### DIFF
--- a/test-resources/rewriter-packs/basic/alpha-rho-middle-skipped.yaml
+++ b/test-resources/rewriter-packs/basic/alpha-rho-middle-skipped.yaml
@@ -4,6 +4,6 @@
 rules:
   basic: ['alpha']
 input: |
-  [[ L> Fn, x -> ? ]](~1 -> Q.y)
+  [[ a -> ?, ^ -> ?, x -> ? ]](~1 -> Q.y)
 output: |-
-  ⟦ L> Fn, x ↦ ∅ ⟧(α1 ↦ Φ.y)
+  ⟦ a ↦ ∅, ρ ↦ ∅, x ↦ ∅ ⟧(x ↦ Φ.y)
--- a/test-resources/rewriter-packs/basic/alpha-rho-void-before-x.yaml
+++ b/test-resources/rewriter-packs/basic/alpha-rho-void-before-x.yaml
@@ -1,0 +1,9 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+rules:
+  basic: ['alpha']
+input: |
+  [[ ^ -> ?, x -> ? ]](~0 -> Q.y)
+output: |-
+  ⟦ ρ ↦ ∅, x ↦ ∅ ⟧(x ↦ Φ.y)
--- a/resources/normalize/alpha.yaml
+++ b/resources/normalize/alpha.yaml
@@ -5,6 +5,15 @@ name: alpha
 pattern: ⟦𝐵1, 𝜏1 ↦ ∅, 𝐵2⟧(α𝑖1 ↦ 𝑒1)
 result: ⟦𝐵1, 𝜏1 ↦ ∅, 𝐵2⟧(𝜏1 ↦ 𝑒1)
 when:
-  eq:
-    - 𝑖1
-    - domain: 𝐵1
+  and:
+    - eq:
+        - 𝑖1
+        - domain: 𝐵1
+    - or:
+        - not:
+            eq:
+              - 𝜏1
+              - ρ
+        - gt:
+            - length: 𝐵1
+            - 0
--- a/resources/normalize/alpha.yaml
+++ b/resources/normalize/alpha.yaml
@@ -9,11 +9,7 @@ when:
     - eq:
         - 𝑖1
         - domain: 𝐵1
-    - or:
-        - not:
-            eq:
-              - 𝜏1
-              - ρ
-        - gt:
-            - length: 𝐵1
-            - 0
+    - not:
+        eq:
+          - 𝜏1
+          - ρ
--- a/src/Rule.hs
+++ b/src/Rule.hs
@@ -125,6 +125,7 @@ numToInt (Y.Domain (BiMeta meta)) (Subst mp) = case M.lookup (Named meta) mp of
   where
     notAsset (BiDelta _) = False
     notAsset (BiLambda _) = False
+    notAsset (BiVoid AtRho) = False
     notAsset _ = True
 numToInt (Y.Literal num) _ = Just num
 numToInt _ _ = Nothing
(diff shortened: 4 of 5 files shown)


VasilevNStas:
@yegor256 
plz review

test-resources/rewriter-packs/basic/alpha-rho-void-before-x.yaml
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+rules:
+  basic: ['alpha']
+input: |
+  [[ ^ -> ?, x -> ? ]](~0 -> Q.y)
```
yegor256, line 7: @VasilevNStas try this too:

```
[[ a -> ?, ^ -> ?, x -> ? ]](~1 -> Q.y)
```

should be rewritten to 

```
[[ a -> ?, ^ -> ?, x -> ? ]](x -> Q.y)
```

works?

test-resources/rewriter-packs/basic/alpha-rho-middle-skipped.yaml
```diff
@@ -0,0 +1,9 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+# SPDX-License-Identifier: MIT
+---
+rules:
+  basic: ['alpha']
+input: |
+  [[ a -> ?, ^ -> ?, x -> ? ]](~1 -> Q.y)
+output: |-
+  ⟦ a ↦ ∅, ρ ↦ ∅, x ↦ ∅ ⟧(x ↦ Φ.y)
```
yegor256: @VasilevNStas also, try this:

```
input: |
  [[ a -> ?, b -> ?, ^ -> ?, c -> ?, d -> ? ]](~2 -> Q.y)
output: |-
  ⟦ a ↦ ∅, b ↦ ∅, ρ ↦ ∅, c ↦ ∅, d ↦ ∅ ⟧(c ↦ Φ.y)
```

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (34) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2376](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@yegor256 Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+2652](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1072, opened 2026-09-05
asNumber crashes with a raw error on non-8-byte operands (asInt is guarded, asNumber is not)


VasilevNStas:
`asNumber` crashes with a raw `error` on non-8-byte operands (asInt is guarded, asNumber is not)

## Problem

`src/Dataize.hs:461-463`:

```haskell
asNumber BtEmpty = Nothing
asNumber bts = Just (either toDouble id (btsToNum bts))
```

`asNumber` passes any non-empty bytes straight into `btsToNum` (`src/Bytes.hs:109-133`), which **throws** `error "Expected 8 bytes for conversion, got N"` (`Bytes.hs:113`) on arrays whose length is not 8. The sibling `asInt` (`Dataize.hs:468-473`) guards with `btsSize bts /= 8 -> Nothing` and returns `Nothing` (⊥) on non-8-byte input — so the two helpers are **asymmetric**: the guarded one yields the terminator, the unguarded one crashes.

The same unguarded `btsToNum` call exists in `src/Functions.hs:66` (`argToNumber`), so the `sum`/`number`/`string` where-functions crash identically.

## Affected call sites

- `Dataize.hs:461-463` — `asNumber` (used by `L_number_plus/times/div/gt/eq`, `Dataize.hs:489-521`).
- `Functions.hs:66` — `argToNumber` (used by `sum` and `_number`).
- `Bytes.hs:113` — `btsToNum` throws on length ≠ 8.
- `Bytes.hs:62` — `btsToWord8 (BtMeta _)` also `error`s (reachable if a meta byte reaches `asNumber`).

## Repro (verified)

```
$ phino dataize --max-steps=60 \
    '⟦ φ ↦ ⟦ λ ⤍ L_number_plus, ρ ↦ ⟦ Δ ⤍ 40-14-00-00-00-00-00-00 ⟧, x ↦ ⟦ Δ ⤍ 68-65 ⟧ ⟧ ⟧'
[ERROR]: Expected 8 bytes for conversion, got 2
CallStack (from HasCallStack):
  error, called at src/Bytes.hs:113:14
```

Also reproduces for a 5-byte operand (`68-65-6C-6C-6F`) with `got 5`, and for `L_number_eq`/`L_number_times`/`L_number_div`/`L_number_gt`.

## Why existing tests don't catch it

[...]

PR #1132, opened 2026-09-09
fix(print): guard short-byte numbers in printer, XMIR writer and string() (#1129)

### LINKED ISSUE #1129: Printing a number whose Δ is not 8 bytes crashes with a raw error (src/CST.hs:273 → src/Bytes.hs:114)
Printing an expression that holds a number whose Δ datum is not exactly eight bytes kills the process with a raw GHC `error`. `btsToNum` (`src/Bytes.hs:110-114`) returns `Either Int Double` but calls `error $ "Expected 8 bytes for conversion, got " ++ show (length bytes)` on any other length, and the φ printer forces it unguarded: `sweetNumber` (`src/CST.hs:273`) pattern-matches on the result, and the two `EX_NUMBER` sites it gates (`src/CST.hs:360` and `src/CST.hs:426`) pass the same bytes in again. The XMIR writer's number comment (`src/XMIR.hs:122`, `either show show (btsToNum bytes)`) has the same problem. On 0.0.118:

```
$ cat n1.phi
⟦ a ↦ Φ.number( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ 21- ⟧ ) ) ⟧
$ phino rewrite n1.phi
[ERROR]: Expected 8 bytes for conversion, got 1
CallStack (from HasCallStack):
  error, called at src/Bytes.hs:114:14 in phn-0.0.118-cc311944:Bytes
```

The same crash comes out with `--sweet`, with `--normalize`, and with `--output=xmir` on a top-level formation. No rewriting rule is involved: the expression is read and printed back, so the crash is in the output stage alone.

This is wrong because a short Δ is a legal expression, not a broken one. Nothing in the grammar promises that a `Φ.number` carries eight bytes, phino reads such a term without complaint, and it can arise from ordinary rewriting. The printer should keep it in its byte form `Φ.number(Φ.bytes(⟦ Δ ⤍ … ⟧))` — exactly what it already does for a NaN with a payload, per the comment above `sweetNumber` — instead of aborting. This is the surviving sibling of #1072: that fix guarded the arithmetic side (`asNumber`, and `argToNumber` at `src/Functions.hs:68`, which now returns a proper `userError`), while the print side was left as it was.

Give `sweetNumber` a leading guard `sweetNumber bts | btsSize bts /= 8 = False`, which makes both `EX_NUMBER` sites fall through to the generic application clause and keep the bytes. `src/XMIR.hs:122` needs its own check to drop the comment when the datum is not eight bytes, and `src/Functions.hs:172` is the third unguarded caller.

VasilevNStas:
Fixes #1129 

## Problem

Printing a `Φ.number` whose Δ datum is not exactly eight bytes kills the process with a raw GHC `error`:

```
$ phino rewrite <<< '⟦ a ↦ Φ.number( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ 21- ⟧ ) ) ⟧'
[ERROR]: Expected 8 bytes for conversion, got 1
```

Three callers force `btsToNum` unguarded:
1. `sweetNumber` in `src/CST.hs:273` — the SWEET printer
2. XMIR number comment in `src/XMIR.hs:122`
3. `string()` build-term in `src/Functions.hs:172`

## Fix

| File | Guard | Effect |
|------|-------|--------|
| `src/CST.hs` | `sweetNumber bts \| btsSize bts /= 8 = False` | Falls through to generic application clause, keeps byte form |
| `src/XMIR.hs` | `_omitComments \|\| btsSize bytes /= 8` | Skips the number comment when datum is not 8 bytes |
| `src/Functions.hs` | `btsSize bts /= 8 -> throwIO (userError ...)` | Proper error with byte count, consistent with `argToNumber` (#1072) |

## Tests

- 2 `sweetNumber` cases: one byte, empty bytes → both `False`
- 2 `PrinterSpec` cases: one byte, three bytes → byte-form round-trip

## Verification

- Full suite: 2350 examples, 13 pre-existing LocatorSpec failures (GHC 9.10.3, unrelated)

### DIFF
--- a/src/CST.hs
+++ b/src/CST.hs
@@ -11,7 +11,7 @@
 module CST where
 
 import AST
-import Bytes (NonFinite, btsToNonFinite, btsToNum, btsToStr)
+import Bytes (NonFinite, btsSize, btsToNonFinite, btsToNum, btsToStr)
 import Data.Maybe (isJust)
 import qualified Data.Text as T
 import qualified Yaml as Y
@@ -270,6 +270,8 @@ expressionToCSTFrom tabs expr = toCST expr (tabs, EOL)
 -- pattern, such as a NaN carrying a payload, is kept in its byte form so that
 -- no bit of it is lost.
 sweetNumber :: Bytes -> Bool
+sweetNumber bts
+  | btsSize bts /= 8 = False
 sweetNumber bts = case btsToNum bts of
   Right dbl | isNaN dbl || isInfinite dbl -> isJust (btsToNonFinite bts)
   _ -> True
--- a/src/Functions.hs
+++ b/src/Functions.hs
@@ -169,7 +169,10 @@ _string :: BuildTermMethod
 _string [Y.ArgExpression expr] subst = do
   expr' <- buildExpressionThrows expr subst
   str <- case expr' of
-    DataNumber bts -> pure (DataString (strToBts (either show show (btsToNum bts))))
+    DataNumber bts
+      | btsSize bts /= 8 ->
+          throwIO (userError (printf "Expected 8 bytes for a number, got %d" (btsSize bts)))
+      | otherwise -> pure (DataString (strToBts (either show show (btsToNum bts))))
     DataString bts -> pure (DataString bts)
     ex ->
       throwIO
--- a/src/XMIR.hs
+++ b/src/XMIR.hs
@@ -22,7 +22,7 @@ module XMIR
 where
 
 import AST
-import Bytes (btsToNum, btsToStr, bytesToBts)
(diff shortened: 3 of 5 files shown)


VasilevNStas:
@yegor256 
please review this pull request

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (29) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2400](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1136, opened 2026-09-09
fix(normalize): exclude bound ρ from domain count (#1134)

### LINKED ISSUE #1134: `domain` still counts a *bound* ρ as a positional slot, so a dispatched atom never saturates (src/Rule.hs:128)
The fix for #1127 (c73cce1, shipped in 0.0.119) excludes only the *void* ρ from the `domain` count. `notAsset` in `src/Rule.hs:122-129` now reads `notAsset (BiVoid AtRho) = False`, so `⟦ ρ ↦ ∅, x ↦ ∅ ⟧( 42 )` binds `x` — but a ρ that is already *bound* is still counted, and that is the shape every real dispatch produces. On 0.0.119:

```
$ phino rewrite --normalize --flat '⟦ ρ ↦ Φ.y, x ↦ ∅ ⟧( Φ.z )'
⟦ ρ ↦ Φ.y, x ↦ ∅ ⟧( α0 ↦ Φ.z )      # no rule fires, argument stuck forever
$ phino rewrite --normalize --flat '⟦ ρ ↦ Φ.y, x ↦ ∅ ⟧( α1 ↦ Φ.z )'
⟦ ρ ↦ Φ.y, x ↦ Φ.z ⟧                # the bound ρ occupies index 0
```

This matters because ρ is bound by the `dot` rule *before* the positional argument is applied, so the atom the reporter of #1127 cared about still never saturates. With `plus.phi` holding

```
⟦ obj ↦ ⟦ plus ↦ ⟦ ρ ↦ ∅, x ↦ ∅, λ ⤍ L_number_plus ⟧ ⟧, φ ↦ Φ.obj.plus( 1 ), ρ ↦ ∅ ⟧
```

`phino dataize --partial --max-steps=100 plus.phi` runs away to the step limit: `x` stays void, so the atom is never saturated and `--partial` can never park it. Delete the atom's `ρ ↦ ∅` and the same file parks `⟦ x ↦ 1, λ ⤍ L_number_plus ⟧` at once. (This is a different cause from #1078, where the atom does fire and its *result* needs universe objects.)

`alpha` and `overa` already carry `not(eq(𝜏1, ρ))`, so ρ is never the target of a positional argument, whatever its position. The count has to agree with that, otherwise a dispatch-bound ρ silently shifts every real slot by one. Add `notAsset (BiTau AtRho _) = False` beside the existing void clause in `src/Rule.hs:128`, so ρ is excluded from `domain` whether it is bound or void.

VasilevNStas:
Fixes #1134.

## Problem

The #1127 fix (PR #1130) excluded only the **void** ρ (`BiVoid AtRho`) from the
`domain` count. But the `dot` rule binds ρ to a dispatch result
(`BiTau AtRho expr`) **before** the positional argument is processed, so a
dispatched atom still counted ρ as a positional slot:

```
$ phino rewrite --normalize --flat '⟦ ρ ↦ Φ.y, x ↦ ∅ ⟧( Φ.z )'
⟦ ρ ↦ Φ.y, x ↦ ∅ ⟧( α0 ↦ Φ.z )      # no rule fires, argument stuck forever
```

`alpha` and `overa` already guard against ρ as a target
(`not(eq(𝜏1, ρ))`), but the `domain` count must agree — otherwise a
dispatch-bound ρ silently shifts every real slot by one.

## Fix

One clause in `src/Rule.hs:129`:

```haskell
notAsset (BiTau AtRho _) = False
```

beside the existing `notAsset (BiVoid AtRho) = False` from #1130, so ρ is
excluded from `domain` whether it is bound or void.

## Verification

| Input | Output | Status |
|---|---|---|
| `⟦ ρ ↦ Φ.y, x ↦ ∅ ⟧( Φ.z )` | `⟦ ρ ↦ Φ.y, x ↦ Φ.z ⟧` | α0 → x ✓ |
| `⟦ ρ ↦ Φ.y, x ↦ ∅ ⟧( α1 ↦ Φ.z )` | `⊥` | α1 out of range ✓ |
| `⟦ ρ ↦ ∅, x ↦ ∅ ⟧( 42 )` | `⟦ ρ ↦ ∅, x ↦ 42 ⟧` | α0 → x (void ρ skipped) ✓ |
| `[[ a -> ?, ^ -> ?, x -> ? ]](~1 -> Q.y)` | `x -> Q.y` | reviewer case from #1130 ✓ |

- Rewriter: 110 tests, 0 failures
- Full suite: 2397 examples, 13 pre-existing (LocatorSpec, GHC 9.10.3)


0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2330](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1128, opened 2026-09-09
XMIR reader rejects `base="⊥"`, so most eo-runtime programs cannot be read (src/XMIR.hs:489)


yegor256:
The XMIR reader refuses any element whose `@base` is the terminator. In `xmirToExpression` (`src/XMIR.hs:489`) the `case base of` ladder handles `.`-prefixed dispatch, `ξ`, `Φ`, `Φ.…` and `ξ.…`, and everything else falls into the catch-all that throws `The @base attribute must be either ['∅'|'Φ'] or start with ['Φ.'|'ξ.'|'.']`. Since `⊥` matches none of those branches, reading stops at the first terminator node. With 0.0.118:

```
$ cat bot.xmir
<?xml version="1.0"?>
<object>
  <o name="foo">
    <o name="x" base="⊥"/>
  </o>
</object>
$ phino rewrite --input=xmir bot.xmir
[ERROR]: Couldn't traverse though given XMIR, cause: The @base attribute must be either ['∅'|'Φ'] or start with ['Φ.'|'ξ.'|'.']
XMIR:
<?xml version="1.0" encoding="UTF-8"?>
<o base="⊥" name="x"/>
```

The terminator is a first-class term everywhere else in phino: `ExTermination` is in `src/AST.hs:31`, the φ parser and printer accept `⊥`, and a dozen normalization rules (`stop`, `null`, `dd`, `dc`, `dca`, `miss`, `over`, `overa`, `dl`, `amiss`) produce it as their result. The `eo` compiler emits it too, so the refusal is not theoretical: in the `eo-runtime` build 32 of the 173 parsed XMIR files carry `base="⊥"` (79 nodes) and 8 of the 12 merged ones (71 nodes), among them `console`, `chunk`, `socket`, `file`, `recovered`, `tuple/slice` and most of `bytes`. None of those programs can be read by phino at all, which blocks objectionary/eo#8548, where the whole program has to be loaded as one universe.

Add a `"⊥" -> pure ExTermination` branch next to the existing `"Φ"` and `"ξ"` ones, rejecting children the same way those two do. The writer needs the mirror clause to round-trip: `expression` in `src/XMIR.hs:100..150` has no `ExTermination` case either, so it currently falls through to `UnsupportedExpression`, and one `expression ExTermination _ = pure ("⊥", [])` closes that side.

yegor256:
[...]

PR #1137, opened 2026-09-09
fix(xmir): accept applied terminator base="⊥" with children (#1135)

### LINKED ISSUE #1135: XMIR reader rejects an *applied* `base="⊥"`, though `dc`/`dca` define its reduction (src/XMIR.hs:488)
The fix for #1128 (1fea1b8, shipped in 0.0.119) reads a *bare* terminator but refuses an *applied* one. `src/XMIR.hs:488-491` copies the no-children guard of the `Φ` and `ξ` branches:

```haskell
"⊥" ->
  if null (cur C.$/ C.element (toName "o"))
    then pure ExTermination
    else throwIO (InvalidXMIRFormat "Application of '⊥' is illegal in XMIR" cur)
```

For `Φ` and `ξ` that guard is right, but for `⊥` it contradicts phino's own rules: `resources/normalize/dc.yaml` is `⊥(𝜏 ↦ 𝑒) → ⊥` and `resources/normalize/dca.yaml` is `⊥(α𝑖 ↦ 𝑒) → ⊥`, and they exist precisely to collapse an applied terminator. EO writes the terminator together with its panic reason, so in practice every occurrence has children. From `eo-runtime`'s `bytes/as-bool.xmir`:

```xml
<o as="α1" base="⊥" line="18" loc="Φ.bytes.as-bool.φ.α1.α1" pos="6">
   <o as="α0" base=".printf" line="19" loc="Φ.bytes.as-bool.φ.α1.α1.α0" pos="65">
```

The practical effect is that the reader is no better off than before the fix. Reading every document of the `eo-runtime` build with 0.0.119, all 32 of the 173 parsed documents that carry a terminator still fail, and 8 of the 12 merged ones — only 4 merged documents can be read at all. It is now the *only* remaining reader failure: the documents that used to trip the short-Δ crash of #1129 read fine.

Replace the guard with `xmirToApplication ExTermination (cur C.$/ C.element (toName "o")) fqn`, which is what the `Φ.`/`ξ.` branches already do through `xmirToExpression'`, and let `dc`/`dca` collapse the result. The writer side needs nothing: `expression ExTermination` at `src/XMIR.hs:100` returns the base `⊥`, and the generic `ExApplication` clause renders the children around it.

VasilevNStas:
Fixes #1135.

## Problem

The #1128 fix read a bare `⊥` but refused an *applied* one with children,
copying the no-children guard from `Φ` and `ξ`. For `⊥` the guard is
wrong:

- The rules `dc` (`⊥(τ→e) → ⊥`) and `dca` (`⊥(αᵢ→e) → ⊥`) define
  reduction of applied terminator
- EO writes `⊥` with its panic reason, so every real-world occurrence has
  children (79 nodes across 32 XMIR files in `eo-runtime`)

With 0.0.119 (the #1128 fix), all 32 XMIR documents containing a terminator
in `eo-runtime` still fail to read — the terminator is never bare.

## Fix

One-line change in `src/XMIR.hs:488`:

```diff
-"⊥" ->
-  if null (cur C.$/ C.element (toName "o"))
-    then pure ExTermination
-    else throwIO (...)
+"⊥" -> xmirToApplication ExTermination (cur C.$/ C.element (toName "o")) fqn
```

Children become arguments that `dc`/`dca` can later collapse.
The writer side needs nothing — `expression ExTermination` already
returns `("⊥", [])`, and the generic `ExApplication` clause
renders the children around it.

## Verification

Round-trip:
```
<o name="x" base="⊥"><o base="Φ.number"/></o>
→ ⟦ x ↦ ⊥( Φ.number ) ⟧
```

- XMIR: 98 examples, 0 failures
- Full suite: 2398 examples, 13 pre-existing (LocatorSpec, GHC 9.10.3)

### DIFF
--- a/src/XMIR.hs
+++ b/src/XMIR.hs
@@ -485,10 +485,7 @@ xmirToExpression cur fqn
           if null (cur C.$/ C.element (toName "o"))
             then pure ExRoot
             else throwIO (InvalidXMIRFormat "Application of 'Φ' is illegal in XMIR" cur)
-        "⊥" ->
-          if null (cur C.$/ C.element (toName "o"))
-            then pure ExTermination
-            else throwIO (InvalidXMIRFormat "Application of '⊥' is illegal in XMIR" cur)
+        "⊥" -> xmirToApplication ExTermination (cur C.$/ C.element (toName "o")) fqn
         'Φ' : '.' : rest -> xmirToExpression' ExRoot "Φ" rest cur fqn
         'ξ' : '.' : rest -> xmirToExpression' ExXi "ξ" rest cur fqn
         _ -> throwIO (InvalidXMIRFormat "The @base attribute must be either ['∅'|'Φ'] or start with ['Φ.'|'ξ.'|'.']" cur)
--- a/test-resources/xmir-parsing-packs/termination-with-application.yaml
+++ b/test-resources/xmir-parsing-packs/termination-with-application.yaml
@@ -1,13 +1,12 @@
 # SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
 # SPDX-License-Identifier: MIT
 ---
-failure: true
 xmir: |
   <object>
     <o name="app">
-      <o base="⊥">
+      <o name="x" base="⊥">
         <o base="Φ.number"/>
       </o>
     </o>
   </object>
-phi: ''
+phi: '[[ app -> [[ x -> T(Q.number) ]] ]]'


0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (10) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2342](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1140, opened 2026-09-09
fix(print): keep a non-UTF-8 string datum in byte form instead of aborting (#1138)

### LINKED ISSUE #1138: Printing a string datum that is not valid UTF-8 aborts the run (src/Bytes.hs:330)
Printing a `Φ.string` whose Δ datum is not valid UTF-8 aborts the run. `btsToUnescapedStr` (`src/Bytes.hs:330`) is

```haskell
btsToUnescapedStr bytes = T.unpack (T.decodeUtf8 (B.pack (btsToWord8 bytes)))
```

and `T.decodeUtf8` is the throwing variant, so any invalid sequence raises `Cannot decode byte '\xfe': Data.Text.Encoding: Invalid UTF-8 stream`. The print path reaches it through `btsToStr`, at `src/CST.hs:356` and `src/CST.hs:429` for the φ literal and at `src/XMIR.hs:137` for the comment beside a string datum. On 0.0.120, with and without `--sweet`:

```
$ cat str.phi
⟦ a ↦ Φ.string( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ FE- ⟧ ) ) ⟧
$ phino rewrite str.phi
[ERROR]: Cannot decode byte '\xfe': Data.Text.Encoding: Invalid UTF-8 stream
```

An arbitrary byte array is a legal datum, and `Φ.string` carries whatever bytes it is given: nothing in the grammar promises they decode. Real EO relies on that, because `eo-runtime` tests malformed input deliberately, with data such as `F0-90-80-41`, a lone `FE-` and a bare `A0-`. Reading the build's documents with 0.0.120, this is now the only remaining failure: `string/slice.xmir` and `string/length.xmir` cannot be read, nor the merged `string.xmir`, which is 1 of the 12 merged documents and blocks loading the whole program as one universe for objectionary/eo#8548. It is the string-side sibling of #1129, which guarded the number side of the very same printer.

Mirror that fix: test the bytes with the non-throwing `T.decodeUtf8'` and, when it fails, keep the datum in its byte form `Φ.string(Φ.bytes(⟦ Δ ⤍ … ⟧))` instead of printing a literal — a `sweetString` guard next to `sweetNumber` (`src/CST.hs:272`), reported through `sweetCollapsible`, plus the same check before the comment at `src/XMIR.hs:137`. That keeps every byte, exactly as a payload NaN is kept today. `argToString` (`src/Functions.hs:63`) and the regex target in `src/Rule.hs:227` decode unguarded too, so a condition on such bytes still throws.

VasilevNStas:
Fixes #1138.

## Problem

Printing a `Φ.string` whose Δ datum is not valid UTF-8 aborted the run with an uncaught `decodeUtf8` exception:

```
$ phino rewrite <<< '⟦ a ↦ Φ.string( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ FE- ⟧ ) ) ⟧'
[ERROR]: Cannot decode byte '\xfe': Data.Text.Encoding: Invalid UTF-8 stream
```

`btsToUnescapedStr` (src/Bytes.hs:330) uses the throwing `T.decodeUtf8`, and the print path reached it unguarded at both `EX_STRING` sites in CST (src/CST.hs:356, and the `applicationToPrimitive` collapse) and at the comment in the XMIR writer (src/XMIR.hs:137). An arbitrary byte array is a legal datum, and eo-runtime tests malformed input deliberately. This is the string-side sibling of #1129.

## Fix

| File | Change |
|------|--------|
| `src/Bytes.hs` | New `btsIsUtf8` predicate using the non-throwing `T.decodeUtf8'` |
| `src/CST.hs` | `sweetString = btsIsUtf8` next to `sweetNumber`; guard on the `EX_STRING` clause and on `sweetCollapsible (DataString ...)` so malformed datums fall through to the generic application clause (byte form) |
| `src/XMIR.hs` | Skip the string comment when the datum is not valid UTF-8 |

## Verification

```
$ phino rewrite --sweet <<< '⟦ a ↦ Φ.string( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ FE- ⟧ ) ) ⟧'
⟦ a ↦ Φ.string( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ FE- ⟧ ) ) ⟧   # byte form kept, no crash

$ phino rewrite --sweet <<< '⟦ a ↦ Φ.string( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ F0-90-80-41 ⟧ ) ) ⟧'
⟦ a ↦ Φ.string( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ F0-90-80-41 ⟧ ) ) ⟧

$ phino rewrite --sweet <<< '⟦ a ↦ Φ.string( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ 77-6F-72-6C-64 ⟧ ) ) ⟧'
⟦ a ↦ "world" ⟧   # valid UTF-8 still collapses to the literal
```

- Full suite: 2408 examples, 13 pre-existing (LocatorSpec, GHC 9.10.3)

### DIFF
--- a/src/Bytes.hs
+++ b/src/Bytes.hs
@@ -15,6 +15,7 @@ module Bytes
   , unescapeStr
   , btsToNum
   , btsToUnescapedStr
+  , btsIsUtf8
   , btsAnd
   , btsOr
   , btsNot
@@ -328,6 +329,22 @@ unescapeStr = go
 -- "5"
 btsToUnescapedStr :: Bytes -> String
 btsToUnescapedStr bytes = T.unpack (T.decodeUtf8 (B.pack (btsToWord8 bytes)))
+
+-- Whether the byte array is valid UTF-8. The string-side counterpart of the
+-- eight-byte check on numbers: a short or malformed datum is legal, and the
+-- printer keeps it in its byte form instead of aborting with an uncaught
+-- 'decodeUtf8' exception (see #1138).
+-- >>> btsIsUtf8 (BtMany ["77", "6F", "72", "6C", "64"])
+-- True
+-- >>> btsIsUtf8 (BtMany ["F0", "90", "80", "41"])
+-- False
+-- >>> btsIsUtf8 (BtOne "FE")
+-- False
+btsIsUtf8 :: Bytes -> Bool
+btsIsUtf8 bytes =
+  case T.decodeUtf8' (B.pack (btsToWord8 bytes)) of
+    Left _ -> False
+    Right _ -> True
 
 -- Bitwise conjunction of two byte arrays, byte by byte. EO's 'BytesRaw.and'
 -- refuses operands of different lengths, so there is nothing to yield for them
--- a/src/CST.hs
+++ b/src/CST.hs
@@ -11,7 +11,7 @@
 module CST where
 
 import AST
-import Bytes (NonFinite, btsSize, btsToNonFinite, btsToNum, btsToStr)
+import Bytes (NonFinite, btsIsUtf8, btsSize, btsToNonFinite, btsToNum, btsToStr)
 import Data.Maybe (isJust)
 import qualified Data.Text as T
 import qualified Yaml as Y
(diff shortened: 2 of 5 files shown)


0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2358](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1144, opened 2026-09-09
ci: round-trip objectionary/home through phino on every PR (#1139)

### LINKED ISSUE #1139: Add an `objectionary.yml` workflow that round-trips all of objectionary/home through phino
Every XMIR reader and printer bug of the last two days was found the same way: by pointing phino at the XMIR of a real EO build and reading the error. #1127, #1128, #1129, #1134, #1135 and #1138 all came out of that one exercise, by hand. CI could do it on every pull request, against the whole standard library, and the repository has no workflow that reads real XMIR today.

The proposal is a new `.github/workflows/objectionary.yml` that takes the `objects/` directory of [objectionary/home](https://github.com/objectionary/home), pushes it through phino, and then runs a program compiled from what phino printed:

1. `cabal build all`, then `phino_dir=$(dirname "$(cabal list-bin phino)")`, the way `regression-check.yml` already does it.
2. Check out `objectionary/home`, install `eoc` (`npm i -g eolang`) and Java 21.
3. `eoc -s objects parse` — the parsed documents land in `.eoc/1-parse/**/*.xmir`.
4. For each of them, `phino rewrite --input=xmir --output=xmir -t <out> <in>`. This step alone is the reader and printer test, and it fails loudly on exactly the defects listed above.
5. `eoc print` with `-Deo.printSourcesDir=<the rewritten dir>` and `-Deo.printOutputDir=objects`, to regenerate the `.eo` sources from phino's XMIR.
6. `eoc -s objects dataize <some object>`, and compare the result with the same dataization run before step 4.

Step 5 is not optional: `eoc compile` and `eoc dataize` re-parse the `.eo` files and overwrite the parsed XMIR, so without regenerating the sources the program that runs is not the one phino printed. I measured it locally on eoc 0.37.0 — after `eoc compile` the document's `author` attribute was back to `eo-parser`.

Most of the wiring already exists: `eoc normalize` (in `eolang/src/commands/normalize.js`) is exactly parse → `phino rewrite --input=xmir --output=xmir --normalize` per file → `eo:print` into the sources directory. The workflow wants the same three steps without `--normalize`, so that the round trip is the identity and any difference is a bug. One pitfall to plan for: with eoc 0.37.0 the objects pulled into `.eoc/2-pull/` failed the parser it pins (`name must start with a lowercase letter`), so the workflow should pin the `eo` version and the `home` revision together rather than tracking both tips.

Worth adding as a scale check: reading the documents of an `eo-runtime` build with 0.0.120, 171 of 173 parsed documents round-trip and 2 fail on #1138; on 0.0.118 it was 141 and 32.

VasilevNStas:
Fixes #1139.

## What

New `.github/workflows/objectionary.yml` that pushes every XMIR document of [objectionary/home](https://github.com/objectionary/home) through phino and checks the round trip is lossless.

## Pipeline

1. `cabal build all`; `phino_dir=$(dirname "$(cabal list-bin phino)")` (like `regression-check.yml`)
2. Check out `objectionary/home` at a pinned revision; install `eoc` (`npm i -g eolang@`) + Java 21
3. `eoc -s objects parse` — documents land in `.eoc/1-parse/**/*.xmir`
4. Each of them through `phino rewrite --input=xmir --output=xmir` into `.eoc/2-rewritten` — this step alone is the reader + printer test; it fails loudly on the exact defects of #1127/#1128/#1129/#1134/#1135/#1138
5. `eoc print --print-input=2-rewritten --print-output=../objects` regenerates the `.eo` sources from phino's XMIR

## Pinning

Found by local experiments on a clean Maven repo: with the parser eoc 0.37.1 pins (0.63.1), neither the home HEAD sources nor the eo-runtime objects pulled by `eo:assemble` parse — their voids use the new `[^ ...]` syntax ('parameter names in voids must be NAME or @'). The pair **home tag 0.62.1 + eo parser 0.62.1** parses all documents with zero errors. `HOME_REV` and `EO_PARSER` are pinned to that pair, per the report's advice to pin eo version and home revision together.

## Dropped: dataize comparison

The original plan compared a `dataize` run before/after the round trip. This is not possible with any parser eoc 0.37.1 provides: `eo:assemble` always pulls eo-runtime objects written in the new syntax, so the comparison cannot run. The remaining pipeline is precisely the reader/printer identity test the report centers on; the dataize comparison can be added back once eoc pins a parser that reads the eo-runtime 0.63.+ sources.

## Notes

- Sources round-tripped: 161 documents at the pinned 0.62.1 pair.
- The workflow would have caught every reader/printer fix since #1127.


VasilevNStas:
@yegor256 
please review this
I succeeded in doing this.

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2432](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1145, opened 2026-09-09
ci(release): remove the image ghcup on ubuntu-24.04 so setup wins (#1141)

### LINKED ISSUE #1141: `release-binary.yml` publishes no `ubuntu-24.04` binary since 0.0.116, so downloads return 403
The `ubuntu-24.04` leg of the matrix in `.github/workflows/release-binary.yml:17` has failed on four consecutive tags — 0.0.117, 0.0.118, 0.0.119 and 0.0.120 — while the other four legs (`macos-15`, `macos-14-large`, `ubuntu-22.04`, `windows-2022`) succeeded each time. Because the upload at line 54 never runs on a failed leg, the last `ubuntu-24.04` binary in the bucket is 0.0.116.

That breaks every downstream consumer of the published URL. `hone-maven-plugin` fetches `http://phino.objectionary.com/releases/ubuntu-24.04/phino-<version>` in CI, and since the bucket denies listing, a missing object comes back as HTTP 403 rather than 404 — so a version bump past 0.0.116 fails there with a bare `curl: (22)` and no hint that the cause is an unpublished artifact. The same four tags are fine on `ubuntu-22.04`, which makes the gap look like a per-runner problem rather than a build break.

The failing step is `haskell-actions/setup@v2.12.0` at line 24 resolving `ghc-version: 9.6.7`. Its log shows two ghcup installations competing — the runner-preinstalled `/usr/local/.ghcup` against `/home/runner/.ghcup` — with repeated `The version '9.6.7' of the tool ghc is not installed` errors and GHCup-00330 `has shadowed binaries` warnings naming exactly that pair. Pinning the leg to the ghcup already on the `ubuntu-24.04` image, or setting `ghcup-release-channel`/`PATH` so only one ghcup wins, should let the leg build and upload again.

VasilevNStas:
Fixes #1141.

## Problem

The `ubuntu-24.04` leg of the `release-binary.yml` matrix failed on four consecutive tags (0.0.117–0.0.120) while the other four legs succeeded. Since the upload step never runs on a failed leg, the bucket holds no `ubuntu-24.04` binary after 0.0.116, and downstream consumers (e.g. `hone-maven-plugin`) get a bare `curl: (22)` 403.

Root cause: the ubuntu-24.04 runner image ships its own ghcup in `/usr/local/.ghcup`, which stays ahead of the `/home/runner/.ghcup` installed by `haskell-actions/setup@v2.12.0` on PATH. `ghc`/`cabal` then resolve to the stale `/usr/local` binaries that lack the requested versions (`The version '9.6.7' of the tool ghc is not installed`, GHCup-00330 `has shadowed binaries` naming exactly that pair).

## Fix

`haskell-actions/setup` (v2.12.0) has no input to prefer or disable the image's ghcup, so the leg removes the pre-installed one before setup runs — only the `/home/runner/.ghcup` remains on PATH, which is exactly the setup the other legs already have:

```yaml
- name: Remove pre-installed ghcup (ubuntu-24.04)
  if: matrix.os == 'ubuntu-24.04'
  run: sudo rm -rf /usr/local/.ghcup
```

Only the failing leg is touched; `macos-15`, `macos-14-large`, `ubuntu-22.04` and `windows-2022` are unchanged.

### DIFF
--- a/.github/workflows/release-binary.yml
+++ b/.github/workflows/release-binary.yml
@@ -20,6 +20,18 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v7
+      - name: Remove pre-installed ghcup (ubuntu-24.04)
+        # The ubuntu-24.04 image ships its own ghcup in /usr/local/.ghcup.
+        # It stays ahead of the /home/runner/.ghcup that haskell-actions/setup
+        # installs, so `ghc` and `cabal` resolve to stale binaries that do not
+        # carry the requested versions ('The version ... is not installed'),
+        # and the leg fails while the other four legs pass. The image ghcup is
+        # dead anyway once setup prepends its own bin to PATH; only one ghcup
+        # may win, see https://github.com/objectionary/phino/issues/1141
+        if: matrix.os == 'ubuntu-24.04'
+        shell: bash
+        run: |
+          sudo rm -rf /usr/local/.ghcup
       - name: Set up GHC and Cabal
         uses: haskell-actions/setup@v2.12.0
         with:


VasilevNStas:
@yegor256 
Please take a look at this pull request and do a review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2444](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1146, opened 2026-09-09
fix(print): name the datum payload φ so a dispatched literal binds (#1142)

### LINKED ISSUE #1142: Rewriting a data object names the bytes payload `data`, which binds no void, so literals reduce to ⊥ (src/AST.hs:264)
A no-op `phino rewrite` turns a reducible program into `⊥` whenever the universe defines `bytes`. The culprit is `dataBytes` at `src/AST.hs:264`, which renders the payload of a data object as `ArTau (AtLabel "data")`, i.e. `Φ.bytes( data ↦ ⟦ Δ ⤍ …, ρ ↦ ∅ ⟧ )`. `matchDataObject` reads back both that name and the positional `α0` (`src/AST.hs:220-221`), but only the positional form can actually bind: eo-runtime declares `[@] > bytes`, so `bytes` has exactly one void and it is named `φ`. `Φ.bytes( α0 ↦ … )` binds it through the `alpha` rule; `Φ.bytes( data ↦ … )` binds nothing, and every dispatch on the literal falls to `⊥`.

Put this in `a.phi`:

```
⟦
  bytes ↦ ⟦ φ ↦ ∅, size ↦ ⟦ λ ⤍ L_bytes_size ⟧ ⟧,
  number ↦ ⟦ as-bytes ↦ ∅, φ ↦ ξ.as-bytes ⟧,
  x ↦ Φ.number( α0 ↦ Φ.bytes( α0 ↦ ⟦ Δ ⤍ 40-10-00-00-00-00-00-00, ρ ↦ ∅ ⟧ ) )
⟧
```

`phino morph --partial --flat --hide-rho --inside 'Φ.x.as-bytes' a.phi` prints `⟦ φ ↦ ⟦ Δ ⤍ 40-10-00-00-00-00-00-00 ⟧, size ↦ ⟦ λ ⤍ L_bytes_size ⟧ ⟧`. Run `phino rewrite a.phi > b.phi` and the very same morph on `b.phi` prints `⊥`. The XMIR path does the same, which is what a user actually hits, because objectionary/eo emits positional arguments: `<o name="v" base="Φ.number"><o as="α0" base="Φ.bytes"><o as="α0">40-10-00-00-00-00-00-00</o></o></o>` through `phino rewrite --input=xmir --output=xmir` comes back as `<o as="as-bytes" base="Φ.bytes"><o as="data">…`. The outer rename is right, since `number`'s void really is `as-bytes`; only the inner one names something that does not exist.

This is invisible while `bytes` is left to the built-in atoms, and fatal as soon as the universe carries the real object, which is exactly what `phino merge --input=xmir` over eo-runtime's XMIR produces: there every number, string and bool literal is dead on its first dispatch, and it blocks whole-program symbolic evaluation for objectionary/eo#8548.

I am not asking to undo #964 — the named style should stay. The smallest fix is to render the payload with `AtPhi` in `dataBytes` (`src/AST.hs:264`) so the printed name is the one `bytes` declares, and to add `AtPhi` to `dataArg` (`src/AST.hs:220`) so the sugar still reads back. Keeping `data` accepted on input preserves jeo compatibility.

VasilevNStas:
Fixes #1142.

## Problem

A data object is printed with its payload bound to the attribute `data`:

```
Φ.number( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ … ⟧ ) )
```

But the real EO `bytes` is declared `[@] > bytes` — its only void is named `φ`. `matchDataObject` reads back `data`, `φ` and the positional `α0`, yet only a name the object actually declares can bind: `Φ.bytes( α0 ↦ … )` hits `φ` through `alpha`, while `Φ.bytes( data ↦ … )` hits **nothing** — so every dispatch on a literal falls through to `⊥`. In a universe merged out of eo-runtime (where `number`/`string`/`bytes` are the real objects), every literal is dead on its first dispatch.

## Fix

Print the payload as the void the object declares — the author's suggested change, plus the two mirror sites so every notation agrees:

| File | Change |
|------|--------|
| `src/AST.hs` | `dataBytes` renders the payload with `ArTau AtPhi`; `dataArg` accepts `AtPhi` alongside `data`/`α0` (legacy `data` input still parses → jeo compatibility) |
| `src/XMIR.hs` | writer emits the payload child with `as="φ"` (reader already maps `φ → AtPhi`) |
| `src/Sugar.hs` | `saltifyPrimitive` names the payload `φ` in salty output, in lockstep with `dataBytes` |
| `test-resources/atoms/primitives.js` | recognizes φ datum shapes (keeps legacy `data`), answers with `φ` |

Fixture universes that modeled `bytes` as an echo (`bytes(data) -> [[ @ -> $.data ]]`) are rewritten to declare the void as `φ` with no `@`-answer: with a φ-bound payload, dataization reaches Δ directly through `delta`, so the echo dispatch is unnecessary — the reduction-label sequence for `5.plus(6)` shortens accordingly, which is the point of the fix.

## Verification

Issue repro on the built binary — morph of `Φ.x.as-bytes` after a **no-op rewrite** (was `⊥`):

```
$ phino morph --partial --flat --hide-rho --inside 'Φ.x.as-bytes' b.phi
⟦ φ ↦ ⟦ Δ ⤍ 40-10-00-00-00-00-00-00 ⟧, size ↦ ⟦ λ ⤍ L_bytes_size ⟧ ⟧   # identical to the source
```

XMIR round-trip: `<o as="α0" base="Φ.bytes"><o as="α0">40-…</o></o>` now comes back with `<o as="φ">`.

- Full suite: **2409 examples, 13 failures — all pre-existing Locator on local GHC 9.10.3**, 0 introduced by this change
- New `PrinterSpec` case pins the invariant: salty output names the payload `φ` and never `data`
- `hlint` and `fourmolu --mode check` clean, `yamllint` clean


VasilevNStas:
Heads-up for anyone running `make test` locally: the suite ends at **13 failures, all `LocatorSpec`** — pre-existing on the local GHC 9.10.3 (`HasCallStack` backtrace leaking into error messages), unrelated to this change and already being fixed in #1100. CI (GHC 9.6.7 / 9.8.4) is green, and `cabal test` on this branch introduces **0** new failures (verified: filtering out `Locator` leaves none).

VasilevNStas:
@yegor256 
plz review this pull request

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (168 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2456](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1147, opened 2026-09-09
fix(dataize): walk --deep innermost-first so nested atoms fire (#1143)

### LINKED ISSUE #1143: `morph --deep` fires an atom before reducing its arguments, so a nested atom is dropped unfired (src/Dataize.hs:319)
`morph --deep` fires an atom before it walks that atom's arguments, so a λ standing inside an argument never fires and its computation is dropped. `deepened` (`src/Dataize.hs:308`) walks a term with `go`, which asks `fired` about the whole term first and descends into `parts` only when nothing fired: `answer <- fired (contextualize term context) univ state' ctx'` on `src/Dataize.hs:319`, then `maybe (parts context term state' ctx') pure answer` on `src/Dataize.hs:320`. The walk is outermost-first, and the answer of the outer firing replaces the entire sub-tree, arguments included.

Two things go wrong at once: the inner atom is discarded unfired, and the outer atom was handed an argument nobody had reduced. Put this in `order.phi`:

```
⟦ a ↦ ⟦ p ↦ ⟦ x ↦ ∅, λ ⤍ L_p ⟧, ρ ↦ ∅ ⟧, probe ↦ ⟦ φ ↦ Φ.a.p( α0 ↦ Φ.a.p( α0 ↦ Φ.a ) ), ρ ↦ ∅ ⟧ ⟧
```

and register one served script for `L_p` that logs the `𝑏` it is given and answers `Φ.a`. Then `phino morph --deep --partial --atoms=order.json --flat --hide-rho --inside 'Φ.probe' order.phi` prints `⟦ φ ↦ Φ.a ⟧` and the log holds exactly one line, the outer call, with `𝑏` equal to `⟦ x ↦ Φ.a.p( α0 ↦ Φ.a ), ρ ↦ … ⟧`. Two atoms are written and one of them fires.

An atom cannot repair this from its own side. The protocol gives it `𝑏` and the universe `𝑒`, so its only recourse is to re-invoke phino on `𝑏.x`, and on real objects `𝑏` is enormous: morphing `(a.minus b).times (a.minus b)` against a universe merged from `eo-runtime` hands `L_number_plus` a body of 428 KB, since every operand carries the whole `number` object inline. Handing that back through `--inside` fails outright with `spawnSync phino E2BIG`, and splicing it through stdin instead takes about two minutes for that one small fragment, because every atom re-runs the reduction its caller had already done. This is what blocks Step 4 of objectionary/eo#8548.

Suggested fix: make the walk innermost-first — in `go` (`src/Dataize.hs:316-320`) descend into `parts` first and ask `fired` about the rebuilt term afterwards, so an atom's arguments are already reduced when its λ fires and no nested λ is thrown away.

VasilevNStas:
Fixes #1143. Follow-up to #1124.

## Problem

`go` in `deepened` (`src/Dataize.hs:316`) asked `fired` about the whole term **first** and only descended into `parts` when nothing fired. An atom's answer replaces the entire sub-tree, so a λ standing inside that atom's **argument** is dropped unfired, and the outer atom receives an argument nobody reduced.

An atom can't repair this from its side: given `𝑏`/`𝑒` it would have to re-invoke phino on `𝑏.x`, but on real operands `𝑏` is enormous (an operand carries the whole `number` object inline) — `--inside` fails with `spawnSync phino E2BIG`, and stdin splicing re-runs the caller's reduction (~2 min for one small fragment). This blocks Step 4 of objectionary/eo#8548.

## Fix

Make the walk **innermost-first** — descend into `parts`, then ask `fired` about the rebuilt term:

```haskell
go context term state' caller = do
  ctx' <- deeper caller
  (walked, walkedState) <- parts context term state' caller
  answer <- fired (contextualize walked context) univ walkedState ctx'
  maybe (pure (walked, walkedState)) pure answer
```

So an atom's arguments are already reduced when its λ fires, and no nested λ is thrown away.

## Reproduction (before → after)

With one served `L_p` logging the `𝑏` it gets and answering `Φ.a`, on the issue's `order.phi`, `morph --deep --inside 'Φ.probe'`:

- **before**: outer `L_p` fired with `𝑏 = ⟦ x ↦ Φ.a.p( α0 ↦ Φ.a ), …⟧` — the inner atom still unreduced, and only it fired.
- **after**: result `⟦ φ ↦ Φ.a ⟧`, **two** `L_p` calls, and the outer's `𝑏 = ⟦ x ↦ Φ.a, …⟧` — the inner atom fired first and reduced the argument.

## Test

A `morph --deep` case `5.plus( 6.plus( 7 ) )` → 18 (`40-32`): the inner `plus` must fire before the outer, else the outer gets an unevaluated operand and the run collapses to `⊥`. Verified this case is **red without the change and green with it**.

## Notes / scope

- Full suite: **2409 examples, 13 failures — all pre-existing `LocatorSpec` on local GHC 9.10.3** (the `HasCallStack` leak, fixed in #1100). CI (9.6.7/9.8.4) is green; 0 new failures here (filtering out `Locator` leaves none).
- `go` fires once per visited node on the rebuilt term (no extra fix-point re-walk of `answer`), matching the original `fired` contract; the issue's suggested minimal change.

### DIFF
--- a/src/Dataize.hs
+++ b/src/Dataize.hs
@@ -316,8 +316,9 @@ deepened expr univ = go ExXi expr
     go :: Expression -> Expression -> State -> DataizeContext -> IO (Expression, State)
     go context term state' caller = do
       ctx' <- deeper caller
-      answer <- fired (contextualize term context) univ state' ctx'
-      maybe (parts context term state' ctx') pure answer
+      (walked, walkedState) <- parts context term state' caller
+      answer <- fired (contextualize walked context) univ walkedState ctx'
+      maybe (pure (walked, walkedState)) pure answer
     -- The parts of a term nothing fired on, walked one by one and put back
     -- where they were, so the term keeps the shape it was written in.
     parts :: Expression -> Expression -> State -> DataizeContext -> IO (Expression, State)
--- a/test/DataizeSpec.hs
+++ b/test/DataizeSpec.hs
@@ -226,6 +226,12 @@ spec = do
         , "[[ x -> Q.number( as-bytes -> Q.bytes( data -> [[ D> 40-26-00-00-00-00-00-00 ]] ) ) ]]"
         )
       ,
+        ( "fires the atom nested in the argument of the atom it fires"
+        , "Q.@"
+        , primitives "[[ x -> 5.plus( 6.plus( 7 ) ) ]]"
+        , "[[ x -> Q.number( as-bytes -> Q.bytes( data -> [[ D> 40-32-00-00-00-00-00-00 ]] ) ) ]]"
+        )
+      ,
         ( "keeps the answer of the last atom fired along one chain of them"
         , "Q.@"
         , primitives "[[ x -> 5.plus( 6 ).plus( 7 ) ]]"


VasilevNStas:
@yegor256 
plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (11) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2452](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1156, opened 2026-09-10
Bind a literal's datum to φ, the void number and string declare

### LINKED ISSUE #1155: Rewriting a data object binds its bytes to `as-bytes`, which is no longer a void of `number`/`string`, so literals reduce to ⊥ (src/AST.hs:257)
`DataObject` at src/AST.hs:257 binds a rewritten literal's bytes to `as-bytes`: `ExApplication (BaseObject label) (ArTau (AtLabel "as-bytes") (dataBytes bts))`. The real `number` and `string` of eo-runtime no longer have a void by that name, so every literal phino prints reduces to ⊥ against the real library. This is objectionary/phino#1142 one level up — that one fixed the inner `bytes` payload, this is the outer binding, and `dataBytes` right below (src/AST.hs:265-269) already binds `AtPhi` with a comment citing #1142.

objectionary/eo commit eb650cea07 (#8620, "let number and string take @ as their void, like bytes") declares them `[@] > number` and `[@] > string`, so `φ` is their only void and a binding named `as-bytes` binds nothing. `as-bytes` still exists on both, but as a read, not as the void. The XMIR writer has the same split at src/XMIR.hs:116 and src/XMIR.hs:130 — the outer element is `[("as", "as-bytes"), ("base", "Φ.bytes")]` while the inner one is already `("as", "φ")` — and so does the sweet printer at src/Sugar.hs:210.

On 0.0.122, `phino rewrite --flat` on `⟦ x ↦ ⟦ φ ↦ 5, ρ ↦ ∅ ⟧, ρ ↦ ∅ ⟧` prints `Φ.number( as-bytes ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ 40-14-00-00-00-00-00-00, ρ ↦ ∅ ⟧ ) )`, and `--output=xmir` writes `as="as-bytes"` outside, `as="φ"` inside. Against a `number` declared the way eo master declares it, that output does not reduce:

```
⟦
  bytes ↦ ⟦ φ ↦ ∅, ρ ↦ ∅ ⟧,
  number ↦ ⟦ φ ↦ ∅, as-bytes ↦ ξ.φ, ρ ↦ ∅ ⟧,
  old ↦ ⟦ φ ↦ Φ.number( as-bytes ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ 40-14-00-00-00-00-00-00, ρ ↦ ∅ ⟧ ) ).as-bytes, ρ ↦ ∅ ⟧,
  new ↦ ⟦ φ ↦ Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ 40-14-00-00-00-00-00-00, ρ ↦ ∅ ⟧ ) ).as-bytes, ρ ↦ ∅ ⟧
⟧
```

`phino dataize --partial --inside 'Φ.old'` answers `dataization reached the terminator ⊥`; `--inside 'Φ.new'` answers `40-14-00-00-00-00-00-00`.

It blocks objectionary/eo#8548, which merges a whole eo-runtime build into one universe: every literal in the standard library dies on its first dispatch, exactly as before #1142.

Suggested fix: bind `ArTau AtPhi` at src/AST.hs:257, and write `as="φ"` at src/XMIR.hs:116 and src/XMIR.hs:130 and `AT_PHI` at src/Sugar.hs:210. Leave `asBytesArg` (src/AST.hs:216) accepting `as-bytes` and `α0` on input, the way `dataArg` already accepts `φ`, `data` and `α0`, so older XMIR keeps sugaring back.

yegor256:
A data object was printed as `Φ.number( as-bytes ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ … ⟧ ) )`. Now it prints `Φ.number( φ ↦ … )`, and `--output=xmir` writes `as="φ"` on the outer element too. The matcher still reads the old `as-bytes` and `α0` spellings, so XMIR from older jeo keeps sugaring back.

Since eo commit eb650cea07, `number` and `string` are declared `[@] > number` and `[@] > string`. φ is their only void; `as-bytes` survives as a read, not as a void. So a binding named `as-bytes` bound nothing, and every literal phino printed reduced to ⊥ against the real library. This is #1142 one level up — that one fixed the inner bytes payload, this one the outer binding.

The fixture universes that declared `number` with an `as-bytes` void now declare φ the way eo master does. That shortens the reduction-label sequence for `5.plus(6)` by the dispatch the void made necessary (`dot`/`ma`/`stay` becomes `ma`/`copy`), which is the point of the change. The README object models were stale in the same way after #1142, so they are updated too, and every example in them was run against the built binary.

Closes #1155

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,7 +269,7 @@ readline.createInterface({ input: process.stdin }).on('line', (line) => {
   const sum = hex(number(`${b}.ρ`) + number(`${b}.x`));
   process.stdout.write(`${JSON.stringify({
     id: message.id,
-    '𝑛': `Φ.number( as-bytes ↦ Φ.bytes( data ↦ ⟦ Δ ⤍ ${sum} ⟧ ) )`,
+    '𝑛': `Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ ${sum} ⟧ ) )`,
   })}\n`);
 });
 ```
@@ -291,8 +291,8 @@ to, and the expression it returned:
 ```bash
 $ cat sum.phi
 ⟦
-  bytes(data) ↦ ⟦ φ ↦ data ⟧,
-  number(as-bytes) ↦ ⟦ φ ↦ as-bytes, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧,
+  bytes ↦ ⟦ φ ↦ ∅ ⟧,
+  number ↦ ⟦ φ ↦ ∅, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧,
   φ ↦ 5.plus( 6 )
 ⟧
 $ phino dataize --atoms=atoms.json --evaluations=atoms.tsv --quiet \
@@ -319,9 +319,9 @@ printed in place of the bytes, and the run ends successfully:
 ```bash
 $ cat partial.phi
 ⟦
-  bytes(data) ↦ ⟦ φ ↦ data ⟧,
-  number(as-bytes) ↦ ⟦
-    φ ↦ as-bytes,
+  bytes ↦ ⟦ φ ↦ ∅ ⟧,
+  number ↦ ⟦
+    φ ↦ ∅,
     plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧,
     times(x) ↦ ⟦ λ ⤍ L_number_times ⟧,
     as-bool ↦ ⟦ λ ⤍ L_number_as_bool ⟧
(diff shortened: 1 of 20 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (228 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2359](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1157, opened 2026-09-10
fix(ast): bind a data object to φ, the only void number and string declare (#1155)

### LINKED ISSUE #1155: Rewriting a data object binds its bytes to `as-bytes`, which is no longer a void of `number`/`string`, so literals reduce to ⊥ (src/AST.hs:257)

VasilevNStas:
Fixes #1155.

## Problem
A data object is printed as `Φ.number( as-bytes ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ … ⟧ ) )`. #1142 fixed the inner payload name, but the outer binding stayed `as-bytes`. eo master (objectionary/eo#8620, commit eb650cea07) declares `[@] > number` and `[@] > string`: φ is their only void, and `as-bytes` is only a read. So every literal phino prints reduces to ⊥ against the real library — this blocks objectionary/eo#8548. Reproduced on 0.0.122 as described in the issue: `--inside Φ.old` answers the terminator, `--inside Φ.new` answers the bytes.

## Fix
Name the outer binding the void the object declares, one level up from #1142:
- `DataObject` (`src/AST.hs`) constructs `ArTau AtPhi`;
- `asBytesArg` (`src/AST.hs`) accepts `φ` alongside the legacy `as-bytes` and `α0` clauses, so older XMIR keeps sugaring back (compat preserved, as suggested in the issue);
- the XMIR writer (`src/XMIR.hs`) emits the outer element with `as="φ"` (reader already maps φ → AtPhi);
- `saltifyPrimitive` (`src/Sugar.hs`) emits `AT_PHI` in salty output, in lockstep.

## Changes
- `src/AST.hs` — canonical construction, reader clause, comments
- `src/XMIR.hs` — outer `as="φ"` for number and string data objects
- `src/Sugar.hs` — `φ` instead of `as-bytes` in salty printing
- `test-resources/atoms/primitives.js` — answers with `Φ.number( φ ↦ … )`, recognizes new φ shapes first, legacy `as-bytes`/`data` shapes stay
- fixture universes in `test/DataizeSpec.hs` and `test/CLISpec.hs` — declare number/string with a φ void and keep `as-bytes` as a read (`$.φ`), the way they were rewritten for bytes in #1142
- expected salty/XMIR outputs in `test/SugarSpec.hs`, `test/ASTSpec.hs`, `test/CLISpec.hs`, `test/DataizeSpec.hs` and the to-salty / rewriter / xmir-printing packs

## Tests
- New `PrinterSpec` case pins the invariant: salty output names both datum bindings φ and never mentions `as-bytes`
- The reduction-label sequence for `5.plus(6)` shortens (the `dot`/`stay` echo dispatch is gone) — that is the point of the fix
- `make test`: 2414 examples, 13 pre-existing `LocatorSpec` failures (GHC 9.10 backtrace, #1077); hlint and fourmolu clean

### DIFF
--- a/src/AST.hs
+++ b/src/AST.hs
@@ -202,10 +202,11 @@ pattern BaseObject label <- (matchBaseObject -> Just label)
 
 -- Minimal matcher function (required for view pattern)
 --
--- The primitive→bytes binding is named 'as-bytes' and the bytes→payload
--- binding is named 'data' (jeo-maven-plugin 0.15.3+, following phi-calculus
--- dropping positional attributes). The legacy positional α0 form is still
--- recognized so XMIR produced by older jeo versions keeps sugaring back.
+-- The real 'number' and 'string' of eo-runtime declare '[@]' as their only
+-- void (objectionary/eo#8620), so both primitive→bytes and bytes→payload
+-- bindings are printed as 'φ' (#1155). The reader still accepts the legacy
+-- names 'as-bytes'/'data' and the positional α0 form so XMIR produced by
+-- older jeo versions keeps sugaring back.
 matchDataObject :: Expression -> Maybe (T.Text, Bytes)
 matchDataObject (ExApplication outer arg)
   | Just inner <- asBytesArg arg = case (matchOuter outer, matchInner inner) of
@@ -213,6 +214,7 @@ matchDataObject (ExApplication outer arg)
       _ -> Nothing
   where
     asBytesArg :: Argument -> Maybe Expression
+    asBytesArg (ArTau AtPhi inner) = Just inner
     asBytesArg (ArTau (AtLabel "as-bytes") inner) = Just inner
     asBytesArg (ArAlpha (Alpha 0) inner) = Just inner
     asBytesArg _ = Nothing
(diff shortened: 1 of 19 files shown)


yegor256:
@VasilevNStas fixed already in #1156

Issue #1082, opened 2026-09-05
--output=xmir on a non-top-level formation fails only at print time; --sequence --output=xmir concatenates XML documents


VasilevNStas:
`--output=xmir` on a non-top-level formation fails only at print time, after all the work; `--sequence --output=xmir` concatenates multiple XML documents

### 1. Late failure for non-top-level formations

`expressionToXMIR` (`src/XMIR.hs:170-175`) only accepts a top-level expression of the shape `[BiTau (AtLabel _) arg, BiVoid AtRho]`; anything else falls through to `throwIO (UnsupportedTopExpression expr)` at `XMIR.hs:263`. The validator `validateXmirOptions` (`src/CLI/Validators.hs:55-59`) checks only `--focus` and the omit flags — it does **not** check the expression shape.

So a valid phi input whose top level does not have the expected `label ↦ …, ρ ↦ ∅` shape runs through the entire rewriting and only then dies on printing:

```
$ phino rewrite --output=xmir '⟦ x ↦ 1, ρ ↦ 2 ⟧'   -- note: ρ ↦ 2, not ρ ↦ ∅
[ERROR]: XMIR does not support such top-level expression: ⟦ x ↦ 1, ρ ↦ 2 ⟧
```

The user pays the whole computation cost and gets the failure at the very end.

### 2. `--sequence --output=xmir` concatenates several XML documents

`printRewrittens` (`src/CLI/Helpers.hs:106-124`) prints each intermediate expression via `printInFormat`, and for XMIR each one is a full XML document (`<?xml …?>` header + `<object>` root). Joining them with `"\n"` yields a single stream of multiple XML documents, which is not parseable as one XML file. `--headers` even interleaves text lines between documents.

## Repro

```
$ phino rewrite --sequence --output=xmir '⟦ x ↦ 1, ρ ↦ ∅ ⟧'
<?xml version="1.0" encoding="UTF-8"?>
<object ...>...</object>
<?xml version="1.0" encoding="UTF-8"?>
<object ...>...</object>
```

## Expected

- Validate the top-level shape up front (fail before rewriting), or broaden `expressionToXMIR` to accept the shapes `rewrite`/`dataize` actually produce.
- For `--sequence --output=xmir`, either wrap the steps in a single document (e.g. `<o>` children under one `<object>`), or reject the combination early with a clear message.

## Impact

[...]

PR #1158, opened 2026-09-10
fix(xmir): print dataization residuals as XMIR and honor rho/omit flags (#1076)

### LINKED ISSUE #1076: --partial cannot print its residual to --output=xmir (always fails); --hide-rho ignored for XMIR
`--partial` cannot print its residual to `--output=xmir` (always fails); `--hide-rho` is ignored for XMIR

## Problem

Two related gaps in the dataize output path (`src/CLI/Runners.hs:162-186`):

### 1. `--partial` + `--output=xmir` always fails

The residual program is printed via `printFocused` → `printInFormat XMIR` → `expressionToXMIR`, which only accepts a top level of the form `[BiTau (AtLabel _) arg, BiVoid AtRho]` (`src/XMIR.hs:170-175`, fallback `throwIO (UnsupportedTopExpression expr)` at `:263`). A residual such as

```
⟦ x ↦ ⟦ λ ⤍ Sym_arg_0 ⟧, λ ⤍ L_number_plus, ρ ↦ … ⟧
```

has three bindings (not the two expected) and a non-`∅` ρ, so it throws `UnsupportedTopExpression`. The validator does **not** forbid this combination — unlike `--evaluations`, which explicitly requires `--output=phi` at `Runners.hs:176-178`.

Repro (verified):
```
$ phino dataize --partial --output=xmir \
    '⟦ bytes(data) ↦ ⟦ φ ↦ data ⟧, number(as-bytes) ↦ ⟦ φ ↦ as-bytes, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧, times(x) ↦ ⟦ λ ⤍ L_number_times ⟧ ⟧, φ ↦ 2.times(3).plus(⟦ λ ⤍ Sym_arg_0 ⟧) ⟧'
[ERROR]: XMIR does not support such top-level expression: ⟦ x ↦ …
```

The error appears only at print time, after all the dataization work — there is no early validation, and `--partial`'s core promise (a residual program the user can inspect/fix) is broken for XMIR.

### 2. `--hide-rho` is ignored for `--output=xmir`

`printPhi` (`src/CLI/Helpers.hs:147-149`) honors `_hideRho`; the XMIR path (`printInFormat XMIR`) does not. README's "Remove every ρ binding from result and intermediate expressions" is therefore not honored for XMIR output.

### 3. `--omit-listing`/`--omit-comments` dead in dataize

The dataize `PrintCtx` uses `defaultXmirContext` (`Runners.hs:186`), so the `--omit-listing`/`--omit-comments` flags parsed for dataize (`Parsers.hs:289-290`, fields `_omitListing`/`_omitComments` in `CLI/Types.hs:84-85`) are never read — silently ignored options.

## Call-site inventory

| Issue | Location |
|---|---|
| residual → XMIR crash | `Runners.hs:162-166` → `Helpers.printFocused` → `Helpers.printInFormat XMIR` → `XMIR.hs:170-175,263` |
| no validation of the combo | `Validators.hs:55-59` (checks only `--focus`/omit flags, not `--partial`) |
| `--evaluations` precedent (requires phi) | `Runners.hs:176-178` |
| `--hide-rho` only in `printPhi` | `Helpers.hs:147-149` vs `:143` |
| omit flags never read | `Runners.hs:186` `defaultXmirContext` |

## Expected

Either validate and forbid `--partial` + `--output=xmir` up front (as `--evaluations` does), or make the residual printable as XMIR (accept a top level with a `λ`/`ρ` binding). Make `--hide-rho` and the omit flags consistent across output formats.

VasilevNStas:
Fixes #1076.

## Problem
`expressionToXMIR` accepted only a top of the exact shape `⟦ name ↦ expr, ρ ↦ ∅ ⟧`, so `dataize --partial --output=xmir` always threw `UnsupportedTopExpression` — at print time, after all the dataization work: a residual is an arbitrary formation (`x ↦ …, λ ⤍ L_number_plus, ρ ↦ …`). On top of that `--hide-rho` was silently ignored for XMIR, and `--omit-listing/--omit-comments` never reached the writer in `dataize`/`morph`. The answer to the question in the issue is the (implicit) one from the cross-ref to objectionary/eo#8548: the `eo-lowering` engine needs residuals **as XMIR**, so option 2 — make them printable, rather than option 1 — forbid the combo.

## Fix
1. A second clause of `expressionToXMIR` (src/XMIR.hs) prints any formation that yields at least one element node: every binding becomes a child of `<object>`, the same `nestedBindings` the inner formations already use. The program path (package extraction, single root binding) is unchanged, just factored into `programToXMIR`/`documentWith`.
2. `xmirToPhi` reads a multi-binding `<object>` back (`uniqueBindings` + `withVoidRho`, same as nested formations), so residual → XMIR → residual round-trips. Stray children (processing instructions, bare text) are still rejected.
3. `XmirContext` gains `_hideRho`: `nestedBindings` drops every bound ρ, matching `printPhi`.
4. The `dataize`/`morph` runners build the context from their parsed options (with the 𝜑-text listing, like `rewrite`), so the dead flags come alive.

The `rewrite` validator from #1082 stays as is: the input of `rewrite --output=xmir` must keep the program shape; residuals reach XMIR through `dataize`/`morph`, which have no such constraint.

## Changes
- `src/XMIR.hs` — residual printing clause, multi-`<o>` reader, `_hideRho`, `strayNodes`, factored `programToXMIR`/`documentWith`
- `src/CLI/Runners.hs` — `XmirContext` built from options in `runDataize` and `runMorph`
- `test/XMIRSpec.hs` — new round-trip + `--hide-rho` cases; the two newly-legal shapes moved out of the prohibit list
- `test-resources/xmir-parsing-packs/multiple-root-formations.yaml` — flipped from failure to a passing round-trip
- `test/CLISpec.hs` — `--partial --output=xmir` prints residual (x/λ/ρ children, real listing by default), `--hide-rho` + `--omit-listing` honored

## Tests
- `make test`: 2418 examples, 13 pre-existing `LocatorSpec` failures (GHC 9.10 backtrace); hlint/fourmolu clean
- Verified on the repro from the issue (fresh φ-literals): `--partial --output=xmir` exits 0 and the printed document reads back to the same 𝜑

### DIFF
--- a/src/XMIR.hs
+++ b/src/XMIR.hs
@@ -24,6 +24,7 @@ where
 import AST
 import Bytes (btsIsUtf8, btsSize, btsToNum, btsToStr, bytesToBts)
 import Control.Exception (Exception (displayException), throwIO)
+import Control.Monad (unless)
 import Data.Bifunctor (bimap)
 import Data.Foldable (foldlM)
 import Data.List (intercalate)
@@ -48,6 +49,7 @@ import qualified Text.XML.Cursor as C
 data XmirContext = XmirContext
   { _omitListing :: Bool
   , _omitComments :: Bool
+  , _hideRho :: Bool
   , _listing :: Expression -> String
   }
 
@@ -58,7 +60,7 @@ gitRevision :: String
 gitRevision = take 7 $(gitHash)
 
 defaultXmirContext :: XmirContext
-defaultXmirContext = XmirContext True True (const "")
+defaultXmirContext = XmirContext True True False (const "")
 
 data XMIRException
   = UnsupportedTopExpression Expression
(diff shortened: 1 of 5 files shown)


src/XMIR.hs
```diff
@@ -173,56 +175,48 @@ namedBinding name expr ctx = do
...
+-- formation carries becomes a child of <object>; 'xmirToPhi' reads the list
+-- back (#1076)
+expressionToXMIR expr@(ExFormation bds) ctx =
+  documentWith ctx [] expr rootNodes
+  where
+    rootNodes :: IO [Node]
+    rootNodes = do
+      roots <- nestedBindings bds ctx
+      unless (any isElement roots) (throwIO (UnsupportedTopExpression expr))
+      pure roots
```
zaguzovmaksim0-hue, line 207: `rootNodes` returns every residual binding as a sibling root `<o>`, but the current XMIR schema allows at most one direct `<o>` under `<object>`: `XMIR.xsd` (gh-pages), complex type `object`, line 496 is `<xs:element name="o" type="o" minOccurs="0" maxOccurs="1"/>`. Thus a residual like `[[ x -> 4, L> L_number_plus, ^ -> ... ]]` produces multiple top-level `<o>` elements and is schema-invalid even though `xmirToPhi` now accepts it. The test `multiple-root-formations.yaml` is therefore being flipped to success for a shape the authoritative schema rejects. Please encode the arbitrary formation beneath a single schema-valid root `<o>` (or land the required XMIR schema change first) and have the test validate the emitted XML against XMIR.xsd, not only round-trip it through phino's relaxed reader.

zaguzovmaksim0-hue (changes requested):
The residual output shape is not valid against the XMIR schema the generated document itself references. Current `objectionary/eo` gh-pages `XMIR.xsd` defines top-level `object` with `<xs:element name="o" type="o" minOccurs="0" maxOccurs="1"/>`; this branch emits one direct `<o>` per residual binding. Please choose a schema-valid representation (for example one root `<o>` containing the residual formation) or coordinate a schema change before making this the XMIR output contract, and add schema validation to the regression test.

VasilevNStas:
Sorry for the delay in replying — the follow-up is filed as #1159: the residual is now encoded under one attribute-free `<o>` (the `o` type is mixed with unbounded nested `<o>`, so it validates against the very `XMIR.xsd` the document references), multiple root `<o>`s are rejected again, and the `objectionary` job gained an `xmllint --schema` step over a freshly printed residual. Round-trip, `--hide-rho` and the omit-flag tests keep passing unchanged. Verified locally with xmllint.

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (251 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2464](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (1) comments.  Your running score is [+207](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1078, opened 2026-09-05
Atoms loop to --max-steps when the universe lacks true/false/number (wider than NaN); --partial gives no residual on OutOfSteps


VasilevNStas:
Atoms loop to `--max-steps` when the universe lacks `true`/`false`/`number` (wider than NaN); `--partial` gives no residual on `OutOfSteps`

## Problem

`src/Dataize.hs` fires an atom and wraps its result in an object that must then be **morphed and dataized through the universe**:

- Number atoms return `DataNumber (numToBts …)` (`Dataize.hs:489-521`), which renders as `Φ.number(as-bytes ↦ Φ.bytes(data ↦ ⟦Δ ⤍ …⟧))` — the AST's `DataNumber` construction (`AST.hs:227-231`).
- `L_number_gt` returns `BaseObject "true"/"false"` (`Dataize.hs:516-521`), which needs `univ.false`/`univ.true` in the universe.

Normalization of these results requires the `universe` morphing rule (`resources/morphing.yaml`) and then a dispatch `univ.number` / `univ.false` in the universe. When the universe does not provide those objects, the dispatch misses and the 𝕄/𝔻 recursion cycles through `md → ma → universe → mf → mphi → ml` until `deeper` (`Dataize.hs:119-122`) throws `OutOfSteps`. This is **not specific to NaN** — it reproduces with entirely ordinary operands.

## Repro (verified on `./phino`)

Plain numbers, no NaN:

```
$ phino dataize --max-steps=40 \
    '⟦ φ ↦ ⟦ λ ⤍ L_number_div, ρ ↦ ⟦ Δ ⤍ 40-45-00-00-00-00-00-00 ⟧, x ↦ ⟦ Δ ⤍ 40-14-00-00-00-00-00-00 ⟧ ⟧ ⟧'
[ERROR]: Dataization did not finish before reaching the limit of steps: --max-steps=40
```

`5.gt(Φ.nan)` without `true`/`false` in the universe:

```
$ phino dataize --max-steps=40 '⟦ φ ↦ 5.gt(Φ.nan) ⟧'
[ERROR]: Dataization did not finish before reaching the limit of steps: --max-steps=40
```

NaN via `5.plus(Φ.nan)`, and `2.div(0)` (Infinity), behave the same way. By contrast, when the universe does provide `number`/`true`/`false`, the same expressions dataize fine — so the loop is a property of the **atom result shape**, not of the operand.

## Relation to existing issues

[...]

Issue #1052, opened 2026-09-01
dataize never stops on undataizable input: morph recursion in Dataize.hs ignores --max-cycles/--depth-sensitive


yegor256:
Running `phino dataize --max-cycles 10 --max-depth 10 --depth-sensitive` (0.0.111, master a7b8382) on a small expression that cannot be dataized never terminates and must be killed externally:

```
printf '⟦ @ ↦ ⟦ λ ⤍ L_number_div, ρ ↦ ⟦ Δ ⤍ 40-45-00-00-00-00-00-00 ⟧, x ↦ ⟦ Δ ⤍ 40-00-00-00-00-00-00-00 ⟧ ⟧ ⟧' > bare.phi
phino dataize --max-cycles 10 --max-depth 10 --depth-sensitive bare.phi
```

With `--log-level DEBUG` the morph chain can be seen cycling through the rules `dot`, `delta`, `copy`, `md`, `universe`, `stay`, `mphi`, `ml`, `mf` and restarting the bounded rewrite phase ("Starting rewriting cycle for all rules: 0 out of 10") 646 times within the first ten seconds, without progress.

This is wrong because `--max-cycles`/`--max-depth` only reach the `normalized` calls inside individual morph steps, while the recursion that actually loops is unbounded: `morph` calls itself at src/Dataize.hs:141 and src/Dataize.hs:146, and `dataize'` calls itself at src/Dataize.hs:224, :231 and :243, none of them counting anything. As a result `--depth-sensitive`, documented as "fail if rewriting is not finished after reaching max attempts", can never fire during dataization, and there is no CLI way to bound a `dataize` run.

A concrete fix: thread a step counter through the `morph`/`dataize'` recursion in src/Dataize.hs, count each morph step against `--max-cycles` (or a dedicated option), and when the budget is exhausted throw the same failure that `--depth-sensitive` produces for rewriting.

yegor256:
@maxonfjvipon could you take a look when you have a moment? Happy to clarify anything in the report

yegor256:
@rultor release, tag is `0.0.112`

0crat:
@yegor256 Thanks for closing this issue! You've earned +12 points for this: +8 as a basis; +4 for resolving it in less than 12 hours. Your running score is [+1760](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1161, opened 2026-09-10
fix(dataize): park --max-steps exhaustion as a residual under --partial (#1078)

VasilevNStas:
Partial fix for #1078 — the "regardless?" half. The design question (whether atom results should be universe-independent, options 1 vs 2 in the issue and in my 09-07 question) is **not** decided by this PR and stays open for the maintainers.

## Problem
`deeper` throws \`OutOfSteps\` **before** `parking` can attach a derivation, so under `--partial --max-steps=N` a term that never reduces to bytes — a missing dispatch cycling through the universe rules (repro from #1134: `5.gt(Φ.nan)` loops even with **no atom fired at all**), or an atom result the universe cannot consume — failed hard instead of yielding the residual program that `--partial` promises. The behavior is re-verified on the current master (comment in #1078).

## Fix
The spent budget is a stuck site: the workflow "give me the stuck place so I can fix it" applies to it exactly as to an unregistered atom.
1. A spine frame now also parks \`OutOfSteps\`, as the new \`OutOfStepsAt\` carrying that frame's derivation; its \`Show\` stays word-for-word the old message, so **non-partial runs are unaffected**.
2. \`unparked\` strips the derivation off the signal escaping a side-computation, mirroring \`StuckAt\`.
3. \`dataize\` and \`morph\` turn \`OutOfStepsAt\` into \`Residual\` under \`_partial\` — the same one-line-adjacent branch as \`StuckAt\` already has.

## Changes
- `src/Dataize.hs` — \`OutOfStepsAt\`, \`parking\`/\`unparked\` and the two wrapper branches (52+/10-)
- `test/DataizeSpec.hs`, `test/CLISpec.hs` — three new cases; a lone-\`it\` \`describe\` gains the \`$ do\` it now needs

## Tests
- `dataize --partial --max-steps=40` on the #1052 div-loop now exits 0 with the parked `Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ 40-35-… ⟧ ) )` residual; without `--partial` the message is unchanged
- `morph --locator=Q.@ --max-steps=10 --partial` parks `5.gt( Φ.nan )`
- `make test`: 2421 examples, 13 pre-existing `LocatorSpec` failures (GHC 9.10.3); hlint/fourmolu clean

### DIFF
--- a/src/Dataize.hs
+++ b/src/Dataize.hs
@@ -95,11 +95,17 @@ data DataizeException
     -- everything reduced before it already in place: the residual program that
     -- '_partial' turns into the 'Residual' outcome.
     StuckAt T.Text (NonEmpty Rewritten)
+  | -- An 'OutOfSteps' caught by a spine frame, carrying that frame's derivation
+    -- just like 'StuckAt': a term that never reduces is a stuck site too, so
+    -- '_partial' parks it and hands back the residual instead of failing hard
+    -- (#1078)
+    OutOfStepsAt Int (NonEmpty Rewritten)
   deriving anyclass (Exception)
 
 instance Show DataizeException where
   show (OutOfSteps limit) =
     printf "Dataization did not finish before reaching the limit of steps: --max-steps=%d" limit
+  show (OutOfStepsAt limit _) = show (OutOfSteps limit)
   show (Stuck func) = printf "Atom '%s' does not exist" (T.unpack func)
   show (StuckAt func _) = show (Stuck func)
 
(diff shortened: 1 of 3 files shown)


VasilevNStas:
@yegor256 
please review this

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2496](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1162, opened 2026-09-10
Let an atom ask phino to reduce its operand

### LINKED ISSUE #1160: The --atoms channel is one-way, so a script must start a new phino to reduce an operand (src/Atoms.hs:399)
The `--atoms` channel is one-way, so an atom that needs an operand reduced has to start a whole new phino.

`asked` (`src/Atoms.hs:392`) writes one request and then reads exactly one line, `BC.hGetLine _output` at `src/Atoms.hs:399`; a reply carrying any other `id` is rejected as an error at `src/Atoms.hs:424`. The script therefore has no way to say anything to phino except "here is your answer". When an operand arrives unreduced — which is normal, since a `!` handle reaches an atom as `Φ.dataized( α0 ↦ … ).as-bytes` and cannot reduce until that very atom fires — the script's only recourse is to serialise `𝑏` back to text, splice it into the universe text, and run `phino` on the result.

That means phino re-parses an AST it emitted moments earlier. Measured on 0.0.123 against a pruned `eo-runtime` universe: the universe alone is 90 KB and `phino rewrite --flat` reads it in 0.04 s, while the spliced expression the script must hand back is 1479 KB and takes 1.48 s to read, out of 3.21 s for the whole probe. The 1.4 MB is not the universe; it is the inlined ρ chain of `𝑏` pasted back as text. One two-void fragment costs 3 m 15 s this way, and essentially all of it is these nested runs. This is the sibling of #1121, which removed the process phino spawned per fire; what remains is the process the script spawns per operand.

Two further costs come from the same place. `--inside` puts the expression on argv, so a real atom body of a few hundred kilobytes dies with `spawnSync phino E2BIG`, which is why the splice exists at all; and each nesting level then needs its own probe name, or the splice fails with `Duplicated attribute 'probe'`.

Suggested fix: let the program write a request of its own, `{"id": <fresh>, "ask": "<φ-expression>"}`, and turn the single `hGetLine` at `src/Atoms.hs:399` into a small loop — a line carrying `𝑛` with the matching `id` is the answer, a line carrying `ask` is served and replied to. Serving it means re-entering the evaluator, which both call sites can already supply: `src/Dataize.hs:409` and `src/Dataize.hs:657` call `fireAtom` from `IO` with the universe and the full context in hand, so a reducer can be passed down into `asked`. A program that never asks behaves exactly as today, so `serve: false` and `rt: exec` are untouched, and in-process nesting is bounded by the existing `--max-depth`.

Once the channel carries questions, `𝑏` need not carry the receiver either: in the calls we logged, `𝑏` with its ρ chain hidden is 70 characters against 1.4 MB with it, and an atom that can ask for `ρ` only when it needs it would not have to be handed the whole chain. That is a separate change and not part of this ask. Both together are the answer to the largest open cost question of objectionary/eo#8548.

yegor256:
The `--atoms` channel now carries questions, not just answers. A program writes a line of its own — an `id` it mints and the 𝜑-expression under `ask` — and phino serves it by re-entering its own evaluator, binding the expression to a synthetic attribute of the universe and dataizing it there, the same trick `--inside` plays. The answer comes back with that `id` under `𝑛`. A program that never asks sees exactly what it saw before.

An operand reaches an atom as it was written, and reducing it can take the very atom being fired, so until now a program's only recourse was to serialise the formation back to text, splice it into the universe and run a whole phino of its own on the result. That re-parse is where the time went. Only a program kept for the run may ask, since phino closes the stdin of one started for the fire behind its request, and a question from such a program fails the fire with a message saying so.

Serving a question can fire the same atom again, which arrives while the question is still open, so the session is let go of before the program is spoken to and its request counter moved into an IORef. The nesting is bounded by the run's step budget, `--max-steps`. Hiding the ρ chain from `𝑏` is the other half of the cost the issue measures and is left for a separate change.

Closes #1160

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,9 +173,10 @@ raw result, normalizing it exactly as it normalizes anything else, so
 A program started for the fire is asked one request, always `id` 1, and its
 `stdin` is closed behind it, so it may read its input whole or line by line, as
 it pleases. It is waited for once it has answered, and a non-zero exit fails
-the run. So does a reply that is not JSON, carries no `𝑛`, answers another
-`id`, or an `𝑛` that does not parse, or a program that quits without
-answering — always with the program's own `stderr` in the message.
+the run. So does a reply that is not JSON, carries neither `𝑛` nor `ask` (the
+next section is about `ask`), answers another `id`, or an `𝑛` that does not
+parse, or a program that quits without answering — always with the program's
+own `stderr` in the message.
 
 Each key of the registry is a regular expression, and it must match the whole
 λ name, so a plain name such as `L_number_plus` means that one atom and nothing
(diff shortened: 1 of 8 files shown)


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any part of the protocol change if something reads odd.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (573 >= 100); -8 for way too many hits-of-code (573 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2363](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1163, opened 2026-09-10
test(atoms): pin a looping ask question to the parked residual under --partial

VasilevNStas:
Follow-up to #1161 over #1162 — **tests only**, no src changes. Part of #1078 (the design question there stays open).

## Why
#1162 gave resident programs the \`ask\` channel; #1161 taught \`--partial\` to park an exhausted budget. The seam between them is untested: when a program asks phino to reduce a term the universe can never finish reducing, the inner dataization behind the question used to throw \`OutOfSteps\` before any frame could park it — today it parks, answers the question with the residual, and the run lives. Nothing pins that, and both features are young enough to regress into each other.

## Changes
- `test-resources/atoms/asking-loops.js` — a resident \`L_number_gt\` that asks about \`Φ.nan.gt( 1 )\` (a term that cycles in any universe that lacks it) and answers every request with \`⟦ Δ ⤍ 2A- ⟧\` whatever comes back, so a run ending on `2A-` is proof from the program's side of the channel
- `test/Fixtures.hs` — `withLoopingAskRegistry`, mirroring the existing `withAskingRegistry`
- `test/CLISpec.hs` — two cases: `--partial` ends on `2A-` (exit 0); without it the run fails with the #1052 message

## Tests
- Deterministic: pure state machine over a frozen budget (`--max-steps=200`), no timing dependence — verified twice on the same binary
- `make test`: 2430 examples, 13 pre-existing `LocatorSpec` failures (GHC 9.10.3); hlint/fourmolu clean; 79+/3-

### DIFF
--- a/test-resources/atoms/asking-loops.js
+++ b/test-resources/atoms/asking-loops.js
@@ -1,0 +1,39 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 Objectionary.com
+// SPDX-License-Identifier: MIT
+
+// The λ function 'L_number_gt', answered by a resident program that asks
+// phino to reduce a term the universe can never finish reducing:
+// 'Φ.nan.gt( 1 )' walks dispatches into a cycle that ends only when the
+// step budget dies. Under '--partial' phino parks where the term cycles and
+// answers the question with the residual (#1078), so the run lives on;
+// without the flag the exhausted budget fails it, exactly as before the ask
+// channel existed (#1160). This script answers every request with the bytes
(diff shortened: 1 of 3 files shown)


VasilevNStas:
@yegor256 
plz review this

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2512](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1079, opened 2026-09-05
meetCondition silently swallows exceptions from when/having conditions (dataize/matches errors become condition not met)


VasilevNStas:
`meetCondition` silently swallows exceptions from `when`/`having` conditions (dataize/matches errors become "condition not met")

## Problem

`src/Rule.hs:270-280`:

```haskell
meetCondition :: Y.Condition -> [Subst] -> RuleContext -> IO [Subst]
meetCondition _ [] _ = pure []
meetCondition cond (subst : rest) ctx = do
  met <- try (meetCondition' cond subst ctx) :: IO (Either SomeException [Subst])
  case met of
    Right first -> ...
    Left _ -> meetCondition cond rest ctx      -- ← exception swallowed
```

Any exception thrown while evaluating a `when`/`having` condition is silently discarded and the substitution is treated as "condition not met". Two concrete sources:

1. **`_matches`** (`Rule.hs:213-220`): runs `_buildTerm ctx "dataize" [Y.ArgExpression expr] subst`, which can throw `userError` from `Functions._dataize` ("Only data objects and bytes are supported…") when the expression dataizes to a non-data object. An error in `matches` becomes a silent "no match".
2. **`_partOf`** (`Rule.hs:222-226`): uses the throwing `buildExpressionThrows`/`buildBindingThrows` (`Builder.hs:145-155`), which throw `CouldNotBuildExpression`/`CouldNotBuildBinding` when a meta is unbound. Again swallowed.

This is **inconsistent** with `extraSubstitutions` (`Rule.hs:287-329`), which does **not** wrap in `try` — there the same class of `_buildTerm` errors propagates to the caller. So one pipeline swallows, the other propagates, for the same underlying failures.

Also inconsistent internally: `_in`, `_disjoint`, `_eq`, `_gt` use the non-throwing `buildAttribute`/`buildBinding` (Either) and correctly yield `[]`; only `_matches`/`_partOf` use the throwing variants, and their exceptions are the ones silently dropped.

## Repro sketch

A `when` condition with a `matches` clause whose dataization fails produces an empty match set and a rewritten output that silently dropped the rule — with no error message, making the rule look "not applicable".

## Expected

- Use the non-throwing builders in `_partOf`/`_matches` (return `[]` on `Left`), removing the exception source altogether, or
- Log the `Left`/exception in `meetCondition` instead of full silence (a debug line), so a broken condition is at least observable.

## Impact

[...]

PR #1164, opened 2026-09-10
fix(rules): report a condition whose evaluation raised, before treating it as not met (#1079)

VasilevNStas:
Part of #1078-style policy is NOT decided here — this PR takes only the second bullet of #1079 ("log the exception instead of full silence"), which changes no behavior. The option-1-vs-2 question (non-throwing builders vs propagating) stays open for the maintainers.

## Problem
`meetCondition` (src/Rule.hs:288) catches every exception from a `when`/`having` condition and drops the substitution without a word. A `matches` whose `dataize` throws ("Only data objects and bytes are supported..."), or a `part_of` with an unbuildable meta, becomes indistinguishable from a condition that was legitimately not met — unobservable in the logs, despite every nearby pipeline (`extraSubstitutions`) logging plenty.

## Fix
Log the raise at debug level right where it is swallowed — condition, substitution outcome unchanged, exception displayed — then continue exactly as before. `--log-level=debug` now shows:
```
[DEBUG]: Condition Matches "hello" (ExMeta "e1") raised and was treated as not met: user error (Only data objects and bytes are supported by ...
```
No rule changes behavior: same substitutions dropped, same rewritten output; only an added debug line when a condition raises.

## Changes
- `src/Rule.hs` — `Left err` branch logs before dropping (+`displayException` import)
- `test-resources/cli/raising-condition.yaml` — fixture rule whose `when` dataizes a formation (the exact throw source named in the issue)
- `test/CLISpec.hs` — asserts the debug line AND that the output is what the silent drop always produced

## Tests
- `make test`: 2429 examples, 13 pre-existing `LocatorSpec` failures (GHC 9.10.3); hlint/fourmolu clean; 31+/1-

### DIFF
--- a/src/Rule.hs
+++ b/src/Rule.hs
@@ -16,6 +16,7 @@ import Builder
   , buildExpressionThrows
   )
 import Bytes (btsToUnescapedStr)
+import Control.Exception (Exception (displayException))
 import Control.Exception.Base (SomeException, try)
 import Control.Monad (when)
 import qualified Data.ByteString.Char8 as B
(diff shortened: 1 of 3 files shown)


VasilevNStas:
@yegor256 
please review this PR

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (32) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2524](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1166, opened 2026-09-10
feat(atoms): serve lean receivers and questions that name operands (#1165)

### LINKED ISSUE #1165: Atoms channel inlines the whole ρ chain into `𝑏`: 23.6 MB and 21.5 s for ten fires (src/Atoms.hs:445)
`asked` renders the atom's formation in full at `src/Atoms.hs:445`:

```haskell
request number = lined (object ["id" .= number, "λ" .= func, "𝑏" .= rendered form])
```

`rendered` (`src/Atoms.hs:545`) is `printExpression'`, so `𝑏` carries the receiver's whole ρ chain inlined as text, and that chain climbs to Φ, which is the universe. A script that only wants `x` or `target` is handed the program.

Measured on phino 0.0.125, with the eo-lowering prototype of objectionary/eo#8548 and a pruned eo-runtime universe of 89,696 bytes. One `[a b] > gap` fragment is ten fires. With ρ hidden every `𝑏` is 54 to 105 characters, 771 in total. On the wire the same ten are 23.6 MB, and the largest is 8,105,772 bytes, 90 times the universe it came from. Printing them costs 21.5 s of the 94 s run; parsing the ten questions back costs another 3.5 s; the script's own JSON handling is 0.04 s, so the whole of it is phino's serialization.

It also compounds, because a question is quoted text. `served` (`src/Atoms.hs:461`) parses it and `reduction` binds it to a synthetic universe attribute before dataizing, so the next `𝑏` printed inside that question inlines it:

```
fire  𝑏 bytes   synthetic attributes present
#7      236,159  --
#8    1,605,678  a🌵7
#9    4,344,732  a🌵7 a🌵8
#10   8,105,772  a🌵7 a🌵8 a🌵9
```

Each nesting level multiplies the message by two to three. Nothing leaks (`#7` carries none of `a🌵1`..`a🌵6`), so this is depth, not accumulation, but three levels are enough to turn 236 KB into 8 MB.

Two changes, both small. First, render `𝑏` with `printExpressionHidingRho'`, which already exists at `src/Printer.hs:50` and is already used for output at `src/CLI/Helpers.hs:179`. Second, add a question that names an operand instead of quoting it: a third constructor of `Said` (`src/Atoms.hs:265`) for `{"id": <minted>, "of": <request>, "attr": "<name>"}`, served from the formation phino already holds for that request rather than from parsed text. Then a question is about 30 bytes, no synthetic attribute is minted, and the compounding is gone.

The two belong together, since ρ is where the value of a receiver lives: with `L_number_plus`, ρ is the left operand. Fetching it unreduced is much cheaper than today but not free -- the receiver of these fires is `Φ.number` with all its attributes, 44,048 bytes with ρ hidden -- while reducing it by reference answers `Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ S1 ⟧ ) )`, 50 characters. So `attr` wants both readings, one that hands the node over and one that dataizes it the way `ask` does today.

A lean `𝑏` cannot be unconditional: a transient program cannot ask, and `served` already fails such a fire with `AtomMute` at `src/Atoms.hs:463`, so an `rt: exec` atom that computes on its receiver would lose it with no way to get it back. Tying the lean channel to `serve: true` needs no new flag: a program gets the short `𝑏` exactly when it is able to ask for what is missing.

Worth noting as a consequence, not as part of the ask: `𝑒` (`src/Atoms.hs:443`) then has no reader left. It was only ever needed so a script could re-feed the universe to a phino of its own, which objectionary/phino#1160 already removed the reason for.

This is the follow-on that #1160 named, and the last cost item of objectionary/eo#8548.

VasilevNStas:
Fixes #1165. Protocol interpretation as stated in my comment there (lean `𝑏` tied to `serve`; `of`+`attr` with an optional `reduce` boolean) — if you wanted a different discriminator for the two readings, it is a one-line change.

## Problem
`asked` rendered the atom's receiver in full at every fire; the ρ chain climbs to Φ, and a question that quotes a receiver compounds it: measured on the eo-lowering prototype — ten fires, 23.6 MB on the channel, 21.5 s of printing, 8 MB in the largest single message (90 times the universe), each nesting level multiplying by 2–3 through synthetic `a🌵` attributes.

## Fix
Both changes from the issue, together, as they belong:

1. **Lean channel for kept programs**: `𝑏`, `𝑒` and every answer to an Open (`serve: true`) program are rendered through `printExpressionHidingRho'`. A transient program keeps the full receiver — it cannot ask, so the leanness is tied to `serve` exactly as the issue proposes, no new flag.
2. **Questions by reference**: a third `Said` constructor — `{"id": q, "of": <request>, "attr": "x"[, "reduce": true]}` — served from the receiver phino still holds for that in-flight request (session bookkeeping: `Running._forms`, inserted at the fire, forgotten once the request is answered). Without `reduce` the node is handed over as it is; with `"reduce": true` it is dataized the way `ask` does. An unknown `of` and an unknown `attr` fail the fire with a message naming which.

Consequences handled: `asking.js` migrated from the `${b}.ρ` quoting trick (dead under lean `𝑏`) to naming its operands — its CLISpec case passes unchanged; the text-`ask` path keeps behavior and coverage (`asking-loops.js`, AtomsSpec cases). `𝑒` stays for now, as the issue notes it is a consequence, not part of the ask.

## Changes
- `src/Atoms.hs` — `Running._forms`, `Said.Reference`, `spelled`/`lean` rendering, `referenced` serving, `forget` bookkeeping, module docs
- `test-resources/atoms/asking.js` — operands by reference
- `test/AtomsSpec.hs` — seven cases: node reading, reduced reading, reduce receives the very node, both refusals, lean `𝑏` for a resident vs full `𝑏` for a transient
- `README.md` — protocol section: exchange example, the two readings, the lean-channel rationale, reply grammar

## Tests
- `make test`: 2435 examples, 13 pre-existing `LocatorSpec` failures (GHC 9.10.3); hlint/fourmolu clean
- Manual end-to-end on the issue's shapes: resident `of`+`attr:"x"` → the node arrives with no re-parse; `attr:"ρ", reduce:true` → `40-14-…` (left operand, as today); a resident `𝑏` carries no `ρ ↦`, a transient one still does; by-reference questions mint no universe attribute, so the compounding table from the issue cannot reproduce

### DIFF
--- a/README.md
+++ b/README.md
@@ -173,10 +173,10 @@ raw result, normalizing it exactly as it normalizes anything else, so
 A program started for the fire is asked one request, always `id` 1, and its
 `stdin` is closed behind it, so it may read its input whole or line by line, as
 it pleases. It is waited for once it has answered, and a non-zero exit fails
-the run. So does a reply that is not JSON, carries neither `𝑛` nor `ask` (the
-next section is about `ask`), answers another `id`, or an `𝑛` that does not
-parse, or a program that quits without answering — always with the program's
-own `stderr` in the message.
+the run. So does a reply that is not JSON, carries neither `𝑛`, nor `ask`, nor
+`of` with `attr` (the next section is about the questions), answers another
+`id`, or an `𝑛` that does not parse, or a program that quits without answering
+— always with the program's own `stderr` in the message.
 
 Each key of the registry is a regular expression, and it must match the whole
 λ name, so a plain name such as `L_number_plus` means that one atom and nothing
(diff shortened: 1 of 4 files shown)


VasilevNStas:
@yegor256 
plz review

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (276 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2536](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1168, opened 2026-09-10
Answer a question with the reduced node, not the whole universe

### LINKED ISSUE #1167: Residual Program Responds with Entire Universe Instead of the Reduced Node
Since #1165 both outbound halves of the `--atoms` channel are lean: `𝑏` reaches a `serve` program with its ρ chain hidden, and a question names its operand by reference. The answer is the half that did not move. A question is served by `reduction` (`src/Dataize.hs:669`), which binds the expression to a synthetic attribute of the universe through `insideUniverse` (`src/Dataize.hs:650`), dataizes it there, and then picks a branch at `src/Dataize.hs:676-677`: `Dataized` answers a bare `⟦ Δ ⤍ … ⟧`, but `Residual` answers the residue — the whole synthetic universe, with the reduced node bound in front of it. Under `--partial` the second branch is the ordinary one, since parking is exactly what happens when the operand reduces to something an atom left in place.

The program then gets an answer thousands of times larger than the thing it asked about. Measured on 0.0.126 with the lowering prototype of objectionary/eo#8548 over a pruned 89,696-byte `eo-runtime` universe, a ten-fire fragment sends 1,023 bytes of `𝑏` and 363 bytes of questions, and receives 650,701 bytes of answers, in which the nodes the program reads total 187 bytes. One answer is 59,081 bytes against 59,028 for that universe printed alone, and the node wanted out of it is `⟦ λ ⤍ S1 ⟧`, 17 bytes. It scales the wrong way too: the universe is one per run, but it is re-printed once per question, so a fragment asking two hundred questions pays two hundred universe prints. Phino's own reference client reads an answer as if it were the node — `test-resources/atoms/asking.js` says "which is always a byte formation, since a question is served by dataizing what it asks about" — which is the `Dataized` branch's shape, not the `Residual` one's.

The smallest fix is in `reduction`: when the outcome is `Residual`, answer with the value bound at the synthetic attribute instead of with the whole formation. Everything needed is already in hand there — `insideUniverse` returns the `aiming` context whose `_locator` names that attribute (`src/Dataize.hs:653`), and `attributeValue` (`src/Atoms.hs:528`) is the lookup. Nothing an asker can use is lost, since the rest of the residue is the universe it was already handed under `𝑒`.

yegor256:
When a program asks phino to reduce something, `reduction` binds the expression to a synthetic attribute of the universe, dataizes it there, and hands back whatever `dataize` ended on. Under `--partial` that outcome is the residue — the entire synthetic universe — so the program gets thousands of bytes wrapped around the one node it asked about. This takes the value at the synthetic attribute out of the residue instead, using the locator `insideUniverse` already aimed the run at. The `Dataized` branch is untouched.

Parking is the ordinary case with `--partial`, since that is exactly what happens when an operand reduces to something an atom left in place. On the eo-lowering prototype ten fires received 650,701 bytes of answers while the nodes actually read totalled 187, and the universe was re-printed once per question rather than once per run. Nothing useful is lost by trimming it: the rest of the residue is the universe the program was already told under `𝑒`.

One CLISpec case covers it, over a resident program whose question parks on an unregistered atom — it answers `FF-` only when phino sent the parked node alone and `00-` when the answer carried the surrounding universe. Full suite is 2439 examples and green, coverage 80%, hlint and fourmolu clean. Worth noting for later: `--inside` under `--partial` prints the same whole synthetic universe through `printOutcome`, which this change deliberately leaves alone.

Closes #1167

### DIFF
--- a/src/Dataize.hs
+++ b/src/Dataize.hs
@@ -660,7 +660,12 @@ insideUniverse expr univ ctx@DataizeContext{_buildTerm = buildTerm} = case univ 
 -- 'ReduceFunc' in 'Atoms'): the expression is bound to a synthetic attribute
 -- of the universe and dataized there, exactly the way the '--inside' option
 -- does it, so the bytes come back as a Δ formation — or, where an atom on the
--- way could not fire and '_partial' parked it, the residual program instead.
+-- way could not fire and '_partial' parked it, the node the question named,
+-- taken out of the residue at the synthetic attribute. The residue is the whole
+-- synthetic universe, and answering with it hands the program a print of the
+-- universe per question, thousands of bytes around the one node it asked about
+-- (#1167); nothing is lost by trimming it, since the rest of that residue is
+-- the universe the program was already told under '𝑒'.
 -- An operand reaches a program unreduced, since reducing it may take the very
 -- atom being fired, and before the channel carried questions the program had
 -- no way to ask: it had to splice the operand into the text of the universe
@@ -670,11 +675,11 @@ reduction :: Expression -> DataizeContext -> ReduceFunc
 reduction univ ctx expr = do
   (universe, aiming) <- insideUniverse expr univ ctx
   (outcome, _) <- dataize universe aiming
-  pure (reduced outcome)
+  reduced aiming._locator outcome
   where
-    reduced :: Outcome -> Expression
-    reduced (Dataized bytes) = ExFormation [BiDelta bytes]
-    reduced (Residual residue) = residue
+    reduced :: Expression -> Outcome -> IO Expression
+    reduced _ (Dataized bytes) = pure (ExFormation [BiDelta bytes])
+    reduced locator (Residual residue) = locatedExpression locator residue
 
 -- phino implements no λ function of its own. Which atoms exist is a property of
 -- the object model being dataized, not of the calculus, so they come from the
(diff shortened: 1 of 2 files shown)


zaguzovmaksim0-hue (approved):
I checked the extraction path against `insideUniverse` and `Locator.locatedExpression`, not just the CLI fixture.

`insideUniverse` creates exactly one synthetic `BiTau attr expr` and returns a context whose locator is `Q.<attr>`. On the `Residual` branch, `dataize` returns the residual synthetic universe, so `locatedExpression aiming._locator residue` walks that formation and returns the expression bound at the same synthetic attribute. That is the same node the asker named, while the `Dataized` branch remains the existing bare `Δ` formation.

The regression also exercises the important partial/unregistered-atom case where the old code returned the surrounding universe. CI is green and the change does not alter `--inside` output or the non-partial dataized path.

I don't see a correctness issue here. Approving.

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2415](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+152](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1176, opened 2026-09-10
#1169: Avoid overflow in extreme byte shifts

### LINKED ISSUE #1169: bytesShift on abs minBound causes integer underflow leading to out-of-bounds crash
### Problem\n\n`Bytes.btsShift` uses `abs bits * (-1)` for negative bit counts. When `bits == minBound :: Int`, `abs minBound` equals `minBound` again (two's complement overflow — `abs` cannot represent `-(minBound)`). This produces a negative `offset` that passes guards and leads to `octet` accessing array indices with `!!` using negative values — an **out-of-bounds crash** instead of a graceful error.\n\nReproducible via any code path that computes a shift amount reaching `Int.minBound`. While the direct trigger requires extreme inputs, the underlying pattern (`abs` without bound check on two's complement types) is fragile.\n\n### Location\n\n**src/Bytes.hs:432–436**\n\n```haskell\nbtsShift bits bts\n  | bits == 0 = ...\n  | bits < 0  = rightwards (abs bits * (-1)) ...  -- HERE: abs minBound = minBound!\n```\n\nInside `rightwards`:\n\n```haskell\nleftwards bytes bits = go 0\n  where\n    offset = abs bits `div` 8       -- line 433: could be negative if abs overflows\n    size   = B.length bytes\n    octet :: Int -> Word8           -- line 436\n    octet index = bytes !! index    -- line 437: !! with negative = crash\n    go source\n      | source >= 0 && source < size  = ...\n```\n\n### Steps to Reproduce\n\nCompile and run with a crafted expression that causes `btsShift` to receive `minBound` as the shift amount. For example, through binary operations in YAML rules that compute byte shifts from user input.\n\n### Expected Behavior\n\n- `btsShift` should saturate or reject extreme shift amounts gracefully.\n- Return `Nothing` or a typed exception with a clear message, not a runtime crash.\n\n### Suggested Fix\n\nAdd a guard before calling `abs`:\n\n```haskell\nbtsShift bits bts\n  | bits == 0 = ...\n  | bits == minBound = word8ToBytes (repeat 0)  -- or saturating zero result\n  | bits < 0 = rightwards (abs bits * (-1)) ...\n  | otherwise = leftwards bytes (abs bits)\n```\n\nOr use a saturating absolute value function.\n\n### Related Issues\n\nNo existing issue found. This class of problem (raw `error` crashes) was previously tracked in old Task 3 / issue #1072 family.

zaguzovmaksim0-hue:
Closes #1169.

`btsShift` now computes the absolute shift magnitude in `Integer`, so `minBound :: Int` never passes through the overflowing `abs :: Int -> Int` path.

A shift whose magnitude is at least the byte array's bit width returns an all-zero array immediately, matching the existing semantics for large shifts. For smaller shifts, `modulo` and `offset` are derived from the safe `Integer` magnitude and converted back to `Int` only after the width guard makes them representable.

Added a regression case using the actual `minBound :: Int`, rather than the existing fixed `-2147483648` case which is not `minBound` on 64-bit runners.

`git diff --check` is clean. GHC/Stack/Cabal are not installed in this Termux environment, so the repository CI is used for Haskell compilation/tests.

### DIFF
--- a/src/Bytes.hs
+++ b/src/Bytes.hs
@@ -419,9 +419,12 @@ btsSlice start len bts
 -- BtMany ["00","00"]
 btsShift :: Int -> Bytes -> Bytes
 btsShift bits bts
+  | magnitude >= toInteger size * 8 = word8ToBytes (replicate size 0)
   | bits < 0 = word8ToBytes (map leftwards indices)
   | otherwise = word8ToBytes (map rightwards indices)
   where
+    magnitude :: Integer
+    magnitude = abs (toInteger bits)
     octets :: [Word8]
     octets = btsToWord8 bts
     size :: Int
@@ -429,9 +432,9 @@ btsShift bits bts
     indices :: [Int]
     indices = [0 .. size - 1]
     modulo :: Int
-    modulo = abs bits `mod` 8
+    modulo = fromInteger (magnitude `mod` 8)
     offset :: Int
-    offset = abs bits `div` 8
+    offset = fromInteger (magnitude `div` 8)
     octet :: Int -> Word8
     octet index = octets !! index
     rightwards :: Int -> Word8
--- a/test/BytesSpec.hs
+++ b/test/BytesSpec.hs
@@ -299,7 +299,8 @@ spec = do
         , BtMany ["02", "03", "00"]
         )
       , ("large negative shift empties out", btsShift (-2147483648) (BtMany ["BF", "F0"]), BtMany ["00", "00"])
-      , ( "minimum Int shift empties out"
+      ,
+        ( "minimum Int shift empties out"
         , btsShift (minBound :: Int) (BtMany ["BF", "F0"])
         , BtMany ["00", "00"]
         )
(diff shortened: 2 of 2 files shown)


yegor256:
Merged, thanks! Computing the magnitude in `Integer` removes the `abs minBound` wrap-around, and the early guard for a shift of at least `size * 8` bits keeps `offset` inside the array. The new `minBound` case in `BytesSpec` covers what the existing `-2147483648` case could not, since that value is not `minBound` on a 64-bit `Int`.

---

0crat:
@zaguzovmaksim0-hue Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+231](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1185, opened 2026-09-11
fix(atoms): add serve guard to referenced like served (#1184)

### LINKED ISSUE #1184: A question asked by reference without `serve` hangs the run forever (src/Atoms.hs:508)
An `--atoms` program whose entry does not say `serve` may still ask a question by reference (`{"id", "of", "attr", "reduce"}`, the form added in #1165), and when it does, the whole run hangs forever, with phino and the program both at 0% CPU. `referenced` (`src/Atoms.hs:508-509`) reduces the attribute and writes the answer unconditionally, while the channel for a program started for a single fire is `Closed`: `said` (`src/Atoms.hs:556`) hands the bytes to `pushed Closed = hClose` (`src/Atoms.hs:572`), the resulting `IOError` is swallowed by the `unheard` handler (`src/Atoms.hs:613`), and phino goes back to `heard` to read a reply that the program cannot send, because it is blocked waiting for the answer phino just dropped.

The quoted form of the same question is guarded: `served` (`src/Atoms.hs:542-544`) checks the channel and throws `AtomMute` with a message that names the missing key — "it asks phino to reduce an expression, while its stdin is closed, since its entry does not say 'serve'". The comment right above it states the contract for both forms: a program started for the fire has nothing to be answered over, so its question must fail the fire rather than hang it. `referenced` is the same question over the same closed channel, and it has no such check. Three runs against a six-line universe (`phino dataize --partial --atoms=…  --inside='Φ.foo'`) with an entry that omits `serve` show the gap: the quoted ask gives the correct `serve` error; the same ask by reference gives the misleading "the program quit without answering", never mentioning `serve` although the program did ask; and by reference from a program that waits for its answer — the realistic shape, a fire suspended on a queue — hangs both sides. Killing the program in the hanging case reports "failed with exit code -15", so phino had decided nothing, whereas the same kill in the quoted case prints the `serve` error immediately.

The smallest fix is to give `referenced` the guard `served` already has: `case channel of Closed -> throwIO (AtomMute …)` before reducing, with the same wording that names `serve`. Worth mentioning as the reason such a failure presents as a hang at all: the wait for a transient program is unbounded (`patiently`, `src/Atoms.hs:591`, `BS.hGetContents _output >> waitForProcess`), so a program that waits for phino while phino waits for it to quit hangs even once phino has decided the failure; bounding it the way `briefly` (`src/Atoms.hs:596`) bounds a resident program would turn that into a report. Found while porting the eo#8548 lowering engine to Java on this channel: a registry entry missing `serve` hung the build instead of failing it.

VasilevNStas:
Fixes #1184.

## Problem

A program whose entry does not say `serve` may still ask a question by reference (`{"id","of","attr","reduce"}`), and when it does, the whole run hangs forever with 0% CPU on both sides.

### Root cause

The `referenced` handler (`src/Atoms.hs:509`) reduces the attribute and writes the answer unconditionally, while the channel for a non-serve program is `Closed`: `said` hands bytes to `pushed Closed = hClose`, the resulting `IOError` is swallowed by the `unheard` handler, and phino goes back to `heard` reading a reply the program cannot send — because it is blocked waiting for the answer phino just dropped.

In contrast, `served` (line 537) already has the guard:

```haskell
case channel of
  Closed -> throwIO (AtomMute func (T.unpack raw) "it asks phino to reduce an expression, while its stdin is closed, since its entry does not say 'serve'")
  Open -> ...
```

`served` fails fast with a clear message naming `serve`. `referenced` had no such check and silently hung both processes.

Why this manifests as a hang: `patiently` waits unbounded (`waitForProcess`), so a program that waits for phino while phino waits for it to quit hangs even after phino decides the failure. The only way to surface the real error is to fail before writing.

## Fix

Give `referenced` the same `case channel` guard that `served` already has. Before any IO, check if the channel is `Closed`; if so, throw `AtomMute` with the same clear wording about `serve`. This matches the documented contract: a transient program has nothing to be answered over, so questions must fail immediately rather than hang.

## Changes

- **src/Atoms.hs** — `referenced` now checks `channel` first:
  - `Closed → throwIO (AtomMute ...)` with message naming `serve`
  - `Open → original logic unchanged`
- Renamed local parameter `name` to `attrName` in `referenced` to avoid shadowing the nested `attributeValue` function's `name` argument.

## Tests

- All 2440 examples pass (13 pre-existing failures unrelated to this change)
- `make hlint` — clean
- `make fourmolu` — clean

### DIFF
--- a/src/Atoms.hs
+++ b/src/Atoms.hs
@@ -506,35 +506,37 @@ asked func Running{..} form univ channel reduce = do
     -- what the attribute carries, as 'ask' does, or to hand the node over as
     -- it is (#1165).
     referenced :: Int -> Int -> T.Text -> Bool -> IO ()
-    referenced minted req name doReduce = do
-      spoken' <- describe
-      case spoken' of
-        Left failure -> throwIO (AtomMute func described failure)
-        Right value -> do
-          logDebug (printf "Atom '%s' asks phino for '%s' of request %d%s as question %d" (T.unpack func) (T.unpack name) req (if doReduce then ", reduced," else ", as it is," :: String) minted)
-          answer <- if doReduce then reduce value else pure value
-          said (lined (object ["id" .= minted, "𝑛" .= spelled answer]))
+    referenced minted req attrName doReduce = case channel of
+      Closed -> throwIO (AtomMute func described "it asks phino for an attribute of a previous request, while its stdin is closed, since its entry does not say 'serve'")
+      Open -> do
+        spoken' <- describe
+        case spoken' of
+          Left failure -> throwIO (AtomMute func described failure)
+          Right value -> do
+            logDebug (printf "Atom '%s' asks phino for '%s' of request %d%s as question %d" (T.unpack func) (T.unpack attrName) req (if doReduce then ", reduced," else ", as it is," :: String) minted)
+            answer <- if doReduce then reduce value else pure value
+            said (lined (object ["id" .= minted, "𝑛" .= spelled answer]))
       where
         described :: String
-        described = printf "{'of':%d,'attr':'%s'}" req (T.unpack name)
+        described = printf "{'of':%d,'attr':'%s'}" req (T.unpack attrName)
         describe :: IO (Either String Expression)
         describe = do
           forms <- readIORef _forms
           pure $ case IM.lookup req forms of
-            Nothing -> Left (printf "there is no in-flight request %d to take '%s' from" req (T.unpack name))
-            Just form' -> case attributeValue name form' of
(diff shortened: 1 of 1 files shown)


0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2576](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1191, opened 2026-09-11
Print a φ-bound formation without an empty @base

### LINKED ISSUE #1189: `--output=xmir` prints a formation bound to φ with an empty @base attribute (src/XMIR.hs:162)
The XMIR printer renders a φ binding through its own clause instead of `namedBinding`. `formationBinding (BiTau AtPhi expr)` (`src/XMIR.hs:160-162`) calls `expression expr ctx` and always emits `object [("name", "φ"), ("base", base)] children`, while the clauses for a label and for ρ right above it (`src/XMIR.hs:158-159`) go through `namedBinding`, whose first equation (`src/XMIR.hs:173`) renders a formation as `object [("name", name)]` with its bindings nested and no `@base` at all. For a formation `expression` answers an empty base (`src/XMIR.hs:103-105`), so a formation bound to φ comes out as `<o base="" name="φ">`.

`phino rewrite --output=xmir` on `⟦ φ ↦ ⟦ λ ⤍ S8 ⟧, ρ ↦ ∅ ⟧` prints `<o base="" name="φ"><o name="λ">S8</o></o>`, and `⟦ φ ↦ ⟦ x ↦ ∅ ⟧, ρ ↦ ∅ ⟧` prints `<o base="" name="φ"><o base="∅" name="x"/></o>`, whereas the same formation under any other name, `⟦ w ↦ ⟦ λ ⤍ S8 ⟧, ρ ↦ ∅ ⟧`, prints `<o name="w"><o name="λ">S8</o></o>` as it should. An empty `@base` is not a formation and not an application in XMIR, so the reader side of the schema treats the element as a broken reference, and every consumer of the output has to special-case φ (this is what drops the residual of a recursive fragment in objectionary/eo#8548, whose φ is exactly a bare marker formation).

The smallest fix is to let the φ clause do what its neighbours do: `formationBinding (BiTau AtPhi expr) ctx = Just <$> namedBinding (show AtPhi) expr ctx`, so a formation under φ nests its bindings and a non-formation keeps its `@base`.

yegor256:
The φ clause of `formationBinding` rendered its expression through `expression`, which answers an empty base for a formation, so a formation bound to φ came out as `<o base="" name="φ">`. It now goes through `namedBinding`, the same way the clauses for a label and for ρ right above it do: a formation nests its bindings and anything else keeps its `@base`.

An empty `@base` is neither a formation nor an application in XMIR, so consumers had to special-case φ, and `xmirToPhi` refused to read the document back. The round-trip works now.

Two tests come with it: one checks the printed φ element carries no `@base`, the other round-trips `[[ k -> [[ @ -> [[ L> S8 ]] ]] ]]` through XMIR and back.

Closes #1189

### DIFF
--- a/src/XMIR.hs
+++ b/src/XMIR.hs
@@ -157,9 +157,7 @@ expression expr _ = throwIO (UnsupportedExpression expr)
 formationBinding :: Binding -> XmirContext -> IO (Maybe Node)
 formationBinding (BiTau (AtLabel label) expr) ctx = Just <$> namedBinding (T.unpack label) expr ctx
 formationBinding (BiTau AtRho expr) ctx = Just <$> namedBinding (show AtRho) expr ctx
-formationBinding (BiTau AtPhi expr) ctx = do
-  (base, children) <- expression expr ctx
-  pure (Just (object [("name", show AtPhi), ("base", base)] children))
+formationBinding (BiTau AtPhi expr) ctx = Just <$> namedBinding (show AtPhi) expr ctx
 formationBinding (BiDelta bytes) _ = pure (Just (NodeContent (T.pack (printBytes bytes))))
 formationBinding (BiLambda (Function name)) _ = pure (Just (object [("name", show AtLambda)] [NodeContent name]))
 formationBinding (BiVoid AtRho) _ = pure Nothing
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -8 for the lack of code review; -4 for too few (14) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2687](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1193, opened 2026-09-11
Keep the deep walk from firing a λ the dispatch does not demand

### LINKED ISSUE #1187: `morph --deep` fires the λ of a dispatch target before reducing the dispatch, unlike the top-level run (src/Dataize.hs:341)
The deep walk enters a dispatch through its target before it ever asks 𝕄 about the dispatch itself. In `deepened`, the `parts` clause for `ExDispatch target attr` (`src/Dataize.hs:341-343`) calls `go context target`, and `go` (`src/Dataize.hs:330-333`) asks `fired` about that target once its own parts are walked. A target that 𝕄 reduces to a saturated formation with a served λ therefore fires, and only afterwards is the dispatch on the answer looked at. The top-level run does it the other way round: `morph` (`src/Dataize.hs:279-283`) reduces the whole `--inside` expression with 𝕄 first, where the `dot` rule resolves the dispatch, and the walk begins on that answer.

The same expression thus answers differently depending on where it stands. With `world.phi` holding `⟦ foo ↦ ⟦ f ↦ ⟦ a ↦ ∅, φ ↦ ξ.a, λ ⤍ L_box ⟧ ⟧ ⟧` and a served script for `L_box` that answers `⟦ Δ ⤍ FF- ⟧`, `phino morph --deep --partial --hide-rho --atoms=atoms.json --inside='Φ.foo.f( a ↦ ⟦ Δ ⤍ 01- ⟧ ).φ' world.phi` prints `⟦ Δ ⤍ 01- ⟧`, the φ of `f` as written. Wrapping the very same term in a binding, `--inside='⟦ φ ↦ Φ.foo.f( a ↦ ⟦ Δ ⤍ 01- ⟧ ).φ ⟧'`, prints `⟦ φ ↦ ⟦ Δ ⤍ FF- ⟧.φ ⟧`: the box fired although nothing dataized `f`, and the dispatch is left stuck on an answer that has no φ. Nested dispatches are the common case in any program `--deep` is meant for, so a formation that carries both a λ and the attribute being dispatched is unreachable through the walk (this is what forces objectionary/eo#8548 to morph one binding at a time and glue the residuals back by hand).

The smallest fix is to give the dispatch clause the order the top level already has: in `parts context (ExDispatch target attr)` reduce the dispatch as a whole with 𝕄 (the `reduced` of `fired` does exactly that) before descending into `target`, walk the answer when the `dot` rule applied, and enter the target only when the dispatch stayed as it was written.

yegor256:
The walk now carries the attribute a term stands under down into `fired`, and the λ of the formation 𝕄 reaches is fired only where that attribute is none of the formation's own — exactly the condition `ml` fires under, since `dot` resolves the dispatch before `ml` is ever reached. Everywhere else the walk fires what `mf` left bare, as before.

That kills the reported breakage: `⟦ φ ↦ Φ.foo.f( a ↦ ⟦ Δ ⤍ 01- ⟧ ).φ ⟧` no longer comes back as `⟦ φ ↦ ⟦ Δ ⤍ FF- ⟧.φ ⟧` with the box fired and the dispatch stuck on an answer that has no φ. The ξ-as-a-dispatch-target shape from your comment goes through the same path, so `>> name!` bindings of a boxed formation stop mangling too. The chains that legitimately fire still do: `5.plus( 6 ).plus( 7 )` reaches 18, because `plus` is absent from the plus-formation and `ml` does demand that λ.

One thing I did not do, and want your call on. I first tried the fix as written in the issue — reduce the dispatch with 𝕄 and stand the answer back in place — and it breaks 6 of the 7 `--deep` tests: `x -> 5.plus( 6 )` turns into ⊥ and `x -> 5.nope` turns into the whole `number` object inlined, which is what `deepened` and `fired` say in their docblocks must not happen ("where none fired it stays as it was written", "a normal form would spell the whole object out in place of the name the program called it by"). So the wrapped term above is now left as written rather than reduced to `⟦ Δ ⤍ 01- ⟧`. Making a nested term reduce exactly as the top level does is a bigger decision — it means the walk standing 𝕄 answers into the program — and I would rather you make it than have me sneak it in here.

Closes #1187

### DIFF
--- a/src/Dataize.hs
+++ b/src/Dataize.hs
@@ -319,18 +319,18 @@ morph universe ctx@DataizeContext{..} = do
 -- registry allows. Every entry is charged to the '--max-steps' budget, which
 -- is what bounds the walk.
 deepened :: Expression -> Expression -> State -> DataizeContext -> IO (Expression, State)
-deepened expr univ = go ExXi expr
+deepened expr univ = go Nothing ExXi expr
   where
     -- A term as it was written, together with what its free ξ stands for: the
     -- formation the walk entered it from, without the binding it came from,
     -- exactly the context the 'dot' rule hands a dispatched body. At the top
     -- there is no such formation, so ξ stands for itself and contextualization
     -- leaves the term alone.
-    go :: Expression -> Expression -> State -> DataizeContext -> IO (Expression, State)
-    go context term state' caller = do
+    go :: Maybe Attribute -> Expression -> Expression -> State -> DataizeContext -> IO (Expression, State)
+    go dispatched context term state' caller = do
       ctx' <- deeper caller
       (walked, walkedState) <- parts context term state' caller
-      answer <- fired (contextualize walked context) univ walkedState ctx'
+      answer <- fired dispatched (contextualize walked context) univ walkedState ctx'
       maybe (pure (walked, walkedState)) pure answer
     -- The parts of a term nothing fired on, walked one by one and put back
     -- where they were, so the term keeps the shape it was written in.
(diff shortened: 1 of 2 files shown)


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2699](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1198, opened 2026-09-14
Drop the Package λ when merge unites an object with a same-named package

### LINKED ISSUE #1197: `merge` keeps `λ ⤍ Package` on an object that is also a package, so `dataize` stops with "Atom Package does not exist" (src/Merge.hs:40)
When `merge --input=xmir` joins a document that defines an object with a document whose `+package` meta names that same object as a package, the object comes out carrying `λ ⤍ Package` beside its own bindings, and the result cannot be dataized. The marker is planted in `src/XMIR.hs:468`, where every segment of the package path is wrapped as `⟦ …, λ ⤍ Package, ρ ↦ ∅ ⟧`; `mergeBindings` in `src/Merge.hs:40` then unites the wrapper with the real object binding by binding and keeps the λ, since nothing on the other side collides with it.

Two documents show it: `bytes.xmir`, holding `[@] > bytes`, and `bytes/as-bytes.xmir`, holding `+package bytes` and `[^] > as-bytes` with `^ > @`. Merging them gives

```
⟦ bytes ↦ ⟦ φ ↦ ∅, as-bytes ↦ ⟦ ρ ↦ ∅, φ ↦ ξ.ρ ⟧, λ ⤍ Package, ρ ↦ ∅ ⟧, ρ ↦ ∅ ⟧
```

and `dataize --inside='Φ.bytes( φ ↦ ⟦ Δ ⤍ 01- ⟧ ).as-bytes'` on it stops with `Atom 'Package' does not exist`, because a λ takes precedence over φ and there is no such atom. `morph --partial` under `--atoms` shows the same thing from the other side: a `reduce` question about such an object comes back as the whole formation, λ included, instead of the bytes its φ decorates to. This is exactly the shape of `org.eolang.bytes` in objectionary/eo, which is an object and a package at once, so any program that reaches `as-bytes` through a `bytes` value cannot be reduced.

This is wrong because the `Package` marker exists only so that `getPackage` in `src/XMIR.hs:224` can tell a pure path segment apart when a program goes back to XMIR; an object that has its own φ and bindings is not a path segment, and a program that phino's own `merge` produces should be one phino's own `dataize` accepts.

The smallest fix is in `mergeBindings` (`src/Merge.hs:40`): when one side of a collision holds `λ ⤍ Package` and the other side holds a φ, Δ or any non-Package binding of its own, drop the `Package` λ from the united formation, so the object wins and its nested package members simply join it.

yegor256:
`mergeBindings` now strips `λ ⤍ Package` from a united formation when only one of the two sides carried it. When both sides carry it, both are pure path segments and the marker survives untouched.

The marker exists so `getPackage` can tell a path segment apart on the way back to XMIR. But an object that has its own φ and bindings is not a path segment, and `merge --input=xmir` was stapling the marker onto one anyway whenever a `+package` meta named an object that another document defines. `dataize` then stopped on it with `Atom 'Package' does not exist`, since a λ wins over φ and no such atom exists. That is exactly the shape of `org.eolang.bytes`, which is an object and a package at once, so nothing reaching `as-bytes` through a `bytes` value could be reduced.

One note on the fix in the issue: I did not key the decision on whether the other side holds a φ, Δ or any non-Package binding, because a legitimate package/package merge (`org.eolang.number` plus `org.eolang.text`) also puts a non-Package binding on each side, and that would have stripped the marker off `eolang` too. Asymmetry is the cleaner signal — every wrapper the XMIR reader builds carries the λ, so a side without it is a real object. Both cases are in `MergeSpec`.

Closes #1197


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (265 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3260](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1204, opened 2026-09-14
Stop the deep walk firing a λ whose argument came back as ⊥

### LINKED ISSUE #1196: `morph --deep` enters an unapplied formation holding `ρ ↦ ∅`, turns its `ξ.ρ` into ⊥ and fires a λ with ⊥ (src/Dataize.hs:355)
The deep walk in `src/Dataize.hs:355` (`bindings`, inside `deepened`) enters the bindings of every formation it meets, including a formation that is still sitting unapplied inside its parent and therefore still holds `ρ ↦ ∅`. Each body is walked in the `scope` of that formation (`src/Dataize.hs:367`) and contextualized, so a `ξ.ρ` in it becomes `⟦ …, ρ ↦ ∅, … ⟧.ρ`, which the `null` rule turns into `⊥`. When the body applies a nested formation to that `ξ.ρ`, the application ends up as `b( k ↦ ⊥ )`: every void is now filled, `saturated` at `src/Dataize.hs:153` is satisfied, and `fired` (`src/Dataize.hs:397`) fires the λ with `⊥` as its argument.

A minimal program:

```
⟦
  foo ↦ ⟦
    x ↦ ∅,
    a ↦ ⟦
      ρ ↦ ∅,
      b ↦ ⟦ ρ ↦ ∅, k ↦ ∅, φ ↦ ξ.k, λ ⤍ L_b ⟧,
      φ ↦ ξ.b( α0 ↦ ξ.ρ )
    ⟧,
    ρ ↦ ∅
  ⟧,
  ρ ↦ ∅
⟧
```

With `morph --deep --partial --atoms=<registry serving L_b> --inside='Φ.foo( x ↦ ⟦ Δ ⤍ 02- ⟧ ).a.φ'`, the entry itself reduces correctly to `b( ρ ↦ foo( x ↦ 02- ).a, k ↦ foo( x ↦ 02- ) )`, but the walk then descends into the argument `k ↦ foo( x ↦ 02- )`, meets the unapplied copy of `a` there, walks its `φ` in its own scope and fires `L_b` first with `𝑏 = ⟦ k ↦ ⊥, φ ↦ ξ.k ⟧`, and only then with the real `k`. The answer of that first firing is then stood into the program in place of the `φ` of the unapplied `a`, so any later dispatch of `a` is broken too. An atom that dataizes its operand, as the atoms of objectionary/eo do, meets the `⊥` and the whole run dies with "dataization reached the terminator ⊥", so a program that nests one private formation inside another cannot be morphed at all. Without `--deep`, or when the same program is entered through `foo( x ↦ 02- ).a` rather than `.φ`, no such firing happens.

This is wrong because a formation with a void `ρ` is not attached to anything yet: its `ξ.ρ` is not an error, it is an input the next dispatch will supply, exactly like the other voids `saturated` already respects. Reducing its bindings in a scope where `ρ` is `∅` cannot produce anything meaningful.

The smallest fix is to make `bindings` leave a formation alone when it still holds a void binding, `ρ` included, the same way `saturated` refuses to fire it: check `whole` for a `BiVoid` before entering its `BiTau` bodies and return the bindings as written when one is found.

yegor256:
A binding standing at ⊥ now counts as unfilled in `filled`, the same way a void does, so `saturated` refuses to hand such a formation to an atom. `filled` moves out of `saturated` to the top level, since both now want it.

The deep walk reduces every body in the scope of the formation around it, and a formation sitting unapplied inside its parent still holds `ρ ↦ ∅`. So a `ξ.ρ` in that body came back as ⊥, that ⊥ filled the last void of a nested λ-formation, and the atom fired with ⊥ for an argument — with its answer then stood back into the program in place of what the unapplied formation had written there. An atom that dataizes its operand, as the ones in objectionary/eo do, died on the ⊥ and took the run with it.

One note on the fix in the issue. I did not make `bindings` skip a formation that still holds a void: phino's own sugar puts `ρ ↦ ∅` on nearly every formation it parses, so that guard switches `--deep` off for almost any program — it breaks 8 existing cases, including `5.plus( 6 )` at the top of a program. The guard belongs where the issue's own reasoning points, next to the voids `saturated` already respects, and there it costs nothing: all 2454 cases pass.

While I was in there, the cases of the deep walk moved out of `DataizeSpec` and into `test-resources/morph-deep-packs`, one pack each, naming the program under `input`, the entry under `location` and the answer under `result` or `fails`. The one case that needs a bespoke resident atom (#1187) stays in Haskell.

Closes #1196


yegor256:
@yegor256 mind taking a look when you get a chance? The fix sits in a different spot than the issue suggested, and I can walk through why if that reads oddly.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (292 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3264](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1208, opened 2026-09-14
Tell a program what an answer carries, under Δ, λ and ∅

### LINKED ISSUE #1206: By-reference answers spell the node, so a program has to parse 𝜑 to tell data from a stuck atom (src/Atoms.hs:518)
`referenced` in `src/Atoms.hs:517-518` answers a question by reference with the node spelled under `𝑛`, whether it was reduced or handed over as written, and `served` in `src/Atoms.hs:551` does the same for `ask`. A program that wants the value of an operand therefore receives text such as `⟦ Δ ⤍ 40-14-… ⟧`, `Φ.bytes( φ ↦ ⟦ Δ ⤍ … ⟧ )`, `⟦ λ ⤍ S4 ⟧` or `Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ S4 ⟧ ) )`, and has to recognize these spellings itself before it can tell data from a stuck atom. A question about a void attribute is worse: `attributeValue` in `src/Atoms.hs:530-538` walks only the `BiTau` bindings, so `x ↦ ∅` is reported as "carries no attribute" and fails the fire, which means a program cannot even ask whether an operand is bound and has to read it off `𝑏` instead.

This is wrong because the shape of an answer is phino's knowledge, not the program's. The README promises that on this wire "neither side ever re-prints or re-parses" a receiver, yet every program that serves atoms ends up keeping a 𝜑 reader of its own, one that breaks whenever the printer changes a detail, as `as-bytes` against `φ` or a hidden `ρ ↦ ∅` did. In objectionary/eo that reader is a set of regular expressions over `𝑏` and over `𝑛`, and it is the one part of the engine that knows syntax at all.

The smallest fix is to say, next to `𝑛`, what the answered node carries: when the answer is an `ExFormation` with a `BiDelta`, add `"Δ": "40-14-…"`; when it has a `BiLambda`, add `"λ": "S4"`; and when `attributeValue` meets a `BiVoid` of the asked name, answer `{"id": 7, "∅": true}` instead of failing. A program then reads facts off the JSON and never touches the text.

yegor256:
When a program asks phino a question over the atom channel, the answer now
says what the node under `𝑛` actually is. A formation with a Δ binding
carries its byte array under `Δ`, one with a λ binding carries the name of
the function it is stuck on under `λ`, and an attribute bound to nothing at
all is answered with `{"id": 7, "∅": true}` instead of failing the fire.

Until now the node was spelled out and nothing else, so every program that
serves atoms kept a 𝜑 reader of its own — in objectionary/eo a pile of
regular expressions over `𝑏` and `𝑛` — just to tell `⟦ Δ ⤍ … ⟧` from
`⟦ λ ⤍ S4 ⟧`, and it broke whenever the printer changed a detail. The shape
of an answer is phino's knowledge, so phino says it. A void attribute used
to be reported as "carries no attribute" too, because the lookup walked only
the τ bindings, which left a program no way to ask whether an operand is
bound.

One thing to watch: an answer may now carry a `λ` of its own, so a line of
phino's is a request when it carries `𝑏` and an answer when it does not. The
README example and the two atom fixtures branch on `𝑏` now; `asking.js` reads
its bytes off `Δ` and parses no 𝜑 at all.

Closes #1206


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (149 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3126](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1209, opened 2026-09-14
Let a by-reference question take a dotted path

### LINKED ISSUE #1207: A by-reference question cannot reach into the attribute it names, such as `ρ.length` (src/Atoms.hs:530)
`attributeValue` in `src/Atoms.hs:530-539` resolves `attr` as one name among the top-level bindings of the receiver phino holds for the request. A program that needs an attribute of that attribute has no way to ask for it: the `length` of a tuple that arrived as `ρ`, or the `φ` of a carrier, sit one level down, and the only other question, `ask`, would have to quote the receiver, which is exactly what the lean wire of #1165 set out to avoid. In objectionary/eo the engine reaches `length ↦ … ⟦ λ ⤍ S4 ⟧` by running a regular expression over the spelled `ρ` it asked for as written.

This is wrong for the reason the question by reference exists at all: phino still holds the receiver, so it can take the value from it directly, and depth is the only thing the question is missing. Making a program re-parse the answer to go one level deeper defeats the point of asking by reference.

The smallest fix is to let `attr` be a dotted path, such as `ρ.length`: split it on `.` in `describe`, fold `attributeValue` over the segments, and fail with the same "carries no attribute" message when a segment is missing, before `reduce` is applied to the leaf as it is today.

yegor256:
The `attr` of a by-reference question is now a path down the receiver, split
on the dot: `{"of": 1, "attr": "ρ.length", "reduce": true}` walks to `ρ`,
then to `length`, and `reduce` applies to the node the path ends at. A
segment the formation does not carry, or one that runs into a void
attribute, fails the fire with the message a missing `attr` has always
failed with.

Until now `attr` resolved a single name among the top-level bindings, so a
program that wanted an attribute of that attribute had nowhere to go: the
`length` of a tuple that arrived as `ρ` sits one level down, and the other
question, `ask`, would have to quote the receiver — the very thing the lean
wire of #1165 exists to avoid. In objectionary/eo that gap is filled by a
regular expression over the spelled `ρ`. phino holds the whole receiver
anyway, so depth was the only thing the question was missing.

The dot is safe to split on, by the way: no attribute of 𝜑-calculus carries
one in its own name — the label parser rejects it.

Closes #1207


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3134](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1211, opened 2026-09-14
Say under Φ. the forma an answer is dispatched by

### LINKED ISSUE #1210: A by-reference answer says nothing about an application as written, so a typed literal loses its forma (src/Atoms.hs:593)
Since #1206, `answered` in `src/Atoms.hs:593` spells next to `𝑛` what a formation carries, its `Δ` and its `λ`, but `carried` has one clause only, for `ExFormation`, so an answer about an application says nothing beyond `𝑛`. A program asking for `x` of a receiver holding `x ↦ Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ 40-08-00-00-00-00-00-00 ⟧ ) )` as it is gets the node alone, and asking reduced gets `Δ` alone, since dataization takes the number apart down to its bytes.

The name `number` is the only place the forma of that literal lives, since 𝜑-calculus types nothing nominally, and the program cannot learn it without reading `𝑛`, which #1206 set out to spare it. The eo lowering engine feeds such literals to the arms of an `if` and to the untyped voids of boxed formations, and once it stops parsing 𝜑 it has no way to tell `-1` from the bytes it is made of, so the Java it emits returns a byte array where a number was promised.

A small addition to `carried` would do: when the answer, stripped of the applications around it, is a chain of dispatches from `Φ`, spell that chain under one more key, say `"Φ.": "number"` for `Φ.number( φ ↦ … )`, `"Φ.": "true"` for `Φ.true` and `"Φ.": "tuple"` for `Φ.tuple( length ↦ …, head ↦ …, tail ↦ … )`, and stay silent for anything else, such as `Φ.number( … ).plus( … )`, whose chain has an application inside it.

yegor256:
An answer that is an application now spells, next to `𝑛`, the chain it is
dispatched off Φ by: `"Φ.": "number"` for `Φ.number( φ ↦ … )`, `"Φ.": "true"`
for `Φ.true`, `"Φ.": "org.eolang.number"` for a chain spelled in full. A chain
with an application inside it, such as `Φ.number( … ).plus( … )`, dispatches
off a term phino would have to dataize before it could name it, so it names no
forma and nothing is said.

`carried` had one clause, for formations, so everything #1206 added stopped at
the `Δ` and the `λ` of a formation and an application got the node alone.
𝜑-calculus types nothing nominally, so the name a literal is dispatched by is
the only place its forma lives — and dataizing the operand does not help,
since `Δ` is all that survives a number taken apart. Reading the forma meant
parsing `𝑛`, which is what #1206 set out to spare a program.

One judgement call worth a look: I let the chain be more than one segment, so
`Φ.org.eolang.number( … )` says `org.eolang.number` rather than going silent.
The issue only spells out the one-segment cases, and it seemed wrong to stay
quiet about a forma that is simply written in full.

Closes #1210


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain anything, in particular the multi-segment chain choice I flagged in the description.

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3142](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1213, opened 2026-09-14
Walk a by-reference path into an application

### LINKED ISSUE #1212: A by-reference path stops at an application although the attribute is bound right there (src/Atoms.hs:576)
A dotted `attr` walks only formations: `attributeValue` in `src/Atoms.hs` around line 560 matches `ExFormation` and every other expression falls to `attributeValue _ _ = Nothing` at line 576, so the fire dies with "carries no attribute". An application binds attributes too, though. With a receiver of `⟦ x ↦ Φ.bool( if ↦ ⟦ guard ↦ ⟦ λ ⤍ S1 ⟧ ⟧ ) ⟧`, the question `{"id": 7, "of": 1, "attr": "x.if.guard", "reduce": false}` fails on 0.0.130 with "the receiver of request 1 carries no attribute 'x.if.guard'", while `if` is bound right there, in the argument, and `guard` under it.

This matters for a program that filled a void with a marker it built itself and wants the marker back as written. Reducing `x` instead dataizes `Φ.bool` down to its `φ`, which fires the `λ` inside the marker and hands back whatever that produces, not the marker; asking as written is the only way to read it, and the path is the only way to reach inside it without parsing `𝑛`, which the answers of #1206 and #1207 were made to spare the program. The README says every segment but the last has to name a formation, so this is a limit the docs own up to, but a `τ ↦ e` argument is as much a binding as a `τ ↦ e` inside a formation.

The smallest change is one more clause before the fall-through: `attributeValue name (ExApplication expr (ArTau attr value))` answers `Bound value` when `attr` prints as `name` and recurses into `expr` otherwise, and `attributeValue name (ExApplication expr _)` recurses into `expr` past a positional argument. The README paragraph on paths would then say a segment names a formation or an application.

yegor256:
A dotted `attr` now walks applications as well as formations. With a receiver
of `⟦ x ↦ Φ.bool( if ↦ ⟦ guard ↦ ⟦ λ ⤍ S1 ⟧ ⟧ ) ⟧`, the question
`{"of": 1, "attr": "x.if.guard"}` reaches the marker instead of dying with
"carries no attribute". An argument binds an attribute the way a τ binding
does, and it binds it from the outside, so it wins over the void it fills; a
positional argument names nothing, so the walk goes past it into whatever the
application applies to.

`attributeValue` matched `ExFormation` and let everything else fall through to
`Nothing`, which is a real gap for a program that filled a void with a marker
it built itself: reducing `x` dataizes `Φ.bool` down to its `φ` and fires the
λ inside the marker, so asking as written is the only way to read it back, and
the path is the only way in without parsing `𝑛` — the very thing #1206 and
#1207 exist to spare a program.

Four cases in the spec cover it, including the precedence one
(`⟦ y ↦ ∅ ⟧( y ↦ … )` answers the argument, not `∅`). The README paragraph on
paths says formation or application now.

Closes #1212


0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3260](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1214, opened 2026-09-14
Keep one rule per file in every rule set

### LINKED ISSUE #1200: Inconsistency in Rule Storage: Monolithic Files vs Individual Files in `resources/`
`resources/normalize/` keeps one rule per file — 14 of them, embedded wholesale
with `embedDir` at `src/Yaml.hs:312`. The other three rule sets sit next to that
directory as monolithic files: `resources/morphing.yaml` (12 rules),
`resources/dataization.yaml` (5) and `resources/contextualization.yaml` (7),
each pulled in by its own `embedFile` at `src/Yaml.hs:504-512`.

This is the arrangement #36 already rejected: one rule per file, no composite
YAML. The three files predate nothing — they arrived with #760 and simply never
followed that decision. The cost is visible in the tests: `RewriterSpec.hs:134`
maps a test pack one-to-one onto `resources/normalize/<name>.yaml`, and no such
mapping exists for a morphing or dataization rule, because there is no file to
map onto.

Give each of the three its own directory — `resources/morphing/`,
`resources/dataization/`, `resources/contextualization/` — with one YAML per
rule, and load them with `embedDir` as the normalization rules already are. Two
details to settle while splitting: `morphing.yaml` and `contextualization.yaml`
declare "applied top-to-bottom, first matching clause wins", so the order needs
to be spelled out in the filenames (`01-mf.yaml`) rather than left to the
directory listing — `dataization.yaml` notes its guard already frees it from
order; and the header docblock each file carries, explaining the judgment as a
whole, needs a home, a `README.md` in each directory being the obvious one.
`extra-source-files` at `phino.cabal:15` needs the new paths too.

yegor256:
Morphing, dataization and contextualization each kept every one of their
clauses in a single composite YAML, while `resources/normalize` had been one
rule per file all along. This gives the other three the same treatment:
`resources/morphing/`, `resources/dataization/` and
`resources/contextualization/`, each a directory of single-rule files named
after the rule they carry, all loaded with `embedDir` the way the
normalization rules already were. `extra-source-files` in the cabal file
follows the new paths.

The point is that a rule can now be pointed at, diffed and tested on its own,
the way `RewriterSpec` already maps a test pack onto
`resources/normalize/<name>.yaml`. A new spec in `YamlSpec` holds the layout
in place: every file in a rule directory has to be named after the rule it
carries, and nothing else may be in there.

Two things worth a reviewer's attention. The clauses of each judgment are
disjoint, so nothing orders them and no file name spells out an order; the
consequence is that `explain` now lists the rules in the order `embedDir`
sorts paths in, which is alphabetical, and the three `explain` expectations in
`CLISpec` were reordered to match. And the header docblocks the composite
files carried are gone along with them, as agreed — they explained each
judgment as a whole and had no single rule file to live in.

Closes #1200


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain anything that looks off, especially the reordered `explain` expectations.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (866 >= 100); -8 for way too many hits-of-code (866 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3264](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1215, opened 2026-09-14
Give the morphing relation its own module

### LINKED ISSUE #1203: src/Dataize.hs holds both the M and D relations, so there is no Morph.hs and no MorphSpec.hs
`src/Dataize.hs` carries both formal functions of the calculus: Morphing (`morph` at src/Dataize.hs:278, `morph'` at src/Dataize.hs:213, with the `deepened` and `fired` helpers at :321 and :397) and Dataization (`dataize` at src/Dataize.hs:452, `dataize'` at :485). The module exports both, runs to 775 lines, and backs two separate CLI commands. There is no `Morph.hs`.

The two relations are disjoint: separate rule sets (PRIM, NMZ, LAMBDA, PHI against DELTA, BOX, NORM) loaded from separate YAML files. One module holding both is an SRP violation, and it leaks into the tests — `test/DataizeSpec.hs` is 887 lines covering both relations, and since test files map 1:1 with source modules, the `morph` command has no `MorphSpec.hs` of its own to live in.

Extract `Morph.hs` holding `morph`, `morph'`, `deepened` and `fired`, leaving `Dataize.hs` to import it; `test/MorphSpec.hs` then follows from the 1:1 rule. It is not a straight lift, though: `dataize'` calls `morph'` at src/Dataize.hs:535, and the morph side reaches back the other way through `_evaluate` -> `atom` (:707) -> `reduction` (:689) -> `dataize`, so that back-edge wants the same injected-function treatment `Deps.hs` already gives `_buildTerm`.

yegor256:
`Dataize.hs` carried both formal functions of the calculus, 775 lines backing two
separate commands, and `DataizeSpec.hs` covered both of them, which is why the
`morph` command had no spec file of its own. Now `Morph.hs` holds 𝕄 and
everything both judgments read — the context, the step budget, the signals a
stuck run raises, the premise plumbing — and `Dataize.hs` keeps 𝔻 alone and
imports it. `MorphSpec.hs` is new and takes the morphing tests over.

Only one edge pointed back the wrong way: an atom asking phino to reduce an
operand of its own is a whole run of 𝔻, and 𝕄 has no business knowing about
that. So it is injected into the context as `_reduce`, the same trick `Deps.hs`
already plays with `_buildTerm` — `Dataize` supplies its own `reduction` and
nothing in `Morph` imports downward.

Two renames fall out of the split: `DataizeContext` becomes `ReduceContext` and
`DataizeException` becomes `ReduceException`, since neither belongs to
dataization any more. The context fixture, `withAtoms` and the primitives
universe moved to `Fixtures.hs`, where both spec files reach them without
duplication. Behaviour is unchanged — the suite runs the same 2475 examples it
ran before.

Closes #1203


0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (2150 >= 100); -8 for way too many hits-of-code (2150 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3268](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1216, opened 2026-09-14
Keep the reduction cases in YAML packs

### LINKED ISSUE #1201: Dataization cases are hardcoded in test/DataizeSpec.hs instead of a test-resources/dataization-packs/ YAML suite
Every other feature of `phino` keeps its cases in a YAML pack suite under `test-resources/`: `rewriter-packs`, `parser-packs`, `condition-packs`, `filter-packs`, `xmir-parsing-packs`, `xmir-printing-packs`, `yaml-packs`. Dataization — arguably the most intricate part of the tool — has none. All of its cases sit hardcoded inside `test/DataizeSpec.hs`, an 887-line module, as Haskell tuple tables: `testMorph` at `test/DataizeSpec.hs:179`, `testDeep` at `test/DataizeSpec.hs:219`, `testDataize` at `test/DataizeSpec.hs:769`. The 𝜑-expressions inside them are stitched together with `unlines [ "[[", "  foo -> [[", ... ]]`.

That shape hides the thing under test. A dataization case is four plain values — the universe, the locator, the λ registry, the expected bytes — and none of them is readable through a wall of quoted string fragments and escaped brackets. Adding a case means editing Haskell and recompiling the suite, so the barrier to covering a new reduction path is higher than it is for any other feature. The λ fixtures the cases need already live outside the binary as `test-resources/atoms/*.js`, so the input side is half-externalized already while the expressions stay locked in the source.

Add `test-resources/dataization-packs/*.yaml`, each pack carrying `input`, `locator`, optional `atoms`, and `result` (or `morphed` for a 𝕄 case), and load them in `DataizeSpec.hs` the way `test/RewriterSpec.hs:104` loads its own — `Files.allPathsIn` over the directory, then `Yaml.decodeFileThrow` per file. Then move the `testMorph`, `testDeep` and `testDataize` tables out of the module and into packs, leaving `DataizeSpec.hs` with only the cases that genuinely need Haskell, such as the context and step-budget assertions.

yegor256:
The dataization cases lived in `DataizeSpec.hs` as Haskell tuple tables, with
the 𝜑 expressions stitched together out of quoted string fragments. They are
packs now: `test-resources/dataization-packs` holds 25 of them, each naming the
program, where the run enters it, whether the fixture λ functions answer it, and
what it must dataize to — or the failure it must raise. 𝕄 got the same
treatment in `test-resources/morph-packs`, and the pack loader the deep walk
already had serves both of its directories, with the directory saying whether
the walk is deep.

The point is that a case should be readable and cheap to add. Every other
feature of phino keeps its cases under `test-resources`, and dataization —
probably the most intricate part of the tool — had none, so covering a new
reduction path meant editing Haskell and recompiling the suite.

What stays in Haskell is what needs it: the raw metas that reach the
unmatched-term fallbacks, the explicit contexts behind the step and depth
budgets, and everything that asserts on the derivation chain or on the reports
𝔼 makes. Nothing changed about what runs — the suite is the same 2475 examples
it was before.

Closes #1201


yegor256:
@yegor256 could you take a look when you get a chance? Happy to explain any of the pack fields or move more cases out if you want them externalized too.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (552 >= 100); -8 for way too many hits-of-code (552 >= 400); -8 for the lack of code review; +8 to give you at least something. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3272](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1221, opened 2026-09-15
Bind the ξ of a node a by-reference question reduces

### LINKED ISSUE #1220: A by-reference question with `reduce: true` loses the ξ binder of the value it reads (src/Atoms.hs:542)
When a served program asks a by-reference question with `reduce: true`, `referenced`
at `src/Atoms.hs:542` hands the bare expression bound at the named attribute to
`reduce`. That is `reduction` at `src/Dataize.hs:186`, which splices the expression
into the universe as a fresh synthetic τ binding (`insideUniverse`,
`src/Morph.hs:557`) and dataizes it there. Every `ξ` inside the expression then
binds to the universe formation instead of the formation the expression was taken
from, and the reduction walks off into ⊥.

The body of a formation is exactly such an expression, so no program can ask phino
to reduce the body of the object it is firing on. With the universe
`⟦ book ↦ ⟦ page ↦ ⟦ a ↦ ∅, φ ↦ ξ.a, λ ⤍ L_box ⟧ ⟧, top ↦ ⟦ one ↦ Φ.book.page( a ↦ ⟦ Δ ⤍ 01-02 ⟧ ) ⟧ ⟧`
and a served program that answers the `L_box` fire with
`{"id": 9, "of": 1, "attr": "φ", "reduce": true}`, `phino morph --deep --inside='Φ.top'`
dies with "dataization reached the terminator ⊥", while the same question with
`attr: "a"`, whose value carries no `ξ`, answers
`{"id":9,"Δ":"01-02","𝑛":"⟦ Δ ⤍ 01-02 ⟧"}` (phino 0.0.132). Quoting the body back
with `ask` is no way around it, since `𝑏` is spelled without the ρ chain (#1165)
and a body reading `ξ.ρ` cannot be reconstructed from the request.

The fix is to reduce the attribute in the scope of its own receiver. `referenced`
already holds that receiver in `_forms`, with its λ binding removed so that it may
be dispatched on (`src/Morph.hs:567`), so handing `reduce` a dispatch of the named
path off the held formation, instead of the bare value, keeps `ξ` bound where it
belongs and changes nothing on the wire.

yegor256:
A by-reference question with `reduce` used to hand the reduction the bare node
bound at the attribute, and the reduction binds whatever it is given to a
synthetic attribute of the universe. Any `ξ` inside that node then stood for the
universe rather than for the formation the node came out of. The node now leaves
its formation contextualized, the way the `dot` rule takes it out, so `ξ` keeps
meaning what it meant where it was written.

This matters most for the one node a program really wants: the body of the
formation it is firing on. Asking for `φ` reduced was the natural way to let
phino evaluate a body while the atom stayed in charge of it, and it could not
work — every real body says `ξ.a` or `ξ.ρ` somewhere, so the run died at ⊥ with
nothing to go on. Quoting the body back with `ask` was no way around it either,
since `𝑏` carries no ρ chain since #1165.

A question that does not reduce is untouched: it still gets the node exactly as
written, `ξ` and all, because binding that `ξ` is what reducing does. Three cases
cover it — one in `AtomsSpec` on what reaches the reduction, one on the
as-written answer staying as written, and one in `CLISpec` that dataizes a real
formation whose body is `ξ.a` through a served program.

Closes #1220


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any part of it if something reads oddly.

0crat:
@yegor256 Thanks for the contribution! You've earned +8 points for this: +16 as a basis; -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3280](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1225, opened 2026-09-15
fix(#1199): split cli test resources into expressions and rules directories

### LINKED ISSUE #1199: `test-resources/cli/` Directory Doesn't Distinguish Between .phi Expressions and .yaml Rule Files
The `test-resources/cli/` directory holds nineteen files of two unrelated kinds side by side: eight `.phi` files that are 𝜑-expressions fed to the CLI as input (`foo.phi`, `bytes.phi`, `normalize.phi`, …) and eleven `.yaml` files that are rewriting rules passed through `--rule` (`simple.yaml`, `swap-a.yaml`, `raising-condition.yaml`, …). Nothing in the layout separates them, so an alphabetical listing interleaves the two, and a reader has to open a file to learn what it is for.

The split already exists in the test code, which makes the flat directory the odd one out. `test/CLISpec.hs:163` defines `resource` and `rule` as two helpers over the same path, and every one of the ten `resource` call sites names a `.phi` file while all thirty-two `rule` call sites name a `.yaml` file. The directory is the only place where the distinction is dropped.

Move the expressions into `test-resources/cli/expressions/` and the rules into `test-resources/cli/rules/`, then point `resource` at the first and let `rule` prepend the second, so each helper carries its own directory instead of sharing one. `test/CLIHelpersSpec.hs:56` spells `test-resources/cli/simple.yaml` inline and needs the same update.

Thayorns:
The `test-resources/cli/` directory kept two unrelated kinds of fixtures side by side: 𝜑-expressions fed to the CLI as input and rewriting rules passed through `--rule`. The eight `.phi` files now live in `test-resources/cli/expressions/` and the eleven `.yaml` files in `test-resources/cli/rules/`. Their content is untouched.

In `test/CLISpec.hs` the `resource` helper points at the expressions directory, and `rule` prepends the rules directory itself instead of going through `resource`, so each helper owns its own location. Three tests that spelled a path inline now go through these helpers, and `test/CLIHelpersSpec.hs` refers to `simple.yaml` at its new place.

Closes #1199


Thayorns:
@yegor256 this splits the cli fixtures into expressions and rules as #1199 describes, happy to clarify anything about the new layout.

0crat:
@thayorns Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (12) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3936](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1228, opened 2026-09-15
Answer λ functions from a YAML file instead of spawning scripts

### LINKED ISSUE #1226: Symbolic morphing: `--symbolic=atoms.yaml` and a tree-shaped `--protocol`
PR #1228 replaced `--atoms` with `--functions`: a YAML file of λ-function rules
consulted in `_evaluate` (`src/Morph.hs`), with `dataize` and `morph` premises,
`when` guards, `where` extras and a `symbols` block minting fresh λ names. That
shipped, and `src/Atoms.hs` and the process channel are gone. The proposal below
replaces the rest of it: the shape of an entry, the option that names the file
and, above all, the protocol.

Two things are wrong with what shipped.

The protocol `--evaluations` writes is a flat list of JSON objects. It records
the straight-line flow of data and almost nothing of what a symbolic run
decided: which entry of a λ answered, what the firing answered with, which
λ functions fired while an operand was being reduced, and where an unknown came
from. Here is the whole protocol of the fixture that forks on an undecided
comparison — three entries of `L_fork` were tried, two declined, and the file
says not one word about either attempt:

```
{"λ":"L_bytes_eq","𝑓0":"S_1"}
{"λ":"S_1","stuck":true}
{"λ":"L_fork","morph":"𝑛1","at":"begin"}
{"λ":"L_fork","morph":"𝑛1","at":"end"}
{"λ":"L_fork","morph":"𝑛2","at":"begin"}
{"λ":"L_fork","morph":"𝑛2","at":"end"}
{"λ":"L_fork","𝑓1":"S_2"}
```

And an entry computes. With `when` and `where` it can guard on data and call a
build-term function, so `5.plus(6)` really does come down to `11`. That is not
what these atoms are for. Their job is symbolic morphing: an atom should answer
a term carrying a fresh symbol, never a value it worked out.

## Proposal

Rename the option to `--symbolic=atoms.yaml` — it says what the run is, not what
the file holds — and replace `--evaluations` with `--protocol`, rendered as a
tree of firings rather than a line of JSON per firing.

An entry loses `when`, `where` and `symbols`, and keeps four keys:

- `λ` — the name, a regular expression as today, now **unique**: guards were the
  only thing telling two entries of one name apart, so a second entry under the
  same key is unreachable and the lookup answers one entry or none.
- `dataize` — operands brought down through 𝔻, each binding a bytes meta `δ`.
- `evaluate` — operands reduced, each binding an expression meta `𝑛`. This is
  the old `morph` block under a name that matches the `E(…)` of the protocol.
- `𝑛` — the answer.

Every entry numbers its own metas from `δ1` and `𝑛1`; the entry they belong to is
what tells two `δ1` apart, and the protocol names that entry on every line.

`𝜎` becomes a meta of the calculus beside `𝑛`, `δ` and `𝑓`. In a term, `𝜎1` is a
concrete symbol — a λ nothing answers, which is what makes the value it stands
for unknown. In an answer, a bare `𝜎` mints a fresh one, counted in the state 𝑠,
which is what `symbols: [𝑓0]` used to do. Minting starts after the symbols the
program already carries. An answer may mention `𝜎` and nothing else: no `δ` it
bound, since nothing computes from data any more.

`𝑓` stays. Three built-in rules match on it to find a λ binding in order to fire
it — `resources/morphing/ml.yaml:6`, `resources/normalize/dl.yaml:5`,
`resources/dataization/fire.yaml:5` — and it means a different thing than `𝜎`:
`𝑓0` stands for any λ name, `L_number_plus` included, while `𝜎` stands for a
symbol. It simply stops appearing in `atoms.yaml`.

Dataizing a symbol never gets stuck. It answers a fixed datum, 42
(`40-45-00-00-00-00-00-00`), the same one for every symbol, so 𝔻 always answers,
a `δ` always holds concrete data and no premise ever declines. The protocol
writes `D(𝜎1)` in place of that 42, so a reader knows the value was manufactured
rather than read out of the program. Nothing else in the engine moves.

## The example

`atoms.yaml`:

```yaml
- λ: L_number_plus
  dataize:
    δ1: $.ρ
    δ2: $.x
  𝑛: Φ.number( φ ↦ Φ.bytes( φ ↦ [[ L> 𝜎 ]] ) )
- λ: L_number_gt
  dataize:
    δ1: $.ρ
    δ2: $.x
  𝑛: Φ.bool( if ↦ [[ L> L_fork, left ↦ ?, right ↦ ?, φ ↦ [[ L> 𝜎 ]] ]] )
- λ: L_fork
  dataize:
    δ1: $.φ
  evaluate:
    𝑛1: $.left
    𝑛2: $.right
  𝑛: 𝑛1
```

The program:

```
[[
  bytes -> [[ φ -> ? ]],
  bool -> [[ if -> ? ]],
  number -> [[
    φ -> ?,
    plus -> [[ x -> ?, L> L_number_plus ]],
    gt -> [[ x -> ?, L> L_number_gt ]]
  ]],
  foo -> [[ x -> ?, φ -> x.gt(0).if(x.plus(x.plus(1)), x.plus(x)).plus(5) ]],
  @ -> [[
    a -> Φ.foo( Φ.number( φ ↦ Φ.bytes( φ ↦ [[ L> 𝜎1 ]] ) ) )
  ]]
]]
```

Morphed deeply at `Q.@.a`, the protocol:

```
M(Q.@.a)
  E(L_number_gt)
    δ1.1 := D(𝜎1)
    δ2.1 := 00-00-00-00-00-00-00-00
    𝑛.1 := Φ.bool( if ↦ [[ L> L_fork, left ↦ ?, right ↦ ?, φ ↦ [[ L> 𝜎2 ]] ]] )
  E(L_number_plus)
    δ1.1 := D(𝜎1)
    δ2.1 := 3F-F0-00-00-00-00-00-00
    𝑛.2 := Φ.number( φ ↦ Φ.bytes( φ ↦ [[ L> 𝜎3 ]] ) )
  E(L_number_plus)
    δ1.2 := D(𝜎1)
    δ2.2 := D(𝜎3)
    𝑛.3 := Φ.number( φ ↦ Φ.bytes( φ ↦ [[ L> 𝜎4 ]] ) )
  E(L_number_plus)
    δ1.3 := D(𝜎1)
    δ2.3 := D(𝜎1)
    𝑛.4 := Φ.number( φ ↦ Φ.bytes( φ ↦ [[ L> 𝜎5 ]] ) )
  E(L_fork)
    δ1.1 := D(𝜎2)
    𝑛1.1 := 𝑛.3
    𝑛2.1 := 𝑛.4
    𝑛.5 := 𝑛1.1
  E(L_number_plus)
    δ1.4 := D(𝜎4)
    δ2.4 := 40-14-00-00-00-00-00-00
    𝑛.6 := Φ.number( φ ↦ Φ.bytes( φ ↦ [[ L> 𝜎6 ]] ) )
```

`M(…)` is the morphing run and its locator. `E(…)` is one firing. `δ1.2` is the
value bound to that entry's `δ1` on its second firing, `𝑛1.1` the same for an
`evaluate` meta, and `𝑛.k` the k-th answer of the whole run, so `𝑛1.1 := 𝑛.3`
reads "the `𝑛1` of this firing is the third answer".

## The order of the firings carries nothing

The blocks come in the order 𝕄 walks the term, and that order is not what the
dependencies are read from — the symbols are. `𝜎3` is minted by the second
firing and consumed by the third as `D(𝜎3)`; `𝜎4` by the third and consumed by
the last; `𝜎2` by the first and consumed by the fork. `𝜎5` is minted and never
consumed, which is how a reader sees that the right branch was computed and
thrown away.

This is why the two `evaluate` lines of `E(L_fork)` are recorded even though
nothing fires under them: they are the only edge from the fork back to the branch
it answered with. Without them `𝑛.5 := 𝑛1.1` names a meta nothing ever bound.

Both branches are already in normal form by the time `L_fork` fires — deep
morphing reduces `x.plus(x.plus(1))` and `x.plus(x)` where they sit as arguments
of `if`, long before the dispatch that fires the fork. Running this program
through the current binary, with the three atoms stripped of their guards, gives
exactly that order:

```
{"λ":"L_number_gt","𝑓0":"S_1"}
{"λ":"L_number_plus","𝑓0":"S_2"}
{"λ":"L_number_plus","𝑓0":"S_3"}
{"λ":"L_number_plus","𝑓0":"S_4"}
{"λ":"L_fork","morph":"𝑛1","at":"begin"}
{"λ":"L_fork","morph":"𝑛1","at":"end"}
{"λ":"L_fork","morph":"𝑛2","at":"begin"}
{"λ":"L_fork","morph":"𝑛2","at":"end"}
{"λ":"L_fork"}
{"λ":"L_number_plus","𝑓0":"S_5"}
```

So no laziness has to be introduced. The indentation stays all the same: an
atom's premise may still fire another atom while reducing an operand, and that
firing belongs one level deeper. In this program it never happens, because deep
morphing gets there first.

## What has to change

- `src/CLI/Parsers.hs` — `--functions` becomes `--symbolic`, `--evaluations`
  becomes `--protocol`
- `src/Parser.hs` — `𝜎` joins `metaVar`, both indexed and bare, and the rest of
  the `AST -> CST -> Text` pipeline carries it: `AST.hs`, `CST.hs`, `Render.hs`,
  `Sugar.hs`, `Lining.hs`, `LaTeX.hs`, `Encoding.hs`
- `src/Lambdas.hs` — drop `_when`, `_extras` and `_symbols`, rename `_morphed`
  to `_evaluated`, and answer one entry instead of a list
- `src/Morph.hs` — `symbol` stops meeting conditions and building extras, and
  records a premise binding, an answer and a depth
- `src/Dataize.hs` — dataizing a symbol answers 42 instead of parking
- `src/Deps.hs` — `Evaluation` gains a depth and renders as indented text
- `test-resources/functions/primitives.yaml` becomes `atoms.yaml` with no guards

Nine of the fourteen packs in `test-resources/dataization-packs/` expect an atom
to produce real data and go: `adds-two-numbers`,
`cannot-add-a-non-numeric-operand`, `dataizes-a-program-of-nothing-but-lambda-functions`,
`fires-twice-down-a-chain-of-dispatches`, `inverts-bytes`, `tells-5-equals-5`,
`tells-5-is-not-equal-to-6`, `tells-different-bytes-are-not-equal`,
`tells-equal-bytes-are-equal`. The tenth, `gets-stuck-on-a-lambda-no-entry-answers`,
expects exactly that and survives. `CmpBytes` in `src/Yaml.hs` stays: it serves a
rewriting rule's `when`, which is untouched.

yegor256:
This drops `--atoms` and puts `--functions` in its place. A λ function is no longer a script phino spawns and talks to over a pipe — it is a rule, written in the same language phino's own judgments are written in, that 𝔼 consults when it fires. An entry keys itself by a regular expression over λ names, reduces the operands it names through 𝔻 (`dataize`) or 𝕄 (`morph`), may mint fresh λ names for what it cannot decide (`symbols`), guards itself with the `when` a rewriting rule takes, computes with the `where` one takes, and answers with the term under `𝑛`. Several entries may answer one name and the guards tell them apart.

The old design paid a process, a JSON round trip and a re-parse of the universe for every firing, and an operand a script could not reduce by itself had to be asked back over a channel of its own — questions, `serve` entries, `ask` lines and all. None of that is needed when the answer is a rule: everything stays in the process, so a dataization is as fast and as reproducible as a rewriting run, and a λ function that cannot decide answers with a fresh symbol rather than failing. Recursion is left alone on purpose — whether a λ function ends is the object model's business — so a run that will not finish ends on `--max-steps`, which `--partial` parks as a residual.

Two things change for anyone downstream. `--atoms` and the whole JSON registry are gone, together with `rt`, `script`, `path`, `serve` and the process channel. And `--evaluations` now writes one JSON object per line — the function under `λ` next to every meta its entry bound — instead of three tab-separated fields, which also means it no longer insists on `--output=phi`. The README has the new file format and the protocol, and `test-resources/morph-symbol-packs/` carries a pack per behaviour, each showing the file, the input program, the full protocol and the answer.

Closes #1226


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any part of the new file format or walk through the protocol change if something reads oddly.

PR #1231, opened 2026-09-16
Spell the bytes meta head as italic delta

### LINKED ISSUE #1230: Bytes meta head is plain Greek δ, while every other meta head is italic (src/Render.hs:131)
Every meta head in the engine is written in mathematical italic, the way the paper on 𝜑-calculus writes its metavariables — 𝐵, 𝑒, 𝑛, 𝑘, 𝜏, 𝑖, 𝑓 — except one: the bytes meta stays plain Greek δ (U+03B4).
`src/Render.hs:131` renders it as `δ` in the same table that renders `𝜏` and `𝑖`, and `src/Parser.hs:164` reads it as `metaVar 'd' "δ"` while its siblings a few lines below are `metaVar 't' "𝜏"` and `metaVar 'i' "𝑖"`.

The odd one out is visible inside a single rule: `resources/dataization/delta.yaml:6` reads `⟦𝐵1, Δ ⤍ δ0, 𝐵2⟧`, where a plain Greek δ0 sits beside an italic 𝐵1.
A meta is a variable, not syntax, so it should look like every other variable; φ, ρ, ξ and α are terminals and stay straight.

Replace δ with 𝛿 (U+1D6FF) in `src/Render.hs:131`, in `src/Parser.hs:164`, in the five dataization rules under `resources/` that carry a `δ0`, in the test packs and in `README.md`.
Plain δ should be dropped rather than accepted beside 𝛿, which breaks rule files written against the old spelling; the ASCII form `!d` and the LaTeX form `\delta` stay as they are.

Thayorns:
Every meta head in the engine is written in mathematical italic, the way the paper writes its metavariables, except the bytes meta, which was still plain Greek δ (U+03B4). This switches it to 𝛿 (U+1D6FF) in the parser and the renderer, so a rule such as `delta` now reads `⟦𝐵1, Δ ⤍ 𝛿0, 𝐵2⟧` with the same kind of glyph on both sides.

The five dataization rules that carry a `δ0`, the custom rewriter pack with `δ_HELLO`, the specs and the README are moved to the new spelling. Plain δ is no longer accepted, and the parser spec pins that. The ASCII form `!d` and the LaTeX form `\delta` stay as they were.

In the README the `explain --dataize` sample for the `delta` rule showed a bare `δ`, which is not what the LaTeX renderer prints; it now shows `\delta_0`, matching the `B_1` index style next to it.

Closes #1230


Thayorns:
@yegor256  take a look please, happy to clarify anything about the switch from δ to 𝛿

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3908](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1233, opened 2026-09-16
fix(#1178): write output files atomically through a temporary file and rename

### LINKED ISSUE #1178: writeFile calls lack atomicity — risk of partial/corrupted writes on crash
### Problem

Several `writeFile` calls in phino lack atomicity protection. If an exception occurs AFTER the file is opened but BEFORE all data is flushed, the file will be corrupted/incomplete. GHC's lazy I/O model makes this particularly risky — `writeFile` = `openFile` + lazy `hGetContents` + `hPutStr`, and the buffer flush happens asynchronously.

Compared to other file operations in the codebase (e.g., `CLI/Helpers.hs:72-79` which correctly uses `bracket opened hClose`), these paths have no cleanup or atomicity guarantee.

### Location

**src/Deps.hs:56** — Step saving for normalization steps
```haskell
writeFile path content
```

**src/CLI/Helpers.hs:232** — printOut for evaluations/outputs
```haskell
printOut target content = case target of
  Nothing -> putStrLn content
  Just file -> writeFile file content
```

**src/CLI/Runners.hs:114, 120** — Output writing
```haskell
output target expr = case (_inPlace, target, _inputFile) of
  (True, _, Just file) -> writeFile file expr           -- L114
  (False, Just file, _) -> writeFile file expr          -- L120
```

### Risk Assessment

| Path | Impact of corruption | Frequency |
|------|---------------------|-----------|
| Deps.hs:56 | Partial step file in --steps-dir, user may see incomplete XML output during debugging | Low (debug-only feature) |
| CLI/Helpers.hs:232 | Corrupted evaluation log (--evaluations), partial PHI/XMIR output to target file | Medium (affects users using --evaluations or --target) |
| CLI/Runners.hs:114 | Corrupted in-place edit of source .phi file → **data loss risk** | High (user loses original content) |
| CLI/Runners.hs:120 | Corrupted output file to --target path | Medium |

The `--in-place` path (line 114) is the most dangerous because it overwrites the user's source file directly. A crash mid-write could leave the source file partially truncated or with invalid φ-calculus syntax that breaks downstream tools.

### Expected Behavior

- All file writes should be atomic: write to temp file first, then rename
- On crash, original files should remain intact
- Partially written outputs should not corrupt existing data

### Suggested Fix

Use atomic write pattern (temp file + rename):

```haskell
import System.IO (withFile, hSetBinaryMode, Binary)
import qualified Data.ByteString.Lazy as BSL
import System.Directory (getTemporaryDirectory, createTempFile, renameFile, removeFile)

atomicWrite :: FilePath -> String -> IO ()
atomicWrite target content = do
  tmpDir <- getTemporaryDirectory
  (tmpPath, tmpHandle) <- createTempFile tmpDir "phino-atom-.tmp"
  BSL.hPut tmpHandle (BSL.pack content)
  hClose tmpHandle
  renameFile tmpPath target   -- atomic on POSIX systems
```

For `--in-place` specifically, consider writing backup first:
```haskell
when (True) $ copyFile srcPath (srcPath <> ".bak")  -- safety net
```

### Related Issues

- **Task 45** (PLAN) — Round-2 finding", \
  --body-file /Users/vstas/AI_Projects/objectionary/phino/.opencode/tmp-issue45.md 2>&1

Thayorns:
Every file phino wrote went through writeFile, which truncates the target before the first byte of the new content is written. When rendering fails half-way, the target is left empty or partial, and with --in-place that means the user's own source file is lost.

The new Files.overwrite writes the content into a temporary file next to the target, closes it, carries over the permissions of the file it replaces and renames it into place. When anything fails before the rename, the temporary file is removed and the target keeps its previous content. The step files of --steps-dir, the --target output and the --in-place rewrite all go through it now.

Closes #1178


Thayorns:
@yegor256 take a look when you get a chance, happy to clarify anything about the atomic write.

0crat:
@thayorns Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3924](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #150, opened 2025-06-27
transitive encapsulation operator is not implemented


yegor256:
This is my expression:

```
{⟦ foo ↦ ⟦ t ↦ ⟦ k ↦ ⟦ bar ↦ ∅ ⟧ ⟧ ⟧ ⟧}
```

This is what I want to do:

```yaml
name: foo
pattern: '⟦ 𝜏 ↝ ∅ ⟧'    # pay attention, a new arrow symbol here!
result: '⟦ 𝜏 ↦ ⟦⟧ ⟧'
```

In order to get this:

```
{⟦ foo ↦ ⟦⟧ ⟧}
```

We introduced this "transitive encapsulation" operator in [this paper](https://dl.acm.org/doi/pdf/10.1145/3638584.3638679). Simply put, `𝜏 ↝ ∅` means that `𝜏` is bound to `∅` _through_ a number of formations.

yegor256:
@maxonfjvipon FYI

VasilevNStas:
@yegor256 @maxonfjvipon 

Interested in giving this a go — the blocker for me is that the operator needs a precise matching semantics first, and the issue example admits (at least) three readings. Fixing one of them seems to be the only "hard" part; the plumbing is straightforward.

**Q1. Which descent makes `τ ↝ B` true?**
The reporter's witness `⟦ foo ↦ ⟦ t ↦ ⟦ k ↦ ⟦ bar ↦ ∅ ⟧ ⟧ ⟧ ⟧` is a chain where every formation has exactly one binding — but that may be incidental:

- **(a) only-binding chain**: `τ ↦ X`, then `X` is a formation whose *sole* binding continues the chain down to `τ′ ↦ B` (the example fits; a sibling binding like `⟦ foo ↦ ⟦ t ↦ ∅, other ↦ 5 ⟧ ⟧` does *not* match);
- **(b) existential any-binding**: descend through *any* binding of each formation (siblings allowed) — closest to "through a number of formations", but one `τ` then matches many witness paths;
- **(c) ρ-following**: descend only through `ρ` (the "encapsulation" axis in the sense of the paper?).

I read the ACM paper's definition as (b), but (a) is what `↦ ↦ ↦ …` on a single spine shows. @yegor256 @maxonfjvipon — which one is the canon?

**Q2. Is `↝` a pattern-only relation, or general?**
[...]

PR #1236, opened 2026-09-16
fix(#153): add locator function that turns a dataized string into a dispatch rooted at Φ or ξ

### LINKED ISSUE #153: locators are not supported by needed
Once #150 is implemented, we need this.

This is my expression:

```
{⟦ foo ↦ ⟦ bar ↦ ⟦ t ↦ ⟦⟧ ⟧ ⟧, m ↦ ∅ ⟧}
```

This is what I want to do:

```yaml
name: foo
pattern: '⟦ 𝜅 ↝ 𝑒, 𝜏 ↦ ∅ ⟧'  # transitive encapsulation here!
result: '⟦ 𝜏 ↦ 𝑒 ⟧'
where:
  - meta: 𝜅
    function: locator
    args:
      - "foo.bar.t"
```

In order to get this:

```
{⟦ m ↦ ⟦⟧ ⟧}
```

The `locator` is currently missing, but needed. It should take the expression, dataize it to bytes, and then convert bytes to kappa (if possible). If not possible (due to syntax error), raise exception.

Thayorns:
Rules had no way to build a locator from data, so a `where` extension could not turn a string like `"Φ.foo.bar.t"` into something a result or pattern can use. This adds a `locator` function next to `tau`: it dataizes its single argument to bytes, parses the text as a phi expression, and returns it only when it is a chain of dispatches rooted at `Φ` or `ξ`. Text that does not parse, or parses into anything else (a formation, a dispatch off a formation), raises an error.

On the question of bare paths raised in the thread: the parser already reads `foo.bar.t` as `ξ.foo.bar.t`, a path relative to the current formation, which is what the example in the issue wants. So nothing is prepended, and both `Φ.foo.bar.t` and `foo.bar.t` are accepted as written.

The result is bound to an expression meta for now. The exact rule from the issue, with the `↝` arrow and a `𝜅` meta, still depends on #150 and can use this function once that lands.

Closes #153


Thayorns:
@yegor256 take a look please, happy to clarify anything about the bare-path handling or the split from #150.

Thayorns:
@yegor256 updated, take a look

Thayorns:
@yegor256, fixed, have a look

Thayorns:
@yegor256 ping, could use a look

PR #1240, opened 2026-09-16
Stop morphing on a term already being reduced, with --acyclic

### LINKED ISSUE #1239: morph has no cycle detection: an entry answering with a firing of itself spends the whole --max-steps budget
The step budget is the only thing that ends the 𝕄/𝔻 recursion. `deeper` (`src/Morph.hs:141`) is charged at four places — `morph'` (`:236`), `deepened.go` (`:352`), `fired` (`:420`) and `dataize'` (`src/Dataize.hs:90`) — and none of them remembers a term it has already reduced. Neither `src/Morph.hs` nor `src/Dataize.hs` imports `Data.Set` or `Data.Map`; there is no visited set, no memo table and no occurs-check anywhere on that path. The rewriter does keep one (`Seen`, `seenMember`, `seenInsert` at `src/Rewriter.hs:34-50`, keyed by `hashExpression` in `src/AST.hs:121`), but it is reset to `Map.empty` on every call to `rewrite` (`src/Rewriter.hs:232`), so it never crosses a 𝕄/𝔻 frame.

The result is that a `--symbolic` entry answering with a firing of itself runs to the budget and then fails naming the budget rather than the loop. `test-resources/morph-symbol-packs/spends-the-step-budget-on-an-entry-answering-itself.yaml` records exactly this: `L_loop` answers `⟦ λ ⤍ L_loop ⟧`, bit for bit the formation just fired, so `morph'` re-enters at `src/Morph.hs:275` with an identical expression, universe and state on every turn, and the pack has to cap `steps: 6` to keep the fixture cheap. With the default `--max-steps=1000` a caller pays a thousand frames to learn that the second frame was already enough.

The smallest fix is an opt-in `--acyclic` flag on `morph`, mirroring `--deep`: add `_acyclic :: Bool` and a `_seen :: Seen` to `ReduceContext` (`src/Morph.hs:86-101`), and in `morph'` check the incoming expression against `_seen` beside the existing `deeper caller` call, inserting it on the way down. Because `ReduceContext` flows down and never back up, that set is a branch set and only a genuine ancestor repeat trips it. On a repeat, raise a `Looping`/`LoopingAt` pair alongside the existing `Stuck`/`StuckAt` and park the site the way `--partial` already parks a stuck λ function, so the run answers the residue and exits 0. Preventing recursion stays the object model's business by default; the flag is how a caller asks otherwise.

Related but distinct: #1175 asks for cycle detection on the atom ask-channel, whose `src/Atoms.hs` was removed by #1226; #1078 is about the universe lacking `true`/`false`/`number`; #797 added the rewriter's own loop-detection set.

yegor256:
This adds a `--acyclic` flag to `morph`. With it on, every frame of 𝕄 carries
the terms the frames above it are reducing, and the moment one comes back the
run stops there and parks that site — the same way `--partial` parks a λ
function that cannot fire. The answer is the term the spine had reached, and
the command exits 0.

Right now `--max-steps` is the only thing that ends the 𝕄/𝔻 recursion, so an
entry answering with a firing of itself burns the whole budget and then fails
on the limit rather than on the loop. That loop is a literal fixed point — the
same expression, universe and state come round again — so one lookup catches it
on the second turn. I reused the rewriter's own loop store for it, so no new
machinery. The context only ever flows down the recursion, which is what makes
the remembered set a branch and not a global: two sibling subterms written
alike stay two terms.

Two things worth a reviewer's eye. The flag is `morph`-only and deliberately
carries no `--partial` guard — the signal can only exist where `--acyclic` put
it, so asking for the guard is already asking to be parked on what it finds.
And I found that a reduction escaping into a universe of its own was leaking
that universe's derivation into `--sequence`, showing the synthetic attribute
an operand was reduced under; `unparked` now drops it, and there is a pack
covering that case.

Closes #1239


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any of it, especially the choice to park without needing --partial.

0crat:
@yegor256 Thanks for the contribution! You've earned +4 points for this: +16 as a basis; -4 for too many hits-of-code (265 >= 100); -8 for the lack of code review. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+3595](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1247, opened 2026-09-17
Render a bytes meta with its ASCII head in ASCII encoding

### LINKED ISSUE #1232: ASCII encoding keeps the Unicode head of a bytes meta in src/Encoding.hs:67
@yegor256 the ASCII encoding in `src/Encoding.hs:67` leaves a bytes meta untouched: `toASCII PA_DELTA{..} = PA_DELTA' bytes` hands the `BYTES` over as they came, there is no `ToASCII BYTES` instance, and `ARG_BYTES` at `src/Encoding.hs:114` is passed through the same way. So `[[ D> !d0 ]]` printed through `printExpression'` with `ASCII` encoding comes out as `D> δ0`, carrying the Unicode head that `src/Render.hs:131` gives `D` (𝛿 once #1230 lands), where `D> !d0` is expected.

Every other meta is rewritten to its ASCII head in the same instance family: `E'`, `N'` and `K'` at `src/Encoding.hs:29`, `B'` at `:41`, `F'` at `:68`, `I'` at `:74`. `META_HEAD` in `src/CST.hs:95` has no ASCII head for bytes at all, since `D'` is the LaTeX `\delta`. That is why the unreachable sibling at `src/Encoding.hs:69` maps `PA_META_DELTA` to `META EXCL D'`, which renders `!\delta` and cannot be parsed back; `test/RenderSpec.hs:146` pins that output. The CLI always prints in Unicode, so only library callers of the ASCII encoding see this.

Adding an ASCII bytes head rendered as `d` to `META_HEAD`, and a `ToASCII BYTES` instance that maps a `BT_META` onto it with `EXCL`, would give `!d0`; `PA_DELTA`, `PA_META_DELTA` and `ARG_BYTES` would then go through it.

Thayorns:
The ASCII encoding left a bytes meta alone. `[[ D> !d0 ]]` printed in ASCII came out as `[[ D> 𝛿0 ]]`, with the Unicode head still there. `META_HEAD` had no ASCII head for bytes. The only other option, `D'`, is the LaTeX `\delta`, so the unused `PA_META_DELTA` branch printed `!\delta`, which the parser cannot read back.

This adds a `D''` head that renders as `d`, plus a `ToASCII BYTES` instance that moves a `BT_META` onto it with the exclamation mark. `PA_DELTA`, `PA_META_DELTA` and `ARG_BYTES` now go through it, the same way every other meta already gets its ASCII head.

The LaTeX renderer runs the ASCII encoding before `toLaTeX`, so `toLaTeX` now maps `D''` back to `D'`. Without that, `explain` printed `d_0` where it should print `\delta_0`. A new round-trip test prints a bytes meta in ASCII and parses the result back.

Closes #1232


Thayorns:
@yegor256 take a look please, happy to clarify anything about the new bytes head.

PR #1251, opened 2026-09-17
Name the 𝕄 block of a λ entry 'morph'

### LINKED ISSUE #1250: The `evaluate:` block of a `--symbolic` entry runs 𝕄, not 𝔼, in src/Morph.hs:697, so its key should be `morph:`
A `--symbolic` entry has two blocks of operands. `dataize:` names the ones brought down through 𝔻 and is spelled after that judgment. Its sibling is spelled `evaluate:` (`src/Lambdas.hs:113`), yet `through` (`src/Morph.hs:697`) reduces each of its operands with `morphing`, which is 𝕄: the operand is taken to its normal form and bound to an expression meta. The comment at `src/Lambdas.hs:24` says so itself, "'evaluate' the ones reduced through 𝕄", and the help text at `src/CLI/Parsers.hs:232` describes the block as "the ones it reduces to a normal form".

This is wrong because 𝔼 already has a meaning in phino, and it is a different one. `_evaluate` (`src/Morph.hs:753`) is the build-term function of 𝔼: it accepts only a formation carrying a λ, fires it, and normalizes the answer, refusing anything else with "Function evaluate() expects a formation with a λ binding". The rules `resources/dataization/fire.yaml:9` and `resources/morphing/ml.yaml:11` use `evaluate:` with exactly that meaning, on a `⟦𝐵1, λ ⤍ 𝑓0, 𝐵2⟧` and the universe, and the protocol marks every firing `𝔼(λ)`. An `evaluate:` operand of an entry is an arbitrary term, often with no λ in it at all: in the fork of `README.md:252`, `$.then` may be a literal, which 𝔼 would refuse and 𝕄 morphs. So one key spells two judgments, and in the entry it spells the one the block does not perform:

```yaml
- λ: L_fork
  dataize:
    𝛿1: $.φ
  evaluate:
    𝑛1: $.then
    𝑛2: $.else
  𝑛: 𝑛1
```

The fix is to rename the key to `morph:`, in `src/Lambdas.hs:113`, the help text at `src/CLI/Parsers.hs:232`, `README.md:252`, `test/LambdasSpec.hs`, and `test/YamlSpec.hs`, so the block is named after the 𝕄 it runs and the `morph` command that exposes it, the way `dataize:` is named after 𝔻 and the `dataize` command. `evaluate:` stays where it is right, as the 𝔼 premise of a rule.

```yaml
- λ: L_fork
  dataize:
    𝛿1: $.φ
  morph:
    𝑛1: $.then
    𝑛2: $.else
  𝑛: 𝑛1
```

yegor256:
A `--symbolic` entry has two operand blocks. The second one was spelled
`evaluate`, and this renames it to `morph`, in the reader, the `--symbolic`
help text, the README, the specs and the test resources. The block itself
works exactly as before.

The name was wrong. That block reduces each operand with 𝕄, binding it to an
expression meta, while 𝔼 in phino means the thing `fire.yaml` and `ml.yaml`
ask for: fire the λ of a formation, refusing anything that carries none. An
operand here is an arbitrary term — in the fork of the README, `$.then` may be
a literal — so one key was spelling two judgments, and in an entry it spelled
the one the block does not perform. `morph` matches the 𝕄 it runs and the
`morph` command that exposes it, the way `dataize` matches 𝔻.

This breaks existing `--symbolic` files: an entry still saying `evaluate`
now binds no operand at all, silently, because an entry ignores a key it does
not know. If that silence bothers you, I can follow up with a check that
refuses an unknown key in an entry.

Closes #1250


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any part of it or split the doc changes out if that reads better.

PR #1253, opened 2026-09-17
Extract Evaluate.hs so each judgment has its own module

### LINKED ISSUE #1252: Lack of Separate Evaluate.hs Causes Misplaced Functions and Tests Related to 𝔼
The calculus has three judgments and phino has two modules for them. `src/Morph.hs` carries Morphing (`morph` at src/Morph.hs:335, `morph'` at :253, `deepened` at :383) and Evaluation as well: `_evaluate` at :753, under a docblock at :741 that opens with "The Evaluation function 𝔼(b, e, s)"; `symbol` at :658, which is the firing itself, from operand reduction through minting to the protocol lines; `fired` at :459; and `lambda` at :167 with `saturated` and `filled` behind it. Those four are called from nowhere else. Together they are about 250 of the module's 778 lines, more than the whole of `Dataize.hs` (221 lines). There is no `Evaluate.hs`.

This is #1203 one judgment later, and the same reasoning holds. 𝔼 is a relation in its own right, not a helper of 𝕄: the protocol names it beside 𝕄 and 𝔻 (`Evaluation` at src/Deps.hs:81), it alone reads the `--symbolic` file, and it alone mints symbols and counts them in `State`. The mixing leaks into the tests exactly as it did before — `describe "execBuildTerm 'evaluate'"` sits in test/MorphSpec.hs:253 and all eighteen packs under `test-resources/morph-symbol-packs/` describe firings — so under the 1:1 rule 𝔼 has no spec file to live in.

Extract `Evaluate.hs` holding `_evaluate`, `symbol`, `fired` and `lambda`, importing `Morph` for the shared machinery, with `test/EvaluateSpec.hs` and `test-resources/evaluate-packs/` following from it. The back-edge is the one #1203 already answered: 𝕄 reaches 𝔼 at `execBuildTerm` (src/Morph.hs:736) and in the deep walk (:395), so it wants the injected-function treatment `ReduceContext._reduce` (src/Morph.hs:101) already gives the 𝔻 edge.

yegor256:
The Evaluation function 𝔼 moves into `src/Evaluate.hs`: `evaluation` (what an `evaluate` premise fires, and what `_evaluate` used to be), `symbol`, `fired` and the `lambda`/`saturated`/`filled` trio that finds the λ to fire. `test/EvaluateSpec.hs` and `test-resources/evaluate-packs/` follow from the 1:1 rule, so the eighteen firing packs and the `execBuildTerm 'evaluate'` cases no longer sit under a morphing name. `Morph.hs` drops from 778 lines to 570.

We had three judgments and two modules for them, which is what #1252 is about. It is the same complaint #1203 made about 𝕄 when `Dataize.hs` held both relations, one judgment later.

`Evaluate` imports `Morph` for the machinery all three judgments share, and the two edges pointing back — an `evaluate` premise and the `--deep` walk — are injected into `ReduceContext` as `_evaluate` and `_fire`, the way `_reduce` already carries the 𝔻 one. Nothing points back at `Evaluate`. Behaviour is unchanged: no rule file, no protocol line and no command-line option moved, and the only rename a reader will notice is `_evaluate` becoming `Evaluate.evaluation`, which keeps it clear of `Control.Exception.evaluate` and matches how `Dataize` exports `reduction`.

Closes #1252


yegor256:
@yegor256 mind taking a look when you get a chance? Happy to explain any part of the split if something looks off.

PR #1294, opened 2026-09-18
Fix oversized regexp group references

### LINKED ISSUE #1275: regexp replacement overflows on an oversized group index
### Problem

Regexp.substituteGroups treats an out-of-range group reference as a literal, but it parses the digit sequence with read before checking the range. A replacement containing a very large group index therefore raises an integer conversion exception instead of remaining literal.

### Reproduction

1. Compile any valid regex, for example a.
2. Call replaceFirst with replacement $999999999999999999999999999999999999999999 and input a.

The read digits expression in substituteGroups raises an overflow exception.

### Expected

An index that cannot be represented or does not exist should be treated like any other missing group and left unchanged as $..., without crashing the replacement operation.

### Evidence

In src/Regexp.hs, substituteGroups computes let idx = read digits before safeIndex. The fallback in fromMaybe only handles a representable index outside the list, not an oversized one.

gemshrine:
## What changed

Fixes #1275.

`Regexp.substituteGroups` previously parsed every numeric replacement reference with `read` before checking whether the referenced capture group existed. A sufficiently large reference such as `$999999999999999999999999999999999999999999` therefore raised an integer conversion exception instead of being preserved as a literal reference.

The replacement parser now uses `readMaybe`. References that do not fit in `Int`, as well as references outside the available capture-group list, follow the existing literal-reference behavior. A regression case covers an oversized group index.

## Verification

- `git diff --check`
- The regression test was added to `test/RegexpSpec.hs`.


yegor256:
Merged, thanks!

---

PR #1295, opened 2026-09-18
Keep UTF-8 characters intact after empty matches

### LINKED ISSUE #1277: replaceAll splits UTF-8 characters when the regex matches an empty string
## What happens

`Regexp.replaceAll` advances one byte after a zero-length match.

With a UTF-8 input such as `é`, a zero-width replacement like `sed("é", "s/^/X/g")` splits the two-byte character: the zero-length branch uses `B.take 1` and `B.drop 1`. The resulting bytes are invalid UTF-8 and `btsToUnescapedStr` fails.

## Expected result

Advancing after a zero-length match must preserve complete UTF-8 code points. The affected logic is in `src/Regexp.hs`, in `replaceAll`.

gemshrine:
## What changed

Fixes #1277.

`Regexp.replaceAll` advances one byte after a zero-length match. When the input contains a multibyte UTF-8 character, a pattern that can match at each position can therefore insert a replacement between the character's bytes. The resulting byte string is no longer valid UTF-8 and the logical character is corrupted.

The zero-length branch now advances over one complete UTF-8 sequence. It validates continuation bytes and falls back to one byte for malformed input, preserving the previous behavior for non-UTF-8 data. A regression test covers an empty lookahead applied to `é`.

## Verification

- `git diff --check`
- The regression test was added to `test/RegexpSpec.hs`.


yegor256:
Merged, thanks!

---

PR #1296, opened 2026-09-18
Keep all text nodes when parsing XMIR

### LINKED ISSUE #1283: xmirToPhi drops text after a comment inside an o element
`xmirToPhi` drops text after a comment or another text-node boundary inside an `<o>` element.

Minimal reproduction:

1. Provide XMIR containing a data binding whose text is split by an XML comment, for example `<o name="Δ">foo<!-- note -->bar</o>` in an otherwise valid object.
2. Parse it with `xmirToPhi`.
3. The resulting `BiDelta` contains only `foo`.

`hasText` correctly notices both text nodes, but `getText` in `src/XMIR.hs` returns only the first `NodeContent` from `C.child cur`. XML permits comments and CDATA boundaries between adjacent text nodes, and the logical text content is the concatenation of all of them. The same truncation affects λ marker text and application arguments.

Expected behavior is to concatenate all text nodes in document order before converting the content to bytes or a function name.

gemshrine:
## What changed

Fixes #1283.

`xmirToPhi` already detected text in every child of an `<o>` element, but `getText` returned only the first `NodeContent`. An XML comment, CDATA boundary, or another node between text fragments therefore truncated data bindings, lambda markers, and application arguments.

`getText` now concatenates every text node in document order. A regression test compares XMIR with a comment separating `foo` and `bar` against the equivalent contiguous text and verifies that both parse to the same expression.

## Verification

- `git diff --check`
- The regression test was added to `test/XMIRSpec.hs`.


yegor256:
@gemshrine build is broken here

yegor256:
Merged, thanks!

---

PR #1297, opened 2026-09-18
Reject overlapping lambda regex keys

### LINKED ISSUE #1276: overlapping lambda regex keys silently shadow later entries
### Problem

The symbolic YAML loader rejects duplicate keys only when their text is identical. Distinct regular expressions can still match the same lambda name. matched then returns the first entry, so the result depends on YAML order and the later entry is silently unreachable.

### Reproduction

1. Add two entries with keys L_(foo|bar) and L_foo, each with a different answer.
2. Load the file with readLambdas.
3. Call matched for L_foo.

The first entry always wins; reversing the entries changes the answer without changing the lambda name.

### Expected

The loader should reject overlapping keys, or document and validate an explicit precedence rule. A configuration that contains an unreachable lambda entry should not load silently.

### Evidence

readLambdas' unique check compares only entry._key for equality, while matched walks entries with find and applies the first matching compiled regex. This leaves overlapping regexes unchecked.

gemshrine:
## What changed

Fixes #1276.

The symbolic loader previously rejected only textually identical keys. Two different regular expressions could still match the same lambda name, and `matched` would silently select whichever entry appeared first in the YAML file. Reordering the entries could therefore change the selected answer while leaving the lambda name unchanged.

After compiling the keys, the loader now checks every pair for an overlap that is represented by either key. Ambiguous configurations fail while loading with an explicit error naming both keys. The existing exact-duplicate validation remains in place, and a regression test covers `L_(foo|bar)` together with `L_foo`.

## Verification

- `git diff --check`
- The overlap regression test was added to `test/LambdasSpec.hs`.


yegor256:
@gemshrine build is broken here

Issue #1292, opened 2026-09-18
text protocol writes 𝑛3.32 := 𝑛1.32 for two terms that differ, since src/Deps.hs:289 names a term by its first symbol


yegor256:
A `symbolize` line of the text protocol claims its term is the one another meta holds, while the XML protocol, which writes both in full, shows they differ. On master 02c3b565, in the fork of `twice(t) ↦ ⟦ φ ↦ ξ.t.gt( 0 ).if( ξ.t.plus( 1 ), ξ.t.plus( 2 ) ).plus( ξ.t.plus( 1 ) ) ⟧` over a library-shaped `number` with `neg ↦ ⟦ φ ↦ ξ.ρ.times( -1 ) ⟧`, the text reads:

```
  𝑛1.32 := 𝑛.31  # ξ.left
  …
  𝔻(⟦ λ ⤍ 𝜎41 ⟧) == BF-F0-00-00-00-00-00-00
  𝑛3.32 := 𝑛1.32  # 𝑛1
```

so `𝑛3` is said to be the very term `𝑛1` holds. The XML for the same two metas, trimmed to the binding they differ in:

```xml
<bind meta="𝑛1.32">⟦ φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎39 ⟧ ), …, neg ↦ ⟦ φ ↦ ρ.times( -1 ) ⟧, … ⟧</bind>
<known symbol="𝜎41">BF-F0-00-00-00-00-00-00</known>
<bind meta="𝑛3.32">⟦ φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎39 ⟧ ), …, neg ↦ ⟦ φ ↦ ρ.times( Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎41 ⟧ ) ) ) ⟧, … ⟧</bind>
```

The `==` fact one line above proves that something was replaced, and the text says nothing was.

`valued` at `src/Deps.hs:285-291` does what its comment says: "a term is looked up by the first symbol it carries and, where that symbol has a name already, written as that name". `𝑛1` and its symbolized copy `𝑛3` both carry `𝜎39` first, so the copy is written as `𝑛1.32` although it carries `𝜎41` where the original carries `-1`. Where two terms differ in `φ` their first symbols differ and the line is written in full, so the shortcut lies exactly when the difference sits deeper, which is where a reader cannot see it.

[...]

Issue #1268, opened 2026-09-17
Inconsistent Object Shapes due to Unmorphed Deep Walk Firing Results


yegor256:
When the deep walk fires a λ function, it stands the entry's answer back into the program as the entry wrote it: `FiringFunc` (`src/Morph.hs:74`) is documented at `src/Morph.hs:68` to answer "the raw term the entry wrote" rather than a normal form, `answered` (`src/Evaluate.hs:138`) builds that term from the template and the fresh symbols and hands it straight back, and `deepened` puts it in place at `src/Morph.hs:381`. The same term written in the program is morphed. On this program, with an entry answering `Φ.number( φ ↦ ⟦ λ ⤍ 𝜎 ⟧ )` for `L_number_plus`:

```
⟦
  bytes ↦ ⟦ φ ↦ ∅ ⟧,
  number ↦ ⟦ φ ↦ ∅, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧,
  a ↦ Φ.number( φ ↦ ⟦ λ ⤍ 𝜎1 ⟧ ),
  c ↦ ⟦ t ↦ Φ.number( φ ↦ ⟦ λ ⤍ 𝜎3 ⟧ ), φ ↦ ξ.t.plus( 1 ) ⟧
⟧
```

`morph --locator=Q.a` answers the expanded formation, `⟦ φ ↦ ⟦ λ ⤍ 𝜎1 ⟧, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧`, while `morph --deep --locator=Q.c` fires `plus` and leaves `φ ↦ Φ.number( φ ↦ ⟦ λ ⤍ 𝜎4 ⟧ )`, the template as written. In a `--symbolic` run this means that a branch of an `if` that came out of a firing is bound to `Φ.number( φ ↦ ⟦ λ ⤍ 𝜎8 ⟧ )`, while a branch that was a literal or a void is bound to `⟦ φ ↦ Φ.bytes( φ ↦ ⟦ Δ ⤍ 3F-F0-… ⟧ ), plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧, … ⟧`: one forma, two different terms. The join asked for in #1246 compares the two branches of a fork leaf by leaf, and it cannot start until they have one shape, so today it would fail on the most common fork, a computed value against a literal.

[...]

PR #1299, opened 2026-09-18
Write a firing's answer on two protocol lines

### LINKED ISSUE #1298: `EvAnswer` Fails to Record Morphed Answer Data in Protocol, Obscuring Traceability
Since #1268 a firing morphs its answer, and the protocol keeps no record of the step. `answered` at `src/Evaluate.hs:203-210` builds the term the entry wrote into `built` (line 207), reduces it through `settled` and then saves the result alone: `ctx._saveEval (EvAnswer ctx._nesting normal)` at line 209. The reader sees a term the entry never wrote. In `test-resources/evaluate-packs/nests-the-firing-an-operand-brought-down.yaml` the entry answers with `Φ.number( φ ↦ ⟦ λ ⤍ 𝜎 ⟧ )` and the protocol line is a formation of 418 characters, `𝑛.1 := ⟦ φ ↦ ⟦ λ ⤍ 𝜎1 ⟧, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧, ρ ↦ ⟦ … ⟧ ⟧`, standing there with nothing to say where it came from.

Every other binding line of the protocol carries both sides of its binding. `EvData` and `EvTerm` append the term the entry wrote as a comment — `𝛿1.1 := 40-14-00-00-00-00-00-00  # ξ.ρ` — and `EvJoin` names the two metas it joined. The answer line alone gives an outcome and no origin, so nothing tells a reader which part of that formation the entry supplied and which part 𝕄 unfolded. It also hides the firings the morphing itself causes: `settled` at `src/Evaluate.hs:306` runs `morph'`, `morph'` fires `ml` (`resources/morphing/ml.yaml`), and a λ function fired while the answer was being morphed opens its own `𝔼(…)` block one level deeper, above the answer line and below nothing that accounts for it. The one step #1268 added is the one step the protocol does not report.

Let `EvAnswer` (`src/Deps.hs:159`) carry the built term beside the normal one and have `answered` save the built one before `settled` runs and the normal one after, so each firing writes two answer lines and whatever the morphing fired stands between them. `written` (`src/Deps.hs:285`) and `elements` (`src/Deps.hs:416`) then spell the pair; the second line has to be written out in full rather than collapsed into the name of the first, since both terms carry the same first symbol and `valued` would otherwise print `𝑛.2 := 𝑛.1`, which is #1292.

yegor256:
A firing used to write its answer on one line of `--protocol`, and since
#1268 that line holds a term nobody wrote: the entry answers
`Φ.number( φ ↦ ⟦ λ ⤍ 𝜎 ⟧ )` and the protocol prints the four-hundred-character
formation 𝕄 makes of it. Now it writes two lines — `𝑛.1.1` for the term the
entry wrote and `𝑛.1.2` for the normal form, the second commented with
`𝕄(𝑛.1.1)` so it says where it came from. The XML gets a `<built>` element
beside the `<answer>` it always wrote, so anything parsing `<answer>` keeps
reading what it read before.

The point is the step in between. Morphing an answer can fire more λ
functions, and those firings already opened their own blocks inside the
firing that caused them — above the answer line, with nothing above them to
account for it. With the built term on a line of its own they now stand
between the two, so the file reads in the order the work was done.

Two smaller things came with it. Answers are numbered by the firing instead
of by a counter of their own, since a firing answers once, which drops
`_answered` and `_answers` and lets `labelled` name an answer the way it
names every other meta — `𝑛1.2 := 𝑛.3.2` now reads "what the third firing
answered" rather than "the third answer". And the built line deliberately
claims no name for the symbol it carries, because the line under it holds a
different term with the same first symbol and #1292 would otherwise print it
as `𝑛.1.2 := 𝑛.1.1`. Where 𝕄 leaves an answer alone the two lines spell the
same term twice; that is the cost of showing the step, and #1292 is the
place to shorten it.

Closes #1298


yegor256:
@yegor256 mind taking a look when you get a moment? Happy to explain any part of it that reads oddly.

Issue #1306, opened 2026-09-18
Inconsistent Commenting of Terms and Judgments in Protocol Reduces Readability


yegor256:
Half the comments of the protocol name a term and half of them name a judgment applied to a term, so a reader has to know which line is which before the column reads as one thing:

```
𝕄(Φ.l🌵)
  𝔼(L_entry)  # Φ.l🌵.e1
    𝛿1.1 := 3F-F0-00-00-00-00-00-00  # ξ.n
    ?(L_number_nope)  # 𝕄(⟦ λ ⤍ L_number_nope ⟧)
    𝑛1.1 := ⟦ λ ⤍ L_number_nope ⟧.φ  # ξ.v.φ
```

`ξ.n` and `ξ.v.φ` are the terms two operands were reduced from, and nothing on either line says that one of them went down through 𝔻 and the other through 𝕄 — which is the whole difference between a line that ends in bytes and a line that ends in a term. Spell the intent that led to the value and every comment says the same kind of thing:

```
𝕄(Φ.l🌵)
  𝔼(L_entry)  # 𝕄(Φ.l🌵.e1)
    𝛿1.1 := 3F-F0-00-00-00-00-00-00  # 𝔻(ξ.n)
    ?(L_number_nope)  # 𝕄(⟦ λ ⤍ L_number_nope ⟧)
    𝑛1.1 := ⟦ λ ⤍ L_number_nope ⟧.φ  # 𝕄(ξ.v.φ)
```

Two lines already read that way and are the argument for the rest: `?(…)  # 𝕄(⟦ λ ⤍ L_none ⟧)` names the judgment that asked for a firing nothing answered (#1300), and `𝑛.1.2 := …  # 𝕄(𝑛.1.1)` names the morphing that made an answer out of the term an entry wrote (#1298).

## Where it is

* `commented` (`src/Deps.hs:363`) writes the comment of every operand line as the bare term, and it is shared by `EvData` (`src/Deps.hs:299`) and `EvTerm` (`src/Deps.hs:308`), so it cannot tell a 𝔻 from a 𝕄 today. Either the letter goes in at the two call sites or the record carries it.
* `EvFiring` (`src/Deps.hs:289`) writes the bare site #1302 gave it. The judgment is already in hand: `caller._judgment` is what `EvStuck` carries, and 𝔼 is fired from the `ml` rule of morphing and the `fire` rule of dataization, so a firing line can say which of the two was running.
[...]

PR #1309, opened 2026-09-18
fix(deps): #1292 name a protocol term by the whole term, not its first symbol

### LINKED ISSUE #1292: text protocol writes 𝑛3.32 := 𝑛1.32 for two terms that differ, since src/Deps.hs:289 names a term by its first symbol

yegor256:
Closes #1292.

## What was wrong

`valued` in `src/Deps.hs` looked a term up in `_named` by `denoted term` — the
*first* symbol the term carries. A `symbolize` line stands the data of a term
into unknowns without touching what the term carries first, so the copy it
binds and the term it was made of open with the very same 𝜎 and differ further
down. The lookup hit, and the copy was written as the original's name:

```
𝑛3.32 := 𝑛1.32  # 𝑛1
```

while the XML protocol, writing both terms in full, shows them differing under
a binding deeper in. The `𝔻(⟦ λ ⤍ 𝜎41 ⟧) == BF-F0-…` fact one line above proves
something was replaced, and the text said nothing was. Where two terms differ
in `φ` their first symbols differ and the line was written out anyway, so the
shortcut lied exactly where the difference sits too deep for a reader to see.

## What changed

`_named` is now a `Named`:

```haskell
type Named = Map.Map Int [(Expression, T.Text)]
```

keyed on a cheap fixed-size digest of the whole term (`hashExpression`) with an
exact structural comparison behind it, the way `Seen` already keys the terms
`--acyclic` has walked. A collision is resolved by `==`, so the common case
stays O(1) on the digest while a name stands for the very term it was given to
and nothing else. Terms don't match verbatim → they are different → the term is
written out in full, however deep the difference sits.

`borrowed` no longer needs a symbol of its own: only a term carrying one is
ever inserted, so a term with no symbol misses the lookup exactly as it used to
miss the `denoted` guard.

The XML protocol is untouched — it always wrote both terms in full, which is
how the issue was spotted in the first place.

## Tests

New pack `test-resources/evaluate-packs/writes-a-symbolized-copy-of-a-term-out-in-full.yaml`
reproduces the bug in the small: `𝑛1` carries `02-` under `k`, `𝑛2` carries the
symbol that datum was stood into, and both open with `𝜎7`. Before the fix it
writes

```
𝑛2.1 := 𝑛1.1  # 𝑛1
```

and with it the term is written out.

The three existing packs that abbreviate a term to a name
(`stands-the-data-of-a-term-into-unknowns`, `joins-the-branches-a-program-forks-into`,
`reduces-the-branches-of-a-fork-through-morphing`) are unchanged, since in each
of them the term really is verbatim the one the name stands for.

## Second commit: the two stale packs `master` was red on

`b766105` is not part of the #1292 fix. #1307 (for #1306, which made a line's
comment carry the judgment that made it) and #1308 landed out of order, so the
two packs #1308 touched kept the comment format `commented` no longer writes,
and `master` has been red since:

* `walks-on-past-a-firing-it-could-not-make.yaml` — expected `# Φ.e1` and
  `# ξ.v`, now `# 𝕄(Φ.e1)` and `# 𝔻(ξ.v)`
* `blames-the-lambda-the-operand-was-parked-on.yaml` — expected `# Φ.x`, now
  `# 𝕄(Φ.x)`

Four comments in two YAML expectations, matching every other pack in
`test-resources/evaluate-packs`. No source file is touched by that commit, so
it changes nothing about behaviour — it only stops the suite asserting a format
the code stopped writing two commits ago.

`cabal test --ghc-options=-Werror`: **2524 examples, 0 failures.**

https://claude.ai/code/session_01W2MNZarFk5dv3ugUViEpDW


yegor256:
~~Every job that runs the test suite is red — `cabal`, `stack` and `coverage` alike — and it is not this PR's failure.~~

**Resolved in `b766105`.** On request I fixed it here rather than leaving it for a separate change, so this PR now carries both.

The two packs #1308 touched kept the pre-#1306 comment format that `commented` no longer writes, because #1307 and #1308 landed out of order. Four comments in two YAML expectations now name the judgment, the way every other pack in `test-resources/evaluate-packs` does:

* `walks-on-past-a-firing-it-could-not-make.yaml` — `# Φ.e1` → `# 𝕄(Φ.e1)`, `# ξ.v` → `# 𝔻(ξ.v)`
* `blames-the-lambda-the-operand-was-parked-on.yaml` — `# Φ.x` → `# 𝕄(Φ.x)` (twice)

No source file is touched by that commit, so it changes nothing about behaviour — it only stops the suite asserting a format the code stopped writing. `coverage` was failing on the same test failure before the 80% threshold was ever measured.

`cabal test --ghc-options=-Werror` locally: **2524 examples, 0 failures.**

---

Issue #1288, opened 2026-09-18
a dataize operand parked by --partial ends the run and blames the firing's own λ in src/Evaluate.hs:117


yegor256:
A λ function no entry answers inside a `dataize` operand is parked by `--partial` and then ends the run anyway, with a message naming a λ that has an entry. On master 77d4594b, with this world:

```
⟦
  bytes ↦ ⟦ φ ↦ ∅ ⟧,
  number ↦ ⟦ φ ↦ ∅, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧, nope ↦ ⟦ λ ⤍ L_number_nope ⟧ ⟧,
  l🌵 ↦ ⟦
    mark(n, v) ↦ ⟦ λ ⤍ L_entry ⟧,
    root(v) ↦ ⟦ λ ⤍ L_root ⟧,
    e1 ↦ Φ.l🌵.mark( 1, Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎1 ⟧ ) ).nope ),
    e2 ↦ Φ.l🌵.mark( 2, Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎2 ⟧ ) ).plus( 5 ) )
  ⟧
⟧
```

and this `atoms.yaml`, where `L_number_nope` has no entry on purpose:

```yaml
- λ: L_number_plus
  dataize:
    𝛿1: $.ρ
    𝛿2: $.x
  𝑛: Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎 ⟧ ) )
- λ: L_entry
  dataize:
    𝛿1: $.n
  morph:
    𝑛1: $.v.φ
  𝑛: Φ.l🌵.root( 𝑛1 )
- λ: L_root
  dataize:
    𝛿1: $.v
  𝑛: ⟦ ⟧
```

the run ends inside `e1`, and `e2`, which is fine, never runs:

```
$ phino morph --deep --symbolic=atoms.yaml --locator='Q.l🌵' --protocol=p.txt --acyclic --partial --quiet world.phi
[ERROR]: No entry of --symbolic answers the λ function 'L_root'
```

with this protocol:

```
𝕄(Φ.l🌵)
  𝔼(L_entry)  # Φ.l🌵.e1
    𝛿1.1 := 3F-F0-00-00-00-00-00-00  # ξ.n
[...]

PR #1310, opened 2026-09-18
fix(dataize): #1290 guard the dataization recursion with --acyclic too

### LINKED ISSUE #1290: --acyclic guards morph only, so a dataize operand over a term that comes back to itself never returns (src/Dataize.hs:89)
`--acyclic` (#1239) stops 𝕄 on a term it is already reducing, but a `dataize` operand of a firing reduces through 𝔻, where the guard is never consulted, so the same term spins forever. On master 02c3b565, with this world:

```
⟦
  bytes ↦ ⟦ φ ↦ ∅ ⟧,
  number ↦ ⟦ φ ↦ ∅, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧,
  demo ↦ ⟦
    cyc(x) ↦ ⟦ φ ↦ Φ.demo.cyc( ξ.x ) ⟧,
    fine(x) ↦ ⟦ φ ↦ ξ.x.plus( 7 ) ⟧
  ⟧,
  l🌵 ↦ ⟦
    mark(n, v) ↦ ⟦ λ ⤍ L_entry ⟧,
    root(v) ↦ ⟦ λ ⤍ L_root ⟧,
    e1 ↦ Φ.l🌵.mark( 1, Φ.demo.cyc( Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎1 ⟧ ) ) ) ),
    e2 ↦ Φ.l🌵.mark( 2, Φ.demo.fine( Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎2 ⟧ ) ) ) )
  ⟧
⟧
```

a mark that only morphs the body parks the cycle in two seconds and goes on to `e2`:

```yaml
- λ: L_entry
  dataize:
    𝛿1: $.n
  morph:
    𝑛1: $.v.φ
  𝑛: 𝑛1
```

```
𝕄(Φ.l🌵)
  𝔼(L_entry)
    𝛿1.1 := 3F-F0-00-00-00-00-00-00  # ξ.n
    𝑛1.1 := ⟦ x ↦ Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎1 ⟧ ) ), φ ↦ Φ.demo.cyc( x ) ⟧  # ξ.v.φ
    𝑛.1 := ⟦ x ↦ Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎1 ⟧ ) ), φ ↦ Φ.demo.cyc( x ) ⟧
  𝔼(L_entry)
    𝛿1.2 := 40-00-00-00-00-00-00-00  # ξ.n
    …
```

A mark that hands the morphed body to a second λ function whose entry dataizes it never returns:

```yaml
- λ: L_entry
  dataize:
    𝛿1: $.n
  morph:
    𝑛1: $.v.φ
  𝑛: Φ.l🌵.root( 𝑛1 )
- λ: L_root
  dataize:
    𝛿1: $.v
  𝑛: ⟦ ⟧
```

```
$ phino morph --deep --symbolic=atoms.yaml --locator='Q.l🌵' --protocol=p.txt --acyclic --partial --quiet world.phi
```

I killed it after 60 s with the protocol file still empty, and `--max-steps=3000` did not end it within two minutes either. A body that grows instead of cycling, `loop(x) ↦ ⟦ φ ↦ Φ.demo.loop( ξ.x.plus( 1 ) ) ⟧`, behaves the same: parked after three firings of `plus` under the first mark, never returning under the second. The `dataize` command itself refuses the flag: `phino dataize --acyclic` answers `Invalid option '--acyclic'`.

`unvisited` at `src/Morph.hs:255-259` is the whole of `--acyclic`, and it is called once, from `morph` at :239. `dataize'` at `src/Dataize.hs:88-96` calls `deeper caller` and walks the dataization rules without ever consulting `_seen`, so a `dataize` operand (`down`, `src/Evaluate.hs:112-114`) runs the very loop the flag was added for.

The fix is to call `unvisited` in `dataize'` too, before the rules are walked, and to let `down` turn the `Looping` into a stuck firing that `--partial` parks; `--acyclic` should be accepted on the `dataize` command in `src/CLI/Parsers.hs` as well. The library objectionary/eo#8548 runs over is full of recursion, so any design that dataizes a result depends on this.

yegor256:
Closes #1290.

## What was wrong

`unvisited` was a `where`-bound helper of `morph'` and `dataize'` never asked it, so `--acyclic` only ever noticed a term 𝕄 came back to. 𝔻 recurses into itself through `box` and `fire` without 𝕄 ever being handed the same term twice, so a body dispatching the object it stands in — the shape a `dataize` operand of a λ function reduces over — went round until `--max-steps` ran out. `dataize` did not even accept the flag.

## What changed

- **`unvisited` is now top level in `Morph`** and exported; `dataize'` asks it before the rules are walked, exactly as `morph'` does.
- **One store per judgment.** `ReduceContext` gains `_dataized` next to `_seen`, and `unvisited` picks between them by `_judgment`. This is load-bearing: `norm` hands 𝕄 the very term 𝔻 was asked about, so a single shared store would read that handover as a repeat and park every term 𝔻 morphs.
- **A loop under a firing parks the site the firing stands at.** With `--partial` the operand's own run of 𝔻 ends on the residue it reached (new `LoopingAt` case in `dataize`) and the entry is left with nothing to bind, so it gets stuck the way one whose λ cannot fire does; without `--partial` the signal travels up to the 𝕄 spine, which parks on it as it already did for a loop 𝕄 caught. Either road, same answer.
- **`dataize` accepts `--acyclic`.** 𝔻 insists on bytes and a parked term carries none, so there the flag wants `--partial` to have a residual program to print; without it the run still stops on the loop, naming the term it came back to instead of running the budget down.
- `Looping`'s message no longer says "Morphing", since either judgment may raise it.

The branch is merged up to master. `Evaluate.hs` is master's, untouched: the `--deep` walk needed a guard around `symbol` in `evaluated` for a loop met while an entry brought its operand down, and #1288 (b0f50cc) had already added one — falling back on the state that firing reached rather than the one the walk came in with, which is better than what this branch originally did. That part is dropped whole and the two `--deep` cases below are re-verified against master's guard.

## Checks

On the merge commit: `cabal test --ghc-options=-Werror` (2530 examples, 0 failures), `hlint src app test` (no hints), `fourmolu --mode check` clean over every touched file. New coverage:

- `test-resources/evaluate-packs/parks-a-firing-whose-operand-dataizes-in-circles.yaml` and `…/parks-the-same-dataization-loop-under-partial.yaml` — the same 𝔻-only loop under a firing, with and without `--partial`, both parking on the same term and writing the same protocol.
- Four `dataize --acyclic` cases in `CLISpec`: the budget failure without the flag, the named loop with it, the residue under `--partial`, and an unchanged answer for a program that never comes back.

## One thing the issue's own world still does not do

The second world in #1290 is not, in fact, a repeat: its terms **grow** on every round. Reduced to its core,

```
⟦ cyc ↦ ⟦ x ↦ ∅, φ ↦ Φ.cyc( ξ.x ) ⟧, t ↦ Φ.cyc( ⟦⟧ ) ⟧            # parks, 0.02s
⟦ cyc ↦ ⟦ x ↦ ∅, φ ↦ Φ.cyc( ξ.x ) ⟧, n ↦ ⟦ φ ↦ ∅ ⟧,
  t ↦ Φ.cyc( Φ.n( φ ↦ ⟦ Δ ⤍ 01- ⟧ ) ) ⟧                            # still runs away
```

Where the argument is a formation the term stabilises and the new guard cuts it. Where the argument is an application over a dispatch, each `box`/`dot` round decorates it with a fresh `ρ` that `stay` never gets to collapse, so the working expression roughly doubles every round (371 → 973 → 2177 → 4585 characters over the first rounds) and no exact-repeat guard can see it. `--acyclic` promises nothing about that shape — the README says so — and `--max-depth`/`--max-cycles` make no difference, so it is not the normalizer giving up early either. Happy to open a separate issue for the `ρ` accumulation if you agree it is one; it looked too much like its own bug to fold in here.


0crat:
@yegor256 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`claude/trusting-carson-zqjqav`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`1309`". Your running score is [+3531](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1291, opened 2026-09-18
benchmark harness times parsing, printing, rewriting, and XMIR only, so a 45x slowdown of --symbolic morphing went unnoticed (benchmark/Main.hs)


yegor256:
`benchmark/Main.hs` imports `Parser`, `Printer`, `Rewriter`, and `XMIR` and times them over `benchmark/tmp/native.phi`; nothing in it runs `morph --symbolic`, `Evaluate`, or `Lambdas`, so `make bench` and the comment `benchmark.yml` posts say nothing about symbolic morphing.

Between 7ea3b52a and 02c3b565, both built on 2026-09-17, the same six entries of a 2 KB demo world went from 0.12 s to 5.6 s under `morph --deep --symbolic --protocol --acyclic --partial --quiet`, with `far` alone at 1.9 s. On 7ea3b52a one entry, `number.neg`, took 0.02 s in a world of six objects and 5.5 s in the 232 KB world of `eo-runtime`, the same body and the same firings. Neither the regression nor the dependence on the size of the world shows up anywhere.

The proposal is a symbolic case in the harness next to the existing ones, over a world like this:

```
⟦
  number ↦ ⟦
    φ ↦ ∅,
    plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧,
    times(x) ↦ ⟦ λ ⤍ L_number_times ⟧,
    gt(x) ↦ ⟦ λ ⤍ L_number_gt ⟧,
    neg ↦ ⟦ φ ↦ ξ.ρ.times( -1 ) ⟧,
    minus(x) ↦ ⟦ φ ↦ ξ.ρ.plus( ξ.x.neg ) ⟧
  ⟧,
  demo ↦ ⟦
    gap(a, b) ↦ ⟦ φ ↦ ξ.a.minus( ξ.b ).times( ξ.a.minus( ξ.b ) ) ⟧,
    far(p, q) ↦ ⟦ φ ↦ Φ.demo.gap( ξ.p, ξ.q ).gt( 100 ) ⟧,
    clamp(x, lo, hi) ↦ ⟦ φ ↦ ξ.x.gt( ξ.hi ).if( ξ.hi, ξ.lo.gt( ξ.x ).if( ξ.lo, ξ.x ) ) ⟧,
[...]

PR #1312, opened 2026-09-18
fix(evaluate): #1287 get stuck on a λ naming a symbol, don't refuse it

### LINKED ISSUE #1287: evaluation crashes on a formation whose λ names a symbol in src/Evaluate.hs:65, where --partial should park the site
Dispatching an attribute off a symbol ends a `--symbolic` run with a `userError` instead of getting stuck. A symbol is `⟦ λ ⤍ 𝜎1 ⟧`, a formation whose one λ names `𝜎1` and nothing else, and the whole point of a symbol is that nobody worked its value out, so `⟦ λ ⤍ 𝜎1 ⟧.plus( 5 )` cannot be reduced. It should be left where it stands, the way `--partial` leaves a λ function no entry answers. Instead the run aborts.

On master 77d4594b, with this world:

```
⟦
  bytes ↦ ⟦ φ ↦ ∅ ⟧,
  number ↦ ⟦ φ ↦ ∅, plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧,
  demo ↦ ⟦
    a ↦ ⟦ λ ⤍ 𝜎1 ⟧.plus( 5 ),
    b ↦ Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎2 ⟧ ) ).plus( 5 )
  ⟧
⟧
```

and this `atoms.yaml`:

```yaml
- λ: L_number_plus
  dataize:
    𝛿1: $.ρ
    𝛿2: $.x
  𝑛: Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎 ⟧ ) )
```

the run ends at `a`, and `b`, which is fine, never runs:

```
$ phino morph --deep --symbolic=atoms.yaml --locator='Q.demo' --protocol=p.txt --acyclic --partial --quiet world.phi
[ERROR]: user error (Function evaluate() expects a formation with a single λ binding naming a function)
```

The protocol holds one line, the root of the walk:

```
𝕄(Φ.demo)
```

`a` is `⟦ λ ⤍ 𝜎1 ⟧.plus( 5 )`. To dispatch `.plus`, 𝕄 fires 𝔼 on the target `⟦ λ ⤍ 𝜎1 ⟧`, since it carries a λ. `evaluation` at `src/Evaluate.hs:53-65` builds the formation, sees it has a λ, and asks `lambda bds` at :333-336 for the name of that λ. `lambda` matches only `BiLambda (Function func)`, a plain name; the binding here is `BiLambda (FnSymbol 1)`, a symbol (`src/AST.hs:95`), so it falls to the `_ -> Nothing` case, and `evaluation` at :65 throws the `userError`. Nothing catches it, so the run ends before `b`.

## What I would like to see

Firing 𝔼 on a symbol should get stuck the way firing a λ no entry answers does, so `--partial` parks the site. The dispatch `.plus( 5 )` off a parked target cannot reduce either, so the whole of `a` is left as written, a taint, and the walk goes on to `b`, which computes. The run exits 0 with this protocol:

```
𝕄(Φ.demo)
  ?(𝜎1)  # 𝕄(⟦ λ ⤍ 𝜎1 ⟧)
  𝔼(L_number_plus)  # Φ.demo.b
    𝛿1.1 := 𝔻(⟦ λ ⤍ 𝜎2 ⟧)  # ξ.ρ
    𝛿2.1 := 40-14-00-00-00-00-00-00  # ξ.x
    𝑛.1.1 := Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎3 ⟧ ) )  # 𝑛
    𝑛.1.2 := ⟦ φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎3 ⟧ ), plus(x) ↦ ⟦ λ ⤍ L_number_plus ⟧ ⟧  # 𝕄(𝑛.1.1)
```

The block of `b`, firing 1, is exactly what a run over `b` alone writes today. The line `?(𝜎1)` is the one new thing: a stuck record for the symbol, spelled the way `?(L_number_nope)` is spelled for a λ no entry answers. Its exact spelling is phino's call — what matters is that the run does not abort, that `a` is left as a taint with no root, and that `b` still runs. To the module of objectionary/eo#8548 `a` is then a taint, since no firing of it bound a root, and `b` is rooted on `𝜎3`.

The fix is at `evaluation`, `src/Evaluate.hs:60-65`: when the single λ of the formation is a `FnSymbol` rather than a `Function`, save an `EvStuck` for the symbol and throw `Stuck`, exactly as `symbol` does at :88-98 for a name no entry answers, rather than falling to the `userError` at :65. The walk's `evaluated` then parks it under `_partial`, once that path parks a `Stuck` at all (objectionary/phino#1288).

This is the one place objectionary/eo#8548 cannot dodge by planting: a void whose carrier the inference tables cannot name has to stand as a bare symbol, and one such void ends the whole run today. A 29-entry run over `eo-runtime` died this way after 3 min 55 s.

yegor256:
Closes #1287.

## What was wrong

A symbol is `⟦ λ ⤍ 𝜎1 ⟧` — a value nobody worked out. Dispatching an attribute off one asks 𝕄 to reduce it, `ml` fires 𝔼 on the target, and `evaluation` asked `lambda` for the name of that λ. `lambda` answers only for a plain `Function`, so a `FnSymbol` fell through to the `userError` at the bottom of the case. Nothing catches it, so the run died on the first such term, everything standing after it never ran, and the protocol kept nothing but its own root.

Reproduced on master `03781a1` with the world and `atoms.yaml` from the issue, verbatim:

```
[ERROR]: user error (Function evaluate() expects a formation with a single λ binding naming a function)
```

## What changed

`evaluation` now tells a λ naming a symbol apart from a λ it genuinely cannot make sense of. Nothing answers a symbol, which is exactly what a λ name the `--symbolic` file does not carry means, so 𝔼 answers it the same way `symbol` does: write the site to the protocol as `?(𝜎1)`, then throw `Stuck`. `--partial` parks it — #1288 having already put the guard around the deep walk's own firing — so one value nobody worked out taints its own binding and the walk goes on.

Several λ bindings, or one standing for a meta or a slot, still fail: the `userError` keeps the ground it was meant for.

After the fix, the issue's own command exits 0 and gives:

```
𝕄(Φ.demo)
  ?(𝜎1)  # 𝕄(⟦ λ ⤍ 𝜎1, ρ ↦ ∅ ⟧)
  ?(𝜎1)  # 𝕄(⟦ λ ⤍ 𝜎1, ρ ↦ ∅ ⟧)
  𝔼(L_number_plus)  # 𝕄(Φ.demo.b)
    𝛿1.1 := 𝔻(⟦ λ ⤍ 𝜎2, ρ ↦ ∅ ⟧)  # 𝔻(ξ.ρ)
    𝛿2.1 := 40-14-00-00-00-00-00-00  # 𝔻(ξ.x)
    𝑛.1.1 := Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎3, ρ ↦ ∅ ⟧ ) )  # 𝑛
    𝑛.1.2 := ⟦ φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎3, ρ ↦ ∅ ⟧ ), … ⟧  # 𝕄(𝑛.1.1)
```

`a` is left exactly as written, a taint with no root; `b` fires and computes, rooted on `𝜎3`. The `b` block is character-for-character the one the issue asked for.

**One difference from the issue's illustration:** `?(𝜎1)` is written twice, not once. The walk passes the site twice — once for the dispatch `….plus` and once for the application `….plus( 5 )` around it — because `_parked` is only populated when the *spine* parks a site, not when the deep walk does. This is pre-existing and not about symbols: on master, `⟦ λ ⤍ L_nope ⟧.plus( ⟦⟧ )` under `--deep --partial` writes `?(L_nope)` twice in exactly the same way. Since the issue says the spelling is phino's call and asks that a symbol behave *like* an unanswered name, I matched that behaviour rather than diverging from it. Happy to file the `_parked` gap separately if you want the walk deduplicated — it needs the walk to thread what it has parked, which is its own change.

## Checks

`cabal test --ghc-options=-Werror` — 2531 examples, 0 failures. `hlint`, `fourmolu 0.17.0.0` (the version CI pins), `yamllint 1.38.0`, `typos 1.50.2` with the repo config, `reuse lint` (556/556) and `markdownlint-cli2 0.23.2` all clean locally.

Coverage:

- `test-resources/evaluate-packs/parks-a-dispatch-off-a-symbol.yaml` — the issue's world end to end: the run finishes, `a` is a taint, `b` computes, and the protocol is asserted verbatim.
- A new `EvaluateSpec` case driving `execBuildTerm "evaluate"` straight at `⟦ λ ⤍ 𝜎1 ⟧`, asserting both the `Stuck` message and the `?(𝜎1)` record.

Two existing tests asserted the old behaviour and are updated, not deleted:

- `EvaluateSpec`'s "the only λ binding carries a symbol rather than a name" moves out of the malformed-argument list into the new case above.
- `CLISpec`'s "writes the terminator as the term a meta was bound to" used the old `userError` as its stopper; it now ends on the symbol and the XML carries `<stuck λ="𝜎1" judgment="morph">⟦ λ ⤍ 𝜎1 ⟧</stuck>`, with the document still closing properly.

README and `CLAUDE.md` updated to say that firing a symbol gets stuck the way an unanswered name does — the README previously implied only dataizing a symbol was defined.

---
_Generated by [Claude Code](https://claude.ai/code/session_01A1L7n3WTj3hggjob3U4LnZ)_


yegor256:
`regression-check` failed here, and it is not this PR's fault — but it is mine, so I fixed it in `d541d89` rather than leave it.

`bench:bench` has not compiled on `master` since #1310 landed. That PR added `_dataized` to `ReduceContext` next to `_seen`; #1291 had meanwhile added `symbolicCtx` in `benchmark/Main.hs`, which builds the record positionally. They touch different files, so git merged them cleanly, and #1310 was merged before `regression-check` reported on it — so the field count has been one short on master ever since, and the check has been failing on every branch since.

Verified rather than assumed: I built `bench:bench` from a worktree at `03781a1` (current master, this PR's base) and it fails there with the same eight `GHC-83865` errors at `benchmark/Main.hs:78-99`. This PR touches nothing under `benchmark/`.

`d541d89` adds the one missing `Map.empty -- _dataized` line. It no-ops the moment master carries the same fix. If you'd rather have it as its own PR against master, say so and I'll move it — I only have the one branch to work on here.

Local state on the current head: `cabal build all --enable-tests --enable-benchmarks --ghc-options=-Werror` clean, `cabal test --ghc-options=-Werror` 2531 examples / 0 failures, and hlint, fourmolu 0.17.0.0, yamllint, typos, reuse and markdownlint all clean — `hlint` and `fourmolu` over `benchmark/` too, since I had reason to touch it.

---

yegor256:
Correcting my previous comment on one point that changes what to do about it: **`regression-check` cannot go green on this PR, or on any other, until the fix is on `master`.** I said the fix "no-ops once master carries it", which is true but understated it.

The workflow's first step is:

```yaml
- name: Build bench and frozen fixture at merge-base
  run: |
    git checkout ${{ github.event.pull_request.base.sha }}
    cabal build all --enable-benchmarks
```

It builds the **base** before it ever builds the head, to get a baseline binary to compare against. The base is `03781a1`, where `benchmark/Main.hs` is still one field short, so the job dies there. The log confirms it: the errors it prints quote `Morphing [] Map.empty lambdas buildTerm …` with a single `Map.empty` — that is the base's source, not this branch's, which has two. The "Build bench at PR head" step never ran.

So this is not a flake and a re-run would fail identically — I confirmed it deterministically twice over, by building `bench:bench` from a worktree at `03781a1` locally and by the single-`Map.empty` fingerprint in the log. Nothing I can push here clears it; the red check is the base's, not this PR's.

Merging this PR is what fixes it: once `d541d89` is on master, the base build succeeds and the check goes green again for everything afterwards. Everything else on this head is green, and the head's own `bench:bench` builds clean locally under `cabal build all --enable-tests --enable-benchmarks --ghc-options=-Werror`.

I'd rather not touch the workflow to work around a real failure, and I only have this one branch, so I can't send the one-liner to master separately — but say the word and I'll do it however you prefer.

---

0crat:
@yegor256 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`claude/trusting-carson-zqjqav`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`1311`". Your running score is [+3537](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1314, opened 2026-09-18
fix(lambdas): #1293 symbolize and join the φ chain and nothing else

### LINKED ISSUE #1293: symbolize and join walk every binding of a formation and mint symbols for literals inside method bodies (src/Lambdas.hs:349)
Each branch of a fork reaches the join as the whole formation of its carrier, methods included, and `symbolized` at `src/Lambdas.hs:344-351` descends into every binding of it. On master 02c3b565, for `twice(t) ↦ ⟦ φ ↦ ξ.t.gt( 0 ).if( ξ.t.plus( 1 ), ξ.t.plus( 2 ) ).plus( ξ.t.plus( 1 ) ) ⟧` over a library-shaped `number` with `neg ↦ ⟦ φ ↦ ξ.ρ.times( -1 ) ⟧`, the `-1` inside `neg` becomes `𝜎41` in one branch and `𝜎42` in the other, and `joined` at :424-434 then pairs them into `𝜎44`:

```
  𝔻(⟦ λ ⤍ 𝜎41 ⟧) == BF-F0-00-00-00-00-00-00
  𝑛3.32 := 𝑛1.32  # 𝑛1
  𝔻(⟦ λ ⤍ 𝜎42 ⟧) == BF-F0-00-00-00-00-00-00
  𝑛4.32 := 𝑛2.32  # 𝑛2
  𝔻(⟦ λ ⤍ 𝜎43 ⟧) ∈ { 𝔻(⟦ λ ⤍ 𝜎39 ⟧), 𝔻(⟦ λ ⤍ 𝜎40 ⟧) }
  𝔻(⟦ λ ⤍ 𝜎44 ⟧) ∈ { 𝔻(⟦ λ ⤍ 𝜎41 ⟧), 𝔻(⟦ λ ⤍ 𝜎42 ⟧) }
  𝑛5.32 := ⟦ φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎43 ⟧ ), …, neg ↦ ⟦ φ ↦ ρ.times( Φ.number( φ ↦ Φ.bytes( φ ↦ ⟦ λ ⤍ 𝜎44 ⟧ ) ) ) ⟧, … ⟧  # [𝑛3, 𝑛4]
```

Only `𝜎43` is the value of the fork. `𝜎41`, `𝜎42`, and `𝜎44` stand for a constant nobody reads, since the value of a term lives where its `φ` chain ends and a method body is code, not data. The real `number` of `eo-runtime` holds dozens of literals inside its methods, so every fork would write dozens of symbols and facts for nothing, the walk pays for each, the protocol grows by them, and a consumer such as objectionary/eo#8548 has to know which symbols to skip.

Both stages already leave `ρ` alone for the same reason, "what sits under `ρ` belongs to the object around this one". The fix is to treat every binding but `φ` the same way: `symbolized` at :349 and `joined` at :432 walk the `φ` chain only and carry every other binding from the first term verbatim, so the answer keeps its shape and its methods, and the only symbols minted are the ones a value can differ by. A separate `strip` block that drops everything but `φ` before the join would not do, since the fork answers the joined term and the program then dispatches on it, `.plus( ξ.t.plus( 1 ) )` above, which a stripped term cannot serve.

yegor256:
Closes #1293.

## What was wrong

`symbolized` and `joined` descended into every binding of a formation. A branch reaches the join as the whole formation of its carrier, methods included, so every literal inside every method got a symbol per branch and a fresh one per pair — the `-1` of a `neg` nobody has called becoming `𝜎41`, `𝜎42` and then `𝜎44`.

## What changed

Both walks now go through `φ` and stop everywhere else, carrying every other binding from the first term verbatim.

That is the rule `ρ` already followed, read off what a term *is* rather than off the attribute name: a term carries the value it stands for where its `φ` chain ends, so a datum sitting anywhere else is not that value. The decisive argument that this is the right rule and not just a convenient one: **`denoted` in `AST.hs:221` has always said exactly this** —

```haskell
-- The symbol a term stands for, if its value is one at all. A term carries its
-- value where the φ chain ends, so that is the only place a symbol names this
-- term: one sitting under ρ, or inside an operand, belongs to the term it was
-- minted for and says nothing about this one.
```

It walks `BiLambda (FnSymbol …)` and `BiTau AtPhi …` and nothing else. The two stages disagreed with it, and that disagreement was the bug.

The joined term keeps its methods, so the program can still dispatch on what the fork answered — which the `strip`-before-join alternative the issue rules out could not have given it.

## Evidence

`test-resources/evaluate-packs/leaves-the-literals-of-a-method-out-of-a-join.yaml` runs the issue's shape end to end. Both branches carry a literal under `φ` and another inside `neg`, and the two `neg` bodies differ:

```
𝔻(⟦ λ ⤍ 𝜎1 ⟧) == 01-
𝑛3.1 := ⟦ φ ↦ ⟦ λ ⤍ 𝜎1 ⟧, neg ↦ ⟦ φ ↦ ⟦ Δ ⤍ 0A- ⟧ ⟧, … ⟧   # 𝑛1
𝔻(⟦ λ ⤍ 𝜎2 ⟧) == 02-
𝑛4.1 := ⟦ φ ↦ ⟦ λ ⤍ 𝜎2 ⟧, neg ↦ ⟦ φ ↦ ⟦ Δ ⤍ 0B- ⟧ ⟧, … ⟧   # 𝑛2
𝔻(⟦ λ ⤍ 𝜎3 ⟧) ∈ { 𝔻(⟦ λ ⤍ 𝜎1 ⟧), 𝔻(⟦ λ ⤍ 𝜎2 ⟧) }
𝑛5.1 := ⟦ φ ↦ ⟦ λ ⤍ 𝜎3 ⟧, neg ↦ ⟦ φ ↦ ⟦ Δ ⤍ 0A- ⟧ ⟧, … ⟧   # [𝑛3, 𝑛4]
```

Three symbols, all on the `φ` chain; the method's `0A-`/`0B-` are untouched, and the join keeps the first branch's copy without comparing it. On master this mints two more symbols and a fourth for their join.

## Tests

2535 examples, 0 failures (2531 on master). Ten tests failed on the first run and I did **not** weaken any of them to match — every one was a fixture that put its data under a plain label and so was testing the walk that is being removed:

- Six `LambdasSpec` cases used `⟦ a ↦ ⟦ Δ ⤍ 00- ⟧, b ↦ ⟦ Δ ⤍ FF- ⟧ ⟧`. Re-pointed onto the `φ` chain (`⟦ φ ↦ Φ.f( φ ↦ ⟦ Δ ⤍ 00- ⟧ )( t ↦ ⟦ Δ ⤍ FF- ⟧ ) ⟧`), they assert exactly what they did before: two data → two symbols, one pair met twice → one symbol, the same counts into the state.
- Three packs and one `CLISpec` case used `l ↦ …` as the carrier. Renamed to `φ ↦ …`; the joins still happen and the protocols are identical in shape.

Four new cases pin the new rule directly: a datum outside the `φ` chain left alone, a method carried from the first branch with nothing minted, and — so the fix can't quietly swallow a real mismatch — two branches whose bindings are *named* differently still refuse to join.

Two protocol lines in the re-pointed packs changed from a written-out term to a back-reference (`𝑛4.1 := 𝑛2.1`). That is #1292's naming rule doing its job: with the value on the `φ` chain, `denoted` now finds a symbol for those terms where with `l` it found none, so they are named and later lines point at them.

## Checks

`cabal build all --enable-tests --enable-benchmarks --ghc-options=-Werror` clean, and `hlint`, `fourmolu 0.17.0.0`, `yamllint 1.38.0`, `typos 1.50.2`, `reuse lint` and `markdownlint-cli2 0.23.2` all clean — hlint and fourmolu over `benchmark/` too.

README updated in both the `symbolize` and `join` sections, which documented the old `ρ`-only rule.

---
_Generated by [Claude Code](https://claude.ai/code/session_01A1L7n3WTj3hggjob3U4LnZ)_


0crat:
@yegor256 It is [not a good idea](https://www.yegor256.com/2014/04/15/github-guidelines.html) to name Git branches the way you named this one: "`claude/trusting-carson-zqjqav`". You've earned -6 points. Next time, better give your branch the same name as the number of the ticket that you are solving. In this case, a perfect name, for example, would be "`1313`". Your running score is [+3543](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1316, opened 2026-09-18
`morph --deep --symbolic` cost scales with the whole world, not the reduced term (`insideUniverse`, src/Morph.hs:607)


yegor256:
`insideUniverse` (`src/Morph.hs:607`) reduces every expression by wrapping it back into the whole universe. It builds `synthetic = ExFormation (BiTau attr expr : bds)`, where `bds` is every top-level binding of the program, and normalizes `expr` inside that. `morphing` (`src/Morph.hs:625`) and `dataize'` (`src/Dataize.hs:218`) call it for each subexpression they reduce, so the reducer and the `--deep` walk re-traverse every object in the world at every step, whether or not the computation ever touches it.

The cost of `morph --deep --symbolic` therefore tracks the size of the whole world, not the work the reduced term does. Measured on `16e131c1` with a 2 KB, six-entry demo world, adding only dead data objects (`⟦ Δ ⤍ 00- ⟧` siblings that no entry dispatches or dataizes): 0 objects → 5.9 s, 200 → 20.9 s, 600 → 97.6 s. Tripling the dead weight multiplied the extra time about sixfold, so the growth is superlinear, and nothing the entries compute changed between the runs. This is the 45x `--symbolic` slowdown named in the title of #1291, which added the benchmark that surfaced it but did not localize it.

The single hot spot is `insideUniverse` at `src/Morph.hs:607-616` and its two callers. One fix is to reduce the expression against the universe by reference, resolving dispatches into `bds` on demand, instead of re-embedding all of `bds` into the term the rewriter rebuilds each step. Another is to mark the universe's bindings as already in normal form so the `--deep` walk and the `unvisited` guard skip them rather than re-walking them every step. Either keeps the cost proportional to the reduced term.

To reproduce, run `morph --deep --symbolic=atoms.yaml --locator='Q.l🌵' --protocol=/dev/null --acyclic --partial --quiet world.phi` over a world of a few entries whose bodies do arithmetic, then add a sibling object holding a few hundred unreferenced `⟦ Δ ⤍ 00- ⟧` bindings and rerun: the same entries take an order of magnitude longer.

PR #1317, opened 2026-09-18
perf(matcher): #1316 walk a formation once to place a meta binding

yegor256:
`matchBindingsMeta` tried every split of the target bindings by cutting the list anew at each index — `splitAt idx tbs | idx <- [0 .. length tbs]`. Placing one meta binding therefore walked a formation of N bindings N times over, and the meta binding after it walked what was left N times again. The `dot` rule carries a meta on either side of the binding a dispatch names, and every dispatch of a program is matched against it, so this ran constantly. The walk now goes down the target once, carrying the run it has already passed and putting that run together only where the rest of the pattern matched; a meta binding with nothing after it takes the whole rest in one step, that being the only split the pattern can still match.

This is why a dead object nobody names cost anything at all. Once `dot` has resolved a `Φ` dispatch the term carries the whole universe under its `ρ`, so every rewriting step re-walked every object in the world. Measured with `morph --deep --symbolic=benchmark/atoms.yaml --locator='Q.l🌵' --protocol=/dev/null --acyclic --partial --quiet` over `benchmark/demo.phi` plus N dead `⟦ Δ ⤍ 00- ⟧` siblings, one machine, same binary shape: N=0 7.85s → 7.59s, N=200 19.40s → 12.13s, N=600 90.10s → 23.31s, N=1200 335.71s → 41.46s. The extra time over the N=0 baseline grew quadratically with the size of the world before and grows linearly after.

Worth saying plainly: the issue names `insideUniverse` (`src/Morph.hs:607`) as the single hot spot, and a profiling build disagrees. Over the N=200 world `matchBindingsMeta` takes 47.7% of the time and 62.2% of the allocation, up from 4.8% and 6.4% over the N=0 world — the one cost centre whose share tracks the dead weight. `insideUniverse` is what hands the world to the matcher, but the superlinear part was the splitting, so this change leaves the representation alone.

`test/MatcherSpec.hs` carries the reproduction: the `dot` pattern matched against a formation of 3000 bindings under a five-second timeout, in the bounded-wait shape `test/RandomSpec.hs` already uses. It times out before the change and finishes in milliseconds after it.

One thing for the reviewer: what is left is still linear in the size of the world, because the world really does ride along inside every term's `ρ` chain — a single protocol line of this run is 617 KB at N=0 and 1.55 MB at N=200. Getting the cost proportional to the reduced term rather than merely linear in the world needs the by-reference reduction the issue sketches, and that changes how the universe reaches a reduced term, so it deserves its own issue instead of riding along here.

`make test` (2537 examples, 0 failures), `make hlint` ("No hints"), `make fourmolu` and `make coverage` (82%, threshold 80%) all pass on this branch.


yegor256:
@yegor256 mind taking a look when you have a moment? Happy to explain the profiling numbers or the split rewrite if anything reads oddly.

Issue #1318, opened 2026-09-19
One morphed term embeds the universe 184 times, so a dead object still costs linear time (resources/normalize/dot.yaml, src/Matcher.hs:175)


yegor256:
#1316 took the superlinear part out of this — `matchBindingsMeta` no longer re-splits a formation once per binding — but the cost of `morph --deep --symbolic` still grows with the size of the world rather than with the term being reduced. Measured on 470b6367 with the command #1316 named, over `benchmark/demo.phi` plus N dead `⟦ Δ ⤍ 00- ⟧` siblings that no entry dispatches or dataizes:

| N | world | time | protocol | protocol with `--hide-rho` |
|---|---|---|---|---|
| 0 | 1,568 B | 8.0 s | 2,200,665 B | 17,146 B |
| 200 | 8,079 B | 12.1 s | 5,935,405 B | 17,146 B |
| 600 | 21,279 B | 23.8 s | 13,479,405 B | 17,146 B |
| 1200 | 41,279 B | 43.4 s | 24,887,405 B | 17,146 B |

The protocol is 151 lines in every run, and with `--hide-rho` the files are byte-identical, same MD5. The entries do exactly the same work and never touch a dead object. Adding 40 KB of dead weight to the world adds 22 MB to those same 151 lines and 35 seconds to the run.

The multiplier is exact and it does not move. Line 52 is the longest line in every run, and in all four it spells `L_number_plus` 184 times: the same term, the same 184 embeddings of the universe, 850 KB of it at N=0 and 9.5 MB at N=1200. Whatever goes into the world is paid for 184 times over inside one term.

Where the copies come from: `resources/normalize/dot.yaml` writes `result: 𝑒1(ρ ↦ ⟦𝐵1, 𝜏1 ↦ 𝑛1, 𝐵2⟧)`, so resolving a dispatch decorates the body with the whole formation it was dispatched on. Once the `universe` morphing rule has resolved `Φ` into the top formation, that formation is the world, and every dispatch inside the result attaches it again, nesting. Nothing downstream drops it. `matchExpressionDeep` (`src/Matcher.hs:175-181`) descends into every τ binding, ρ included, so every rule of every rewriting step re-walks the copies; `hashExpression` under `--acyclic` and the rewriter's fixpoint check (`src/Rewriter.hs:255`) walk them too. The `--deep` walk is the one place that already knows better: `bindings` skips ρ at `src/Morph.hs:482`, "since ρ names the object around it rather than one inside it".

[...]

PR #1334, opened 2026-09-19
perf(normalize): #1318 name the universe in a ρ instead of holding it

yegor256:
The `dot` normalization rule wrote `result: e(ρ ↦ ⟦𝐵1, 𝜏1 ↦ 𝑛1, 𝐵2⟧)`, so resolving a dispatch decorated the body with the whole formation it dispatched on. Once the `universe` morphing rule had resolved Φ into the top formation, that formation was the world, and every dispatch inside the result attached it again, nesting. Over plain `benchmark/demo.phi`, with no dead weight at all, the universe is about 1,000 nodes while the term handed to the rewriter has a median of 2,776, a p90 of 45,257 and a maximum of 804,112 — 184 nested copies of the world under ρ.

A rewriting rule now carries an optional `e-match`, the universe-argument matcher a morphing rule already spells the same way, and a rule carrying one is matched against the universe too and reads what it binds there. `dot` gained the guard "the formation dispatched on is not the universe", and a new `dotg` sibling carries its exact complement and writes `ρ ↦ Φ` — the name of the program rather than the program itself. Whoever reads that ρ resolves Φ through the `universe` morphing rule, which answers with the program in normal form, the very formation that would have stood there, since a dispatch only reaches either rule once Φ has already been resolved that way. Where no universe is known — the `rewrite` command rewrites a term with no world around it, and `isNF` asks about a term on its own — the meta binds nothing, the guard cannot hold, and `dot` answers every dispatch exactly as before.

For the plumbing: `ReduceContext` gained `_universe`, worked out once per run by `universed` in `src/Morph.hs` and inherited by every frame below it, and `insideUniverse` extends it by consing the already-normalized binding on, so the world is never walked twice. `RewriteContext` gained the same field and `Rule.matchExpressionWithRuleIn` seeds a match from it. One related fix rode along: `_eq` now builds both sides under the substitution before comparing them, so a side written as a whole term is compared as the term it stands for and not as the raw pattern it was written as. It compared patterns before, which no rule had yet relied on.

Measured with `morph --deep --symbolic=benchmark/atoms.yaml --locator='Q.l🌵' --protocol=... --acyclic --partial --quiet` over `benchmark/demo.phi` plus N dead `⟦ Δ ⤍ 00- ⟧` siblings, both binaries built back to back on one machine:

| N | before | after | protocol before | protocol after |
|---|---|---|---|---|
| 0 | 8.29s | 0.34s | 2,200,665 B | 20,142 B |
| 200 | 12.14s | 0.44s | 5,935,405 B | 20,142 B |
| 600 | 23.50s | 0.65s | 13,479,405 B | 20,142 B |
| 1200 | 41.65s | 1.07s | 24,887,405 B | 20,142 B |

Run with `--hide-rho`, the protocol is byte-identical between the two binaries at both N=0 and N=1200 (all four MD5s equal), and so is the morphed answer. Nothing but what rides in ρ changed.

`test-resources/morph-packs/names-the-universe-in-a-rho-instead-of-holding-it.yaml` carries the reproduction. Five `morph-deep-packs` expectations, two in `test/CLISpec.hs` and one in `test/DataizeSpec.hs` now spell `ρ ↦ Φ` where they used to spell the whole world out. Those are the printed normal forms this change alters, and they are the part of the diff worth a careful eye.

Two things to flag. The guard failing is always safe: it falls back to `dot`, which writes what it always wrote, so an inexact `_universe` costs speed and never correctness. And `make bench` fails on my machine on a stale `benchmark/tmp/native.phi` — the Makefile generates it with whatever `phino` is on PATH, and the installed 0.0.133 emits the `{⟦ … ⟧}` package syntax the in-tree parser cannot read. Master fails on it identically, so it is not from this change; it deserves its own issue. The benchmark itself compiles and links here, having gained the two new record fields.

`make test` (2538 examples, 0 failures, rebuilt clean under `-Werror`), `make hlint` ("No hints"), `make fourmolu` and `make coverage` (82%, threshold 80%) all pass on this branch.


yegor256:
@yegor256 could you take a look when you get a moment? Happy to explain any part of it, particularly the five `morph-deep-packs` expectations that now print `ρ ↦ Φ`.

PR #1337, opened 2026-09-19
Avoid following directory symlink cycles

### LINKED ISSUE #1273: allPathsIn recurses forever through directory symlink cycles
### Problem

Files.allPathsIn follows every entry for which doesDirectoryExist returns true, including symbolic links. It does not track visited directories or reject symlinks. A directory containing a link back to its parent therefore never terminates.

### Reproduction

1. Create a directory tree with mkdir -p root/sub.
2. Create a cycle with ln -s .. root/sub/back.
3. Call allPathsIn "root".

The function keeps descending through root/sub/back/sub/back until the process runs out of stack or path length.

### Expected

The traversal should either skip symbolic-link directories or track canonical visited paths and return each reachable file once.

### Evidence

The recursion in src/Files.hs checks only doesDirectoryExist path before calling allPathsIn path; there is no symlink or visited-path guard. This can hang any command that scans a user-controlled directory tree.

gemshrine:
Fixes #1273.

allPathsIn followed symbolic links that pointed to directories. A link back to an ancestor therefore made the recursive traversal visit the same directories forever.

The traversal now skips symbolic-link directories while retaining ordinary files and directories. A regression test builds a directory link back to its parent and verifies that the walk returns the real files once.

Checks:

- git diff --check passed.
- The Haskell test suite could not run locally because the container has no Cabal or GHC executable.


yegor256:
Merged, thanks!

---

